PR #60, opened 2018-10-15
(not ready for review) Create sar_pyspark_ml_wrapper.py

dciborow:


nikhilrj:
Closing - we will use Spark SAR directly via MML Spark or another implementation instead of using a wrapper to use our python version

Issue #254, opened 2018-11-19
Save github statistics every day


miguelgfierro:
### *What* is affected by this bug?
We have the code to gather the github statistics and add it to cosmos. We will need to set up a VSTS pipeline so this code is executed every day

### *Where* on the platform does it happen?
1. *Azure Data Science Virtual Machine.*

yueguoguo:
Labelled as high-prior as the repo is open sourced. We need to know the statistics to measure the impact.

yueguoguo:
Solved in #367 #195

miguelgfierro:
Reopening this, the metrics are not being saved every day, the last day we have is `"date" : "Dec 15 2018 08:09:53",`.

It is important to fix this asap, since all the metrics we don't record are lost

PR #367, opened 2018-12-12
GitHub statistics

yueguoguo:
### Description
Updates of scripts to generate Github stats.

### Motivation and Context
To have an idea about the stats change of the repo.

Note because of some issues in Azure DevOps command line task, the Python that imports custom module does not work well in terms of module path. One needs to add the full system path (not the relative ones like `..` or `../..`) to make sure the custom module / script is visible. 

Running the script directly from shell with `python` executable does not have the problem. 

### Releated Issues
#254 

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/scripts/repo_metrics/track_metrics.py
+++ b/scripts/repo_metrics/track_metrics.py
@@ -2,11 +2,10 @@
 # Licensed under the MIT License.
 
 import sys
-sys.path.append('..')
-sys.path.append('../..')
-sys.path.append('/data/home/recocat/cicd/21/s/scripts')
-sys.path.append('/data/home/recocat/cicd/21/s')
-print(sys.path)
+# Need to append a full path instead of relative path.
+# This seems to be an issue from Azure DevOps command line task.
+# NOTE this does not affect running directly in the shell.
+sys.path.append('cicdworkdir/scripts')
 import os
 import argparse
 import traceback
--- a/scripts/repo_metrics/track_metrics.py
+++ b/scripts/repo_metrics/track_metrics.py
@@ -6,8 +6,7 @@ import os
 # Need to append a full path instead of relative path.
 # This seems to be an issue from Azure DevOps command line task.
 # NOTE this does not affect running directly in the shell.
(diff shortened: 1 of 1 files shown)


scripts/repo_metrics/track_metrics.py
```diff
@@ -1,6 +1,13 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+import sys
+# Need to append a full path instead of relative path.
+# This seems to be an issue from Azure DevOps command line task.
+# NOTE this does not affect running directly in the shell.
+sys.path.append('cicdworkdir')
```
miguelgfierro: I'm normally against hardcoding, but I think it is important to have the tracking system working every day. So I will approve and then add some suggestions, if you want to add them now or later is up to you. 
Suggestions:
1) not sure how we are executing this, but I'm sure we can add to sys path the [current folder](https://github.com/miguelgfierro/codebase/blob/master/python/system/paths.py#L15) or maybe the [parent folder](https://github.com/miguelgfierro/codebase/blob/master/python/system/paths.py#L27).
2) It would be good to add a parameter with the date, by default it could be `now()` but let's say we want to add an event that didn't happen today, with the current code we can't do it.

An additional comment, we should record events to make sure we are able to match metrics with marketing actions that we do with respect of time. One example of event that we should record is the day we opensource the repo, when we have a new release or when we publish a new blog

yueguoguo: > not sure how we are executing this, but I'm sure we can add to sys path the current folder or maybe the parent folder.
We execute the python script directly with the command line task in the Azure DevOps pipeline, but the problem is that the command line script somehow cannot recognize relative paths such as `.` or `../..`. Thus I add it manually. But I think it would be definitely better to avoid hardcoding - now the guy is just to make sure things work :)
> An additional comment, we should record events to make sure we are able to match metrics with marketing actions that we do with respect of time.
For this one, the script should be manually executed, right? Yea we can do that definitely.

Issue #369, opened 2018-12-13
Linux Spark staging build failed


wutaomsft:
### *What* is affected by this bug?
Linux Spark staging build failed, as seen in README.md.

yueguoguo:
This is fixed in #376

PR #370, opened 2018-12-13
TEST: remove integration spark test with large dataset

yueguoguo:
### Description
Temporarily remove Spark integration test

### Motivation and Context
Spark integration test is a replicate of Spark smoke test. Use of large data sets lead to issues in the CI/CD pipeline. 

Temporarily remove the tests until there are good fix.

### Releated Issues
#369 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/tests/integration/test_notebooks_pyspark.py
+++ b/tests/integration/test_notebooks_pyspark.py
@@ -10,8 +10,8 @@ from tests.notebooks_common import OUTPUT_NOTEBOOK, KERNEL_NAME
 TOL = 0.5
 
 
+@pytest.mark.skip.(reason="disabling test with large datasets")
 @pytest.mark.spark
-#@pytest.mark.integration
 @pytest.mark.parametrize(
     "size, result_list",
     [
--- a/tests/integration/test_notebooks_pyspark.py
+++ b/tests/integration/test_notebooks_pyspark.py
@@ -11,7 +11,7 @@ TOL = 0.5
 
 
 @pytest.mark.spark
-@pytest.mark.integration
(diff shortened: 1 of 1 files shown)


tests/integration/test_notebooks_pyspark.py
```diff
@@ -11,7 +11,7 @@
 
 
 @pytest.mark.spark
-@pytest.mark.integration
+#@pytest.mark.integration
```
gramhagen: this will work for our case, but I think a more explicit way to handle this would be
```
@pytest.mark.skip(reason="disabling test with large datasets")
```

miguelgfierro:
It looks it is taking longer than 40min, so it time out:
```
2018-12-13T15:03:18.1314392Z 
2018-12-13T15:42:17.9239285Z tests/integration/test_notebooks_pyspark.py FF
2018-12-13T15:42:17.9856461Z ##[error]The task has timed out.
```
probably some of the new changes make the notebook to take longer

yueguoguo:
@anargyri good point - also given characteristics of 100k and 1mil may not be homogeneous based on our observations in the evaluation results.

anargyri:
I think the change of TOL to 0.5 is extreme, as in PR 371. I would make sure everything is deterministic (seed, split etc.) for the tests. If you give ALS the same seed it should do the same thing every time. There is some variation due to machine precision etc. but this should not be 50%.

yueguoguo:
@anargyri will hold this PR until update of the ALS seed.

yueguoguo:
Resolved in #376

PR #371, opened 2018-12-14
TEST: TOL=0.05 may be too strict

yueguoguo:
### Description
Change TOL in smoke test.

### Motivation and Context
Rel tol of 0.05 is too strict, which cannot tolerate the randomness of the calculated metric values. So it has been changed to 0.1. 

### Releated Issues
#369 

### Checklist:

- [x] My code follows the code style of this project, as detailed in our [contribution guidelines](../CONTRIBUTING).
- [x] I have added tests.
- [x] I have updated the documentation accordingly.

PR #376, opened 2018-12-17
Fix pyspark tests

gramhagen:
### Description
- Updates to pyspark notebook 
  - set seed value for ALS
  - update displayed values for new seed
- Updates to smoke test for pyspark notebook
  - using updated values from above
- Updates to integration test for pyspark notebook
  - dropping 10m, and 20m tests, updating values based on seed used above

### Motivation and Context
Fix failing smoke / integration tests and nightly pyspark staging builds
This removes variability introduced during ALS training to make tests work better
Additionally it reduces the size of tests to ensure integration tests can complete in a timely fashion
Addresses #369 

### Releated Issues
Should replace #370 and #371 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/notebooks/00_quick_start/als_pyspark_movielens.ipynb
+++ b/notebooks/00_quick_start/als_pyspark_movielens.ipynb
@@ -13,7 +13,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 19,
    "metadata": {},
    "outputs": [
     {
@@ -57,7 +57,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 20,
    "metadata": {
     "tags": [
      "parameters"
@@ -83,7 +83,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 21,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -101,7 +101,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 22,
    "metadata": {},
    "outputs": [
     {
@@ -161,7 +161,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 23,
    "metadata": {},
    "outputs": [
     {
@@ -190,7 +190,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 24,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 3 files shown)


notebooks/00_quick_start/als_pyspark_movielens.ipynb
```diff
@@ -222,7 +223,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Took 3.917855978012085 seconds for training.\n"
+      "Took 1.6378734111785889 seconds for training.\n"
```
yueguoguo: This is dependent on the VM we use for running the notebook. Do we want to add information to make it more clear?

gramhagen: Sure I can add something above this

tests/integration/test_notebooks_pyspark.py
```diff
@@ -1,50 +1,34 @@
...
-    result_rmse = df.loc[df["name"] == "rmse", "value"].values[0]
-    assert result_rmse == pytest.approx(result_list[4], TOL)
-    result_mae = df.loc[df["name"] == "mae", "value"].values[0]
-    assert result_mae == pytest.approx(result_list[5], TOL)
-    result_exp_var = df.loc[df["name"] == "exp_var", "value"].values[0]
-    assert result_exp_var == pytest.approx(result_list[6], TOL)
-    result_rsquared = df.loc[df["name"] == "rsquared", "value"].values[0]
-    assert result_rsquared == pytest.approx(result_list[7], TOL)
+    results = nb.dataframe.set_index("name")["value"]
+    start_or_get_spark("ALS PySpark").stop()
```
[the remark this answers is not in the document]
gramhagen: I chose to put it here to avoid inadvertently closing the session when using the notebook. If you run all cells and then want to fiddle with something in the notebook you won't be able to. But if we just manage it in the test framework we can avoid that. It's not ideal but I think this is the safest way to use it.

Is there a way to have a cell that only executes if the notebook is under test?

anargyri:
I checked the history of the integration test and TOL = 0.5 goes back to November 1. So, there have been many changes in the notebook since, and the integration tests were meaningless. since you are at it, it is a good opportunity to change also the other test files that have TOL = 0.5. I found the following:
integration/test_notebooks_python.py
smoke/test_notebooks_pyspark.py (it is 0.1)
smoke/test_notebooks_python.py

yueguoguo:
@anargyri probably we do another PR so that this one can resolve the immediate problem. What do you think?

anargyri:
It looks like @gramhagen changed all the occurrences of TOL = 0.5. The unit tests passed; Scott could you check too that smoke and integration tests work (and if not to change the assert statements as appropriate)?

gramhagen:
@anargyri smoke and integration tests pass for normal and spark setups, so I think we're good to go

https://msdata.visualstudio.com/AlgorithmsAndDataScience/_build/results?buildId=2053258
https://msdata.visualstudio.com/AlgorithmsAndDataScience/_build/results?buildId=2053344

Issue #375, opened 2018-12-17
User-friendly guidance on data preparation


yueguoguo:
It takes time in real-world problems that user feel frustrated about preparing data for recommender algorithms. It would be good to have notebooks to detail that.

WessZumino:
Totally agree! In the RBM deepdive I have used some space to explain the logic behind the data splitting and preparation for movielens. However, this is quite a simple scenario as the data are already given in the correct format. It would be good to find a real case scenario in which more work is needed for data preparation

atimesastudios:
May we please also have guidance for FFM dataset preparation to feed xDeepFM, especially with integrating customer and item features into the input dataset along with implicit/explicit feedback.

yueguoguo:
@atimesastudios yea we do have some preparation functions for this
* The `negative_feedback_sampler` (see [here](https://github.com/microsoft/recommenders/blob/f53d2639b0f41348a95aeaaa6da5514039e14c2a/reco_utils/dataset/pandas_df_utils.py#L262)) is the one that samples negative feedback when data is about implicit feedback.
* The `LibffmConverter` (see [here](https://github.com/microsoft/recommenders/blob/f53d2639b0f41348a95aeaaa6da5514039e14c2a/reco_utils/dataset/pandas_df_utils.py#L86)) is the one that performs libffm format conversion from feature data set.

PR #384, opened 2018-12-20
NOTEBOOK: add data transform guidance

yueguoguo:
### Description
Data transformation guidance notebook.

### Motivation and Context
Make it easy for users to understand different data preparation methods for collaborative filtering algorithms.

### Releated Issues
#375 

### Checklist:

- [ ] I have added tests.


anargyri:
Nice notebook. A few comments:
- some of the English text could be improved
- could you remove the python warning (e.g. make a copy of the DF not a view) 
- the resulting affinity / feedback column has different names in the various cells; better to be consistent
- do we want to have a case of multiple implicit feedback columns (e.g. click, time spent, purchase or not)?

maxkazmsft:
Nice notebook - after the discussion today, it might be nice to mention https://en.wikipedia.org/wiki/Ordinal_regression and explain that we don't need thresholds for SAR's ratings which the dependent variables are mapped to. By citing ordinal regression, you can explain why you're mapping dependent categories to integers. Also, I think the integer count should start with 0 like in the Wikipedia article, not 1.

yueguoguo:
> Nice notebook - after the discussion today, it might be nice to mention https://en.wikipedia.org/wiki/Ordinal_regression and explain that we don't need thresholds for SAR's ratings which the dependent variables are mapped to. By citing ordinal regression, you can explain why you're mapping dependent categories to integers. Also, I think the integer count should start with 0 like in the Wikipedia article, not 1.

Good points. Koren's paper [here](https://pdfs.semanticscholar.org/934a/729409d6fbd9894a94d4af66bd82222b5515.pdf) explains how to deal with the ordinal regression settings. I cited and added texts to explain.

Issue #378, opened 2018-12-18
Nonnegative = false in quick start ALS


anargyri:
### *What* is affected by this bug?
Quick start ALS notebook.

### Expected behavior (i.e. solution)
For recommenders NMF is less common, so `nonnegative = False` is typical.

PR #385, opened 2018-12-20
ALS: set nonnegative = False in quick start notebook

anargyri:
See issue #378.


anargyri:
I added some explanation.

PR #390, opened 2018-12-28
Restricted Boltzmann Machine implementation in Tensorflow

WessZumino:
### Description
RBM implementation of collaborative filtering in TF. Although the code runs on a CPU in reasonable times (few minutes), it is advised to use a GPU enabled DSVM. All time metrics reported in the notebooks have been obtained on a single GPU DSVM. 

### Motivation and Context
It adds neural network based,  probabilistic CF capabilities. 

### Checklist:
- [x] I have added tests.


notebooks/00_quick_start/rbm_movielens.ipynb
```diff
@@ -0,0 +1,885 @@
...
+       "4     166      346       1  886397596"
+      ]
+     },
+     "execution_count": 3,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# MovieLens data have different data-format for each size of dataset \n",
```
yueguoguo: `load_pandas_df` can be used for doing this. See [here](https://github.com/Microsoft/Recommenders/blob/master/reco_utils/dataset/movielens.py)

reco_utils/dataset/rbm_splitters.py
```diff
@@ -0,0 +1,251 @@
...
+
+#for logging
+log = logging.getLogger(__name__)
+
+#========================================================
+#Generate the User/Item affinity matrix from a pandas DF
+#========================================================
+
+
+class splitter:
```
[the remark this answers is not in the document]
yueguoguo: Maybe we can consider generalizing the splitter for sparse matrix typed data if it is frequently used for a few recommender algorithms.

notebooks/00_quick_start/rbm_movielens.ipynb
```diff
@@ -0,0 +1,885 @@
...
+     "text": [
+      "done training, Training time 1.7138962\n"
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "training epoch 10 rmse Train 0.702141\n",
+      "precision on the train set 0.40427575\n",
```
yueguoguo: What is `precision` here? In the text it is called "generalization accuracy". And there is also RMSE. It is a little bit confusing....

notebooks/00_quick_start/rbm_movielens.ipynb
```diff
@@ -0,0 +1,885 @@
...
+       "0   0.153406  "
+      ]
+     },
+     "execution_count": 14,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "eval_100k= ranking_metrics(\n",
```
yueguoguo: It would be good to use `papermill` to record the evaluation results for testing purpose. See [SAR](https://github.com/Microsoft/Recommenders/blob/master/notebooks/00_quick_start/sar_single_node_movielens.ipynb) for example.

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1449 @@
...
+    "\n",
+    "$P = \\frac{1}{Z} \\, e^{- \\beta \\, H}$, \n",
+    "\n",
+    "where $Z$ is a normalization constant known as the partition function, $\\beta$ is a noise parameter with units of inverse energy and $H$ is the Hamiltonian, or energy function of the system. For this reason, this class of models is also known as *energy based* in computer science. In physics, $\\beta$ is the inverse temperature of the system in units of Boltzmann's constant, but here we will effectively rescale inside $H$, so that this is now a pure number. $H$ describes the behaviour of two sets of stochastic vectors, typically called $v_i$ (visibles) and $h_j$ (hidden). The former constitutes both the input *and* the ouput of the algo (this will be clear later), while the hidden units are the latent factors we want to learn. This structure results in the following Neural Network topology\n",
+    "\n",
+    "![rbm1](https://393229b54263dsvm.file.core.windows.net/images/RBM1.png?sp=rl&st=2018-12-03T13:06:28Z&se=2020-12-04T13:06:00Z&sv=2017-11-09&sig=XP5y%2BXuU4r3Snmb%2FIgksMR%2B26XZfgrV2ZVs183Nwn3k%3D&sr=f)\n",
```
yueguoguo: Image is not rendered well.

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1449 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1.2 Model \n",
+    "\n",
+    "The central quantity of every SM model is the [Boltzmann distribution ](https://en.wikipedia.org/wiki/Boltzmann_distribution); this can be seen as the least biased probability distribution on a given probability space $\\Sigma$ and can be obtained using a maximum entropy principle on the space of distributions over $\\Sigma$. Its typical form is: \n",
+    "\n",
+    "$P = \\frac{1}{Z} \\, e^{- \\beta \\, H}$, \n",
```
yueguoguo: Two dollar signs can be make the equation aligned to the center, which may display better.

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1449 @@
...
+    "\n",
+    "$\\frac{\\partial}{\\partial w_{ij} } \\Delta F = v_i \\, P_0(v_i = q|h, a) - v_i P_k(v_i| h,a) \\equiv \\langle v_i\\rangle_0 - \\langle v_i \\rangle_k$. \n",
+    "\n",
+    "This process is repeated for each training epoch, eventually until $\\Delta F =0$, i.e. the learned distribution faithfully reproduces the empirical one. In this sense, the $v_i$ serve both as input and output of the model. As $w_{ij}$ contains informations on how users' votes are correlated, we can use this information to generate ratings for the unseen movies by sampling from the learned, marginal distribution:\n",
+    "\n",
+    "$\\langle v_i \\rangle = \\sum_{v_i} v_i \\, P(v)$ \n",
+    "\n",
+    "The entire workflow is summarised below \n",
+    "\n",
+    "![gibbs](https://393229b54263dsvm.file.core.windows.net/images/Gsampling.001.png?sp=rl&st=2018-12-04T08:34:33Z&se=2020-12-05T08:34:00Z&sv=2017-11-09&sig=N0GXdj0UqLsXM2cwIZsWdMgqUTlYvu1pmGhnNmuAOWI%3D&sr=f)\n"
```
yueguoguo: It is not shown well. Perhaps you can put the image on a blob with public access.

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1449 @@
...
+       "\n",
+       "    .dataframe tbody tr th {\n",
+       "        vertical-align: top;\n",
+       "    }\n",
+       "\n",
+       "    .dataframe thead th {\n",
+       "        text-align: right;\n",
+       "    }\n",
+       "</style>\n",
+       "<table border=\"1\" class=\"dataframe\">\n",
```
yueguoguo: This guy does not show properly....

reco_utils/dataset/rbm_splitters.py
```diff
@@ -0,0 +1,251 @@
...
+        #Options to save the model for future use
+        self.save_model_ = save_model
+        self.save_path_ = save_path
+
+    def gen_index(self):
+
+        '''
+        Generate the user/item index
+
+        Args:
```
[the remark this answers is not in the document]
WessZumino: Not sure I understood this. The Docstring is in the relative function. Is this what you meant? The Google example looks the same to me.

anargyri:
Why are there separate RBM splitter functions instead of putting these in `python_splitters.py`? The splitting functionality is more general than RBM, isn't it? What is not covered in the existing splitting methods?

anargyri:
I see a test for splitters but is there a test for the core RBM functionality? Could you also add smoke and integration tests too?

reco_utils/dataset/rbm_splitters.py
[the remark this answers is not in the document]
WessZumino: I found it useful to keep the methods together. The main reason is that in this repo a dataframe oriented approach is used, whereas when working with Neural networks one always need to move to arrays. Another solution is to add index + matrix generation in the main algo, as for the SAR algorithm. Nevertheless, this brings up other problems as I cannot use the splitter logic implemented here.

reco_utils/recommender/rbm/Mrbm_tensorflow.py
```diff
@@ -0,0 +1,950 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
WessZumino: Sure, it is not a problem. I have tried to mantain the same name type used for the other algos like SAR single node versus pyspark etc...

reco_utils/recommender/rbm/Mrbm_tensorflow.py
```diff
@@ -0,0 +1,950 @@
...
+        keep_prob= KEEP_PROB,
+        momentum = 0.93,
+        init_stdv = STDV,
+        learning_rate= ALPHA,
+        minibatch_size= MINIBATCH,
+        training_epoch= EPOCHS,
+        display_epoch = 10,
+        CD_protocol = [50, 70, 80, 90, 100],
+        save_model= False,
+        save_path = 'reco_utils/recommender/rbm/saver',
```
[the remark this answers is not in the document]
WessZumino: note sure about this. Personally I would like my models to be saved in a permanent directory as I will need to retrieve them later. This is an option for saving a trained model for later use, without the need to train again every time.

notebooks/00_quick_start/rbm_movielens.ipynb
WessZumino: the test set precision is complemetary to the generalization error. To be more precise, the generazation error is the limit of the test set error for a test size going to inifinity. This is used to understand the generalization capababilities of a model, this is why it is being comprared to the train set error. It is simply given by: 1/m Sum_{mu=1}^{m} Sum{i=1}^Nv 1/s(i) I(v-vp = 0)_{mu,i}, where  m = Nusers, Nv = number of items = number of visible units and s(i) is the number of non-zero elements per row. Also, I() is an indicator function, v the input value of the visible unit (either form train or test set) and vp is the one predicted by the algo. The explanation is in the Docstring of `Mrbm_tensorflow.py`, but I will also add it to the deep dive notebook for clarity.

WessZumino:
@anargyri yes you are right, the splitting functionality has nothing to do with the algo used at the end. I found more useful to split data in a different way that mantains per user splitting ratio. At the end, I think it is a kind of hybrid stratified/random splitter. Also, I found that doing this directly on numpy's arrays is faster than the dataframe way, especially for the stratified splitter.

WessZumino:
@anargyri yes you are right, I am working on the tests for the core algo. This is more involving due to the complexity of the algo, but I will add it asap. Apart from that, all function, methods and results have been extensively tested.

miguelgfierro:
Agree with Andreas, splitters should be general functions, not related to a specific algo.
In the past we discussed about using dataframes Vs numpys, we had a lot of code in Df so we left it like that, but as you said numpys should be faster. 

I think a middle ground approach would be to have numpy and df splitters,  to follow single responsibility and explicit.

anargyri:
> @anargyri yes you are right, the splitting functionality is partially independent from the algorithm one uses after. Nevertheless, it is essential to have the same user/item matrix structure both in the train and in test set when using the RBM. Using the random splitter does not work and indeed I got errors and/or incosistent results when using it. The stratified splitter on the other hand does the job, but my understanding is that it does this by time and it is therefore deterministic, while a random splitter usually improves the generalizability of the results.

I suspect the issue you refer to is more general than RBM and applies to collaborative filtering, ALS etc. too. I don't see why stratified vs. random etc. is specific to RBM. In any case, we have had plans to add more splitters that cover all cases in the diagram. Feel free to discuss with Le and Miguel about how to add any methods that you need.

Splitting is tricky with all algorithms, but there are reasons to keep it separate from the algorithms. Apart from software best practices like modularity, a more important reason is that splitting relates to how you evaluate an algorithm offline. Hence it has to be suited to the use case and the data scientist needs to have flexibility to choose different splitting methods i.e. we should not force them to use a specific splitter when doing RBM. There are constraints of course and issues like cold users / items, but the data scientist should have flexibility to choose among splitters the ones that are applicable and preferable for their use case. Moreover, in order to be able to compare different algorithms in a fair way, the splitters should be independent of the algorithms.

> Due to the last point, I prefered to keep it separate from the python_splitter, where all oprations are performed on DFs.

This is a valid point but it could be addressed in another way. You could add a `numpy_splitters.py` similar to what we did with `python_splitters.py` and `spark_splitters.py`.

WessZumino:
> I suspect the issue you refer to is more general than RBM and applies to collaborative filtering, ALS etc. too. I don't see why stratified vs. random etc. is specific to RBM. In any case, we have had plans to add more splitters that cover all cases in the diagram. Feel free to discuss with Le and Miguel about how to add any methods that you need.

Indeed, it is more general than RBM and it applies to all CF algorithms. If I understood the issue, I can just remane the splitter as `numpy_splitter.py` as suggested by you and @miguelgfierro. 

> Moreover, in order to be able to compare different algorithms in a fair way, the splitters should be independent of the algorithms.

The splitter is independent of the RBM, and it can be used with ALS and SAR as well. The only constraint, as I said before, is that the train/test set should contain the same number of users/items ( to be more precise, the only constraint is the number of items to be conserved). For this reason, the output of the splitter is somehow redundant, as it generates the train/test set both as numpy sparse matrices and pandas df; the latter can be used for ALS and SAR. It is on my To-do list to check how the performance of these two algos is affected by the splitter.  

I think my problem was as follows: 

The splitter works directly on the user/item affinity matrix X . This means that the workflow should be: 
1.  first generate X from the DF, then apply the splitter to obtain Xtr and Xtst (both np matrices)
2. feed the matrices to the algo for training and prediction 
  
However, the existing algos in the repo follow the inverse workflow:
1.  first split the data from the DF and obtain tr_df and test_df (both data frames)
2. use the dfs directly in the algo, where X_tr and X_tst are obtained. 

This is the reason why I decided to separate the splitter from the available ones. Still, as the metrics are evaluated from a DF, I need to convert everything back at the end, but this takes additional time.

yueguoguo:
> The splitter is independent of the RBM, and it can be used with ALS and SAR as well. The only constraint, as I said before, is that the train/test set should contain the same number of users/items ( to be more precise, the only constraint is the number of items to be conserved). For this reason, the output of the splitter is somehow redundant, as it generates the train/test set both as numpy sparse matrices and pandas df; the latter can be used for ALS and SAR.

I guess the point here is that, if the numpy splitter is generically useful, it would be great to have them implemented separately utility functions together with other splitter implementations, instead of a method under `rbm_splitter`. 

We do have stratified splitter for Pandas DataFrame by the way (see line 114 [here](https://github.com/Microsoft/Recommenders/blob/master/reco_utils/dataset/python_splitters.py)). However, it does stratification on EITHER "user" OR "item". This is why I think having an numpy or sparse matrix implementation that keeps both train and test data the same shape, as you use in RBM here, would be useful.

notebooks/02_model/rbm_deep_dive.ipynb
yueguoguo: I suspect it works on Jupyter because the image file is on the DSVM where Jupyter notebook is hosted. I tried the URL of the image file and it does not show me anything... You can create a new container under an existing storage and authorize open access to the container blob. Ideally the URL of an accessible file is https://<storage_name>.blob.core.windows.net/<container>/<your_file>

reco_utils/dataset/numpy_splitters.py
```diff
@@ -0,0 +1,131 @@
...
+
+import numpy as np
+import random
+import math
+from decimal import Decimal
+
+
+X = np.random.randint(low= 0, high=6, size= ( 20, 50) )
+
+def round(s):
```
miguelgfierro: This function doesn't look very efficient, looping through an array. 

Tell me if what you wanted is this:

```
np.around(s).astype(np.int)
```

reco_utils/dataset/numpy_splitters.py
```diff
@@ -0,0 +1,131 @@
...
+        for i in range(tst[u]):
+            sub_el = random.choice(idx)
+            idx.remove(sub_el)
+            idx_tst.append(sub_el)
+
+        Xtr[u, idx_tst] = 0  #change the selected rated movies to unrated in the train set
+        Xtst[u, idx] = 0  #set the movies that appear already in the train set as 0
+
+        #verify that the number or ratings is conserved by the split function
+        assert(np.sum(Xtr[u,:] != 0) + np.sum(Xtst[u,:] !=0) == rated[u])
```
miguelgfierro: it is a good practice to throw exception instead of using asserts in the code

reco_utils/dataset/numpy_splitters.py
```diff
@@ -0,0 +1,131 @@
...
+
+    #find the number of rated movies per user
+    rated = np.sum(Xtr !=0, axis=1)
+
+    #for each user, cut down a test_size% for the test set
+    tst = round( (rated*test_cut)/100 )
+
+    Nusers, Nitems = X.shape #total number of users and items
+
+    for u in range(Nusers):
```
miguelgfierro: there is a double for loop, is there a way of doing this more efficiently using slices? We can create an issue if you don't see a quick solution and solve it in another PR

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,207 @@
...
+
+        '''
+        Generate the user/item index
+
+        Returns:
+            map_users, map_items: dictionaries mapping the original user/item index to matrix indices
+            map_back_users, map_back_items: dictionaries to map back the matrix elements to the original
+            dataframe indices
+
+        Basic mechanics:
```
miguelgfierro: This should go before Returns

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,207 @@
...
+        log.info("Generating the user/item affinity matrix...")
+
+        self.gen_index()
+
+        r_ = self.df_[self.col_rating]    #ratings
+        itm_id = self.df_['hashedItems']  #itm_id serving as columns
+        usr_id = self.df_['hashedUsers']  #usr_id serving as rows
+
+        #check that all 3 vectors have the same dimensions
+        assert((usr_id.shape[0]== r_.shape[0]) & (itm_id.shape[0] == r_.shape[0]))
```
miguelgfierro: Same here, it is more common to throw exceptions

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,207 @@
...
+    def __init__(
+        self,
+        DF,
+        col_user = DEFAULT_USER_COL,
+        col_item = DEFAULT_ITEM_COL,
+        col_rating = DEFAULT_RATING_COL,
+        col_time = DEFAULT_TIMESTAMP_COL,
+        save_model = False,
+        save_path = 'saver',
+        debug = False,
```
miguelgfierro: It looks we are not using debug anywhere

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,207 @@
...
+
+        self.df = DF #dataframe
+
+        #pandas DF parameters
+        self.col_rating = col_rating
+        self.col_item = col_item
+        self.col_user = col_user
+
+        #Options to save the model for future use
+        self.save_model_ = save_model
```
miguelgfierro: Not sure why the extra `_` in `self.save_model_`, we can do directly `self.save_model`. 
Optionally, a trick to do what you want to do, is to have just one variable, `save_path`, initialized to None. Then you do:
```
if self.save_path is not None:
     n p.save(...)
``` 
this way you are saving one variable and reducing complexity

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,207 @@
...
+        #generate a sparse matrix representation using scipy's coo_matrix and convert to array format
+        self.AM = coo_matrix((r_, (usr_id, itm_id)), shape= (self.Nusers, self.Nitems)).toarray()
+
+        #---------------------print the degree of sparsness of the matrix------------------------------
+
+        zero   = (self.AM == 0).sum() # number of unrated items
+        total  = self.AM.shape[0]*self.AM.shape[1] #number of elements in the matrix
+        sparsness = zero/total *100 #Percentage of zeros in the matrix
+
+        print('Matrix generated, sparsness: %d' %sparsness,'%', 'size:', (self.AM.shape) )
```
miguelgfierro: We could use the log here

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+from reco_utils.recommender.rbm import (
+    HIDDEN,
+    KEEP_PROB,
+    STDV,
+    ALPHA,
+    MINIBATCH,
+    EPOCHS,
+    #MOMENTUM,
+    DEFAULTPATH,
+    _user_item_return_type,
```
miguelgfierro: a private function should never be imported, either make it public or put it in the same file

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+        return h_sampled
+
+    #Multinomial sampling
+    def M_sampling(self, pr):
+
+        '''
+        Multinomial Sampling of ratings
+
+        Args:
+            pr:  a distributions of shape (m, n, r), where m is the number of examples, n the number
```
miguelgfierro: In most arguments, the type of it is missing

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+                 of features and r the number of classes. pr needs to be normalized, i.e.
+                 sum_k p(k) = 1 for all m, at fixed n.
+            f:   normalized, uniform probability used for comparison.
+
+        Returns:
+            v_samp: an (m,n) array of sampled rankings from 1 to r .
+
+        Basic mechanics:
```
miguelgfierro: here again, this should be before returns. Please take a look at google docstring format https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+
+        g = np.random.uniform(size= pr.shape[2]) #sample from a uniform distribution
+        f = tf.convert_to_tensor(g/g.sum(), dtype= tf.float32 )#normalize and convert to tensor
+
+        samp = tf.nn.relu(tf.sign(pr - f) ) #apply rejection method
+        v_samp = tf.cast( tf.argmax(samp , axis=2)+1, 'float32') #select sampled element
+
+        return v_samp
+
+    #Multinomial distribution
```
miguelgfierro: One example to follow google docstrings would be this:
```
     def Pm(self, phi): 
         """ Multinomial distribution. Probability that unit v has value l given phi: P(v=l|phi).
         Args:
             phi (np.array): Linear... 
```
Not sure whether the input is a np.array or a tf tensor...

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+
+        num = [tf.exp(tf.multiply(tf.constant(k, dtype='float32'),phi)) for k in range(1,self.r_+1)]
+        den = tf.reduce_sum(num, axis=0)
+
+        pr = tf.div(num, den)
+
+        return tf.transpose(pr, perm= [1,2,0])
+
+    #Free energy
+    def Fv(self, x):
```
miguelgfierro: I would recommend to rename these functions to something more descriptive, maybe `free_energy(...)`

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+            #Metrics
+            Mserr  = self.msr_error(self.v_k)
+            Clacc  = self.accuracy(self.v_k)
+
+        if self.save_model_: #save the model to file
+            saver = tf.train.Saver()
+
+        init_g = tf.global_variables_initializer() #Initialize all variables in the graph
+
+        #Config GPU memory
```
miguelgfierro: I would give the option to overwrite this, instead of defining it here. That way the user can control these parameters

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+            log.info("done training, Training time %f2" %elapsed)
+
+            #Display training error as a function of epochs
+            plt.plot(Mse_train, label= 'train')
+            plt.ylabel('msr_error', size= 'x-large')
+            plt.xlabel('epochs', size = 'x-large')
+            plt.legend(ncol=1)
+
+            #Final precision scores
+            print('Train set accuracy', precision_train)
```
miguelgfierro: I would use logs instead of prints

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,946 @@
...
+
+            self.sess.run(self.iter.initializer, feed_dict={self.vu: xtst, self.batch_size_: xtst.shape[0]})
+            precision_test = self.sess.run(Clacc)
+            rmse_test = self.sess.run(Mserr)
+
+            elapsed = self.time()
+
+            log.info("done training, Training time %f2" %elapsed)
+
+            #Display training error as a function of epochs
```
miguelgfierro: I think we should extract the plots from the fit, in case the user doesn't want to plot anything

reco_utils/dataset/numpy_splitters.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
WessZumino: Thanks for the comment. The splitter does not use TF, only the RBM algo. For the latter, the TF version is printed when initializing the model, i.e. `model = RBM( )`. Let me know if this answer your question.

reco_utils/dataset/numpy_splitters.py
```diff
@@ -0,0 +1,131 @@
...
+            t = np.floor(i)
+        else: t = np.ceil(i)
+
+        a.append(t)
+
+    a = np.array(a, dtype = np.int32)
+    return a
+
+
+def random_stratified(X, ratio= 0.75, seed= 123):
```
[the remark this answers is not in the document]
WessZumino: yes, no problem for me. I can rename it numpy_stratified_split

reco_utils/dataset/sparse.py
WessZumino: good tips, thanks!

reco_utils/dataset/sparse.py
WessZumino: you mean that `basic mechanics` should go before `Returns`? If this is the case, I need to change this everywhere. Let me know if this is what you meant.

reco_utils/dataset/sparse.py
WessZumino: good idea, thanks !

reco_utils/dataset/sparse.py
WessZumino: it is used in functions like the `cd_protocol` to check how the k_step is changing during learning

reco_utils/dataset/sparse.py
WessZumino: Do you mean that `Basic mechanics` should go before `Returns` ? If this is the case, I should change it everywhere

reco_utils/recommender/rbm/rbm.py
WessZumino: sure, I will reformat it in this way. Personally I find it more logic in this way: first define the variables than describe how it works, otherwise in the description you need to descfribe undefined objects.

reco_utils/recommender/rbm/rbm.py
WessZumino: Sorry, I am not sure I understood what you meant here.

reco_utils/recommender/rbm/rbm.py
WessZumino: yes, the plot and metrics are already optional. By default they are not evaluated/shown unless one uses `with_metrics = True` in the model initialization. Without metrics the training is slighlty faster; In the notebooks I explain that this parameter is optional and I recommend to print metrics to inspect learning.

notebooks/02_model/rbm_deep_dive.ipynb
WessZumino: thanks @yueguoguo. Actually it does visualize well in the jupyter notebook also from other locations other than the dsvm. I often open it on my local machines (laptops, workstations) and still works fine. I suspect it is a problem with git. An easy solution would be  to load the image files on git, as done here 

https://github.com/Microsoft/Recommenders/tree/master/notebooks/04_operationalize

reco_utils/recommender/rbm/rbm.py
WessZumino: looking at the link you sent, I am not sure which option to compare to. For example, the format used in the example below seems similar to mine 

`def` example_generator(n):
    """Generators have a ``Yields`` section instead of a ``Returns`` section.

    Args:
        n (int): The upper limit of the range to generate, from 0 to `n` - 1.

    Yields:
        int: The next number in the range of 0 to `n` - 1.

    Examples:
        Examples should be written in doctest format, and should illustrate how
        to use the function.

        >>> print([i for i in example_generator(4)])
        [0, 1, 2, 3]'

Where Yields --> Returns and Examples --> Basic mechanics, that explains how the function is used.

reco_utils/recommender/rbm/rbm.py
WessZumino: yes I can add the type. The reason for omitting is that they are all float32 tensors, so I thought to save space in this way.

reco_utils/dataset/numpy_splitters.py
WessZumino: yes, this does the job. I use a different convention for 0.5 (round to lower int) but it is just a convention and it does not matter much. I will change the code accordingly. Thanks!

notebooks/02_model/rbm_deep_dive.ipynb
miguelgfierro: Aaron added an image folder in 02_model: https://github.com/Microsoft/Recommenders/tree/staging/notebooks/02_model/img. One option would be to put it there. 

Another option would be to host all the images in a blob and link the images from there.

notebooks/02_model/rbm_deep_dive.ipynb
miguelgfierro: +1 to what @yueguoguo said about creating a new container. We have a blob for the repo project called recodatasets, I just created a blob called images. Please let me know if you find it

reco_utils/dataset/sparse.py
miguelgfierro: yes exactly

reco_utils/recommender/rbm/rbm.py
miguelgfierro: This code:
```
Config = tf.ConfigProto(log_device_placement = True, allow_soft_placement=True)
Config.gpu_options.per_process_gpu_memory_fraction = 0.5
```
is configuring parts of the process, right now if I want to use more than 50% of the GPU memory I can't do it, because it is fixed in the code. If instead we add these parameters as inputs, the programmer can choose the usage of the GPU programatically.

reco_utils/dataset/numpy_splitters.py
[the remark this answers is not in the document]
miguelgfierro: Sorry, I put the comment in the wrong place, for some reason I couldn't comment on the notebook. 
In the NCF notebook, Aaron added TF in the notebook, please take a look at this: https://github.com/Microsoft/Recommenders/blob/staging/notebooks/02_model/ncf_deep_dive.ipynb.

reco_utils/recommender/rbm/rbm.py
```diff
@@ -0,0 +1,973 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: The rbm tests are missing. This is how we did the tests in SAR: https://github.com/Microsoft/Recommenders/blob/staging/tests/unit/test_sar_singlenode.py.
Let me know if you need help on how to structure the tests

reco_utils/recommender/rbm/rbm.py
WessZumino: ok got it. That was one "experiment", thanks for bringing it to my attention. The reason is that I experience GPU memory problems when using the 10m dataset. I have also tried to allow TF to increase the memory allocation dynamically but it did not solve the problem. An easy solution would be to move to a DSVM with more memory (currently using the smallest available). I will open an issue if you think it is worth spending some time on this.

notebooks/02_model/rbm_deep_dive.ipynb
WessZumino: ok I fixed this as well, there was a misplaced $ sign ... I wonder why it worked in the Jupyter notebook :s

reco_utils/dataset/numpy_splitters.py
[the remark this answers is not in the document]
WessZumino: I see what you mean. The point is that TF is not used explictly in the notebook as it is loaded only in the RBM class. I could load it in the notebook as well but this makes the first cell even slower to load. If you think printing the version here instead of at the moment the algo runs is better, I will change it accordingly.

reco_utils/recommender/rbm/rbm.py
WessZumino: I changed this back to 

`
Config = tf.ConfigProto(log_device_placement=True, allow_soft_placement=True)
Config.gpu_options.allow_growth = True  # dynamics memory allocation
`
In this way memory allocation is dynamical and the user does not need to modify anything.

reco_utils/recommender/rbm/rbm.py
WessZumino: Thanks, my question is as follows: The class contains several methods, do they need individual tests? If so, one probably need to use TF eager execution or create several sessions inside the test. I cannot find the unit test for the ncf algo that also uses TF, there is only `test_ncf_dataset`. It would be good to see how the test has been written in that case.

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1160 @@
+{
```
miguelgfierro: Small comment here, can you move all the figures to the common blob that we have? https://recodatasets.blob.core.windows.net/images the credentials are in the wiki on teams

notebooks/00_quick_start/rbm_movielens.ipynb
```diff
@@ -0,0 +1,806 @@
...
+    "data.head()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 1.2 Split the data using the stratified splitter  \n",
+    "\n",
+    "As a second step we generate the user/item affiity matrix and then split the data into train and test set. If you are familiar with training supervised learning model, here you will notice the first difference. In the former case, we cut off a certain proportion of training examples from dataset (e.g. images), here corresponding to users (or items), ending up with two matrices (train and test) having different row dimensions. Here we need to mantain the same matrix size for the train and test set, but the two will contain different amounts of ratings, see the [deep dive notebook](../02_model/rbm_deep_dive.ipynb) for more details. The affinity matrix reads     "
```
anargyri: Maybe skip the supervised learning part? I assume readers of the repo are at least familiar with the concept of collaborative filtering.

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Args:
+            DF: A dataframe containing at least UserID, ItemID, Ratings
+
+        Returns:
+            RM: user-affinity matrix of dimensions (Nusers, Nitems) in numpy format. Unrated movies
+            are assigned a value of 0.
+
+        Basic mechanics:
+            As a firts step, two new columns are added to the input DF, containing the index maps
```
anargyri: "first"

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,204 @@
...
+    DEFAULT_RATING_COL,
+    DEFAULT_TIMESTAMP_COL,
+    PREDICTION_COL,
+)
+
+# for logging
+log = logging.getLogger(__name__)
+
+
+class AffinityMatrix:
```
anargyri: Nice job! This looks similar to the reindexing in the SAR notebooks; maybe we could replace that too.

reco_utils/recommender/rbm/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+DEFAULTPATH = 'saver'
+
+def _user_item_return_type():
+    return float
+
+
+def _predict_column_type():
+    return float
+
+def _predict_column_type():
```
anargyri: This looks like a duplicate. @miguelgfierro @yueguoguo  How did this pass lint or black?

tests/unit/test_numpy_splitters.py
```diff
@@ -0,0 +1,102 @@
...
+    X = np.random.randint(
+        low=0,
+        high=6,
+        size=(test_specs["number_of_users"], test_specs["number_of_items"]),
+    )
+
+    return X
+
+
+def test_random_stratified_splitter(test_specs, python_dataset):
```
anargyri: Why "random" and not "numpy"? (you are testing `numpy_stratified_split()`)

tests/unit/test_numpy_splitters.py
```diff
@@ -0,0 +1,102 @@
...
+    assert (Xtr.shape[0] == X.shape[0]) & (Xtr.shape[1] == X.shape[1])
+
+    assert (Xtst.shape[0] == X.shape[0]) & (Xtst.shape[1] == X.shape[1])
+
+    X_rated = np.sum(X != 0, axis=1)  # number of total rated items per user
+    Xtr_rated = np.sum(Xtr != 0, axis=1)  # number of rated items in the train set
+    Xtst_rated = np.sum(Xtst != 0, axis=1)  # number of rated items in the test set
+
+    # global split: check that the all dataset is split in the correct ratio
+    assert (
```
anargyri: Couldn't you use `pytest.approx()` for these as in other tests?

reco_utils/recommender/rbm/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+ALPHA = 0.004
+#number of minibatches
+MINIBATCH = 100
+#number of epochs for the optimization
+EPOCHS = 20
+
+#Set default directory for saving files
+DEFAULTPATH = 'saver'
+
+def _user_item_return_type():
```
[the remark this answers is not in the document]
WessZumino: no we are not, I forgot to remove them when I moved all the inits in the min algo. Good catch, thanks!

reco_utils/recommender/rbm/__init__.py
WessZumino: Thanks, I removed all those functions from the init as I was not using them anymore, as pointed out by @miguelgfierro

reco_utils/dataset/sparse.py
WessZumino: yes, this is where I took it from :) I just thought that adding it into this function would have been easier for the user, as he/she doesn't need to implement it explicitly in the notebook.

reco_utils/dataset/sparse.py
WessZumino: thanks, fixed it.

tests/unit/test_numpy_splitters.py
WessZumino: right, the name referred to the previous nomenclature and I forgot to change it. Thanks for catching this

tests/unit/test_numpy_splitters.py
WessZumino: Hi Andreas, I have added back the `pytest.approx()` method for the test of the global split. On the other hand, this method does not work with the "local" test, where the assert is over an array. This was the original reason for not using the `pytest.approx()` method. Nevertheless, there may be a simple way to make `pytest.approx()` works for an array, please let me know if you have any suggestions.

notebooks/00_quick_start/rbm_movielens.ipynb
```diff
@@ -0,0 +1,806 @@
...
+    "\n",
+    "### Advantages of RBM: \n",
+    "\n",
+    "The model generates ratings for a user/movie pair using a collaborative filtering based approach. While matrix factorization methods learn how to reproduce an instance of the user/item affinity matrix, the RBM learns the underlying probability distribution. This has several advantages: \n",
+    "\n",
+    "- Generalizability : the model generalize well to new examples.\n",
+    "- Stability in time: if the recommendation task is time-stationary, the model does not need to be trained often to accomodate new ratings/users. \n",
+    "- Scales well with the size of the dataset and the sparsity of the user/affinity matrix. \n",
```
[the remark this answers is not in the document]
WessZumino: I meant the dimension of the matrix. Nevertheless, I did not finish the scaling analysis, so it is better to remove this sentence.

notebooks/00_quick_start/rbm_movielens.ipynb
WessZumino: yep, removed

tests/unit/test_numpy_splitters.py
anargyri: For scalar values we use `pytest.approx()` a lot. For arrays, I think it should work too 
https://docs.pytest.org/en/latest/reference.html#pytest-approx
but I don't recall if we have used it.

notebooks/00_quick_start/rbm_movielens.ipynb
```diff
@@ -0,0 +1,803 @@
...
+    }
+   ],
+   "source": [
+    "#Model Fit\n",
+    "train_time= model.fit(Xtr, Xtst)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
```
anargyri: The label of the y-axis is not defined. What is msr_error?

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1154 @@
...
+    "\n",
+    "This notebook is organized as follows:\n",
+    "\n",
+    "1. RBM Theory \n",
+    "2. Tensorflow implementation and model parameters  \n",
+    "3. Data preparation and inspection\n",
+    "4. Model application, performance and analysis  \n",
+    "\n",
+    "Sections 1 and 2 require basic knowledge of linear algebra, probability theory and tensorflow while  \n",
+    "sections 3 and 4 only require some basic data science understanding. **Feel free to jumpo to the section you are most interested in!**"
```
anargyri: "jump"

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1154 @@
...
+    "print(\"Pandas version: {}\".format(pd.__version__))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1. RBM Theory \n",
+    "\n",
+    "## 1.1 Overview and main differences with other recommender algorithms\n",
```
anargyri: with -> from

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1154 @@
...
+    "\n",
+    "Being designed to handle and organize a large amount of data, SM finds ideal applications in modern learning algorithms. In the context of **recommender systems**, the idea is to learn typical user behaviour instead of particular instances. To better understand this consider the most general setup of a recmmendation problem: there are $m$ users rating $n$ items according to some scale (e.g. 1 to 5). In a typical scenario of online shopping, streaming services or decision processes, the user only rates a subset $l \\ll m$ of the products. If we now create a matrix representation of this problem, we obtain the user/item affinity matrix $X$. In a more readable table form, $X$ will look like this:\n",
+    "\n",
+    "\n",
+    "|  $X$   |$i_1$  |$i_2$  |$i_3$  |  ... |$i_m$  | \n",
+    "|-----|-------|-------|-------|------|-------|\n",
+    "|$u_1$|5      |0      |2      |0 ... |1      |\n",
```
anargyri: Better use Nans or empty instead of zeros.

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1154 @@
...
+    "\n",
+    "$$ F =- \\log Z =- \\log \\sum_{ v_i, h_i } P(v, h) $$.\n",
+    "\n",
+    "In the language of probability theory, the above quantity is the cumulant generating function. One way of evaluating the free energy is to use a [Markov-chain Montecarlo sampling](https://en.wikipedia.org/wiki/Monte_Carlo_method#Computer_graphics) algorithm such as the Metropolis-Hasting; here we will use instead an approximate method called Contrastive divergence, based on [Gibbs sampling](https://en.wikipedia.org/wiki/Gibbs_sampling) (see below). The latter has the advantage of being faster than Montecarlo. Once the candidate $F$ has been found, we fix the learning parameters by extremizing $F$. Let us see how this works in practice in the next section. "
```
anargyri: extremizing -> minimizing

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1154 @@
...
+    "In this section we briefly describe how the algorithm is implemented in Tensorflow and which parameters can be customized by the user during training. We also discuss some best practices to be used when training the RBM model on a recommendation task. Further technical details are explained directly in the code. \n",
+    "\n",
+    "Tensorflow (TF) is an open source framework to develope deep learning (DL) models in a fast and efficient way. One of the shared characteristics of DL frameworks is autodifferentiation, i.e. the symbolic evaluation of gradients, that will be particulary useful here. The other advantage of TF is the geration and optimization of the symbolic operations defined on a computational graph, for fast and scalable deployment on CPU and GPU. For more informations on TF see [here](www.tensorflow.com). Unfortunately, TF is tailor made for supervised learning tasks, so its application to unsupervised model needs some more work. Note: although TF has recently started developing a [set of libraries to perform probabilistic inference](www.tensorflow_probability.com), we found their performance still not optimal and therefore we will not use them here. \n",
```
anargyri: "develop"

anargyri: "generation"

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1154 @@
...
+    "\n",
+    "\n",
+    "The RBM model is instantiated as a class with several methods to build the graph, perform sampling, training and inference. The skeleton of the graph is built at the moment the class is instantiated; mandatory fields are: \n",
+    "\n",
+    "- `hidden_units` integer (Default =500) : number of hidden units\n",
```
anargyri: These need some spaces to show better.

notebooks/02_model/rbm_deep_dive.ipynb
```diff
@@ -0,0 +1,1154 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 3.1 Split the data using the stratified splitter  \n",
+    "\n",
+    "As a second step, we split the data into train and test set by mantaining the same matrix size. Clearly, the two matrices will contain different ratings in different proportions.  \n",
+    "\n",
+    "- First, we use the `AffinityMatrix` class to generate the $(m,n)$ user/affinity matrix $X$ defined in section **1.1**; this also returns the sparsness percentage. For example, for the 1m dataset, $95$ % of the matrix entries are zeros. This represents a challenge for the learning task: fixing $95$ % of entries with only $5$ % of data points. \n",
```
anargyri: "sparseness"

anargyri:
These notebooks look very nice! Great job!

notebooks/02_model/rbm_deep_dive.ipynb
WessZumino: Thank I will change it to minimising. What I meant is that we are only evaluating the first order derivative, so we are only searching for optima; only looking at the Hessian we know the nature of the optima. In most of the cases in ML there is no guarantee we reach the minimum as the problem is non-convex and most likely we end up in some low error saddle point. But yes, at the end our intention is to minimise the cost function, so the second term is more appropriate in this context.

notebooks/00_quick_start/rbm_movielens.ipynb
WessZumino: sorry, I just noticed that we use rmse instead, I will change it accordingly.

WessZumino:
Thanks @anargyri for carefully reading the notebooks and the suggested fixtures/improvements!

tests/unit/test_numpy_splitters.py
WessZumino: Thanks for the link, I found the problem and now it works with `pytest.approx()` both for scalars and arrays :)

WessZumino:
I have added the unit tests for the rbm, thank you @miguelgfierro @anargyri and @yueguoguo for all the great suggestions, fixtures and improvements!

Issue #297, opened 2018-11-27
Hyperdrive for movielens with one algo


anargyri:
### *What* is affected by this?
Benchmarking results in Readme and notebooks for the algos.

### Expected behavior (i.e. solution)
Get optimal hyperparameters for all algos by exhaustive grid search. Do this only for Movielens100K and 1M in order to limit computation time.

anargyri:
@yueguoguo has made some good points in the discussion today:
- The various methodologies for hyperparameter tuning may not be well adapted to recommendations problems when we wish to optimize ranking metrics, especially to the loss involved in Bayesian personalized ranking. 
- Hyperdrive uses multi-arm bandits, which are more online in spirit. Bayesian optimization approaches also use exploration - exploitation.
- There are some researchers working on optimizing BPR loss in the case of ranking metrics.  

My comments are that: 
- We have not tried BPR to know how scalable and efficient it is. Spotlight has an implementation https://maciejkula.github.io/spotlight/losses.html?highlight=bpr#spotlight.losses.bpr_loss but this is in Pytorch, and we need to check efficiency first. It also applies only to matrix factorization. I haven't found any BPR implementation on Spark.
- It seems that Hyperdrive can work as a black box and so, we can pass to it a python script that uses e.g. the Spotlight code above. Of course, hyperparam search will not be informed by the nature of the method. One advantage of Hyperdrive that I see is the parallelization is done automatically, which would probably not be the case with a third-party method from academia. I.e. they have done the dirty work for us.
- If we implement ourselves a regression based approach for tuning it could in fact work better than bandits or Bayesian tuning. However we would need to also implement the parallelization and would lose the benefits offered by AML.

PR #401, opened 2019-01-07
Spark recommender hyperparameter tuning

yueguoguo:
### Description
Hyperparameter tuning for Spark recommender. 

### Motivation and Context
Hyperparameter tuning plays a vital role in finding the optimal model.  

### Releated Issues
#177 #297 

### Checklist:

- [ ] I have added tests.


notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "import numpy as np\n",
+    "import time\n",
+    "\n",
+    "import pyspark\n",
+    "from pyspark import SparkContext, SparkConf\n",
+    "from pyspark.ml.recommendation import ALS\n",
+    "from pyspark.ml.tuning import ParamGridBuilder, TrainValidationSplit\n",
+    "from pyspark.ml.evaluation import RegressionEvaluator\n",
+    "\n",
+    "from hyperopt import fmin, tpe, hp, STATUS_OK, Trials\n",
```
miguelgfierro: Do we need to add this hypteropt to the conda file?

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "source": [
+    "# Hyperparameter tuning (Spark based recommender)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Hyperparameter tuning for Spark based recommender algorithm is important to select a model with the optimal performance. This notebook introduces good practices in performing hyperparameter tuning for building recommender models with the utility functions provided in the [Microsoft/Recommenders](https://github.com/Microsoft/Recommenders.git) repository."
+   ]
```
miguelgfierro: I would add a paragraph to explain briefly what is the notebook about, especially telling that we are going to use 2 hyperparameter methods. The first time I read the notebook I was a little bit lost.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "metadata": {},
+   "source": [
+    "## 2 Hyper parameter tuning with Azure Machine Learning Services"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `hyperdrive` module in the [Azure Machine Learning Services](https://azure.microsoft.com/en-us/services/machine-learning-service/) is supposed to run [hyperparameter tuning and optimizing for machine learning model selection](https://docs.microsoft.com/en-us/azure/machine-learning/service/how-to-tune-hyperparameters). At the moment, the service supports running hyperparameter tuning on heterogenous computing targets such as cluster of commodity compute nodes with or without GPU devices (see detailed documentation [here](https://docs.microsoft.com/en-us/azure/machine-learning/service/how-to-set-up-training-targets)). It is feasible to run parameter tuning on a cluster of VM nodes. In this case, the service containerizes individual and independent Spark session on each node of the cluster to run the parameter tuning job in parallel, instead of inside a single Spark session where the training is executed in a distributed manner.  \n",
```
miguelgfierro: This section is weird, we are talking about a library that we are not using and that will be available in the future (but when?). I would vote to remove it or push it to the bottom after the other 2 hyperparameter methods

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+       " 2.786421555547063]"
+      ]
+     },
+     "execution_count": 12,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "model.validationMetrics"
```
miguelgfierro: What metric is this, rmse?

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# define a search space\n",
+    "space = {\n",
+    "    'rank': hp.quniform('rank', 10, 20, 5),\n",
```
miguelgfierro: In this one we are checking 10, 15, 20. In the previous section, are we also checking this or only 10 and 20?

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "        implicitPrefs=False,\n",
+    "        alpha=0.1,\n",
+    "        regParam=reg,\n",
+    "        coldStartStrategy='drop',\n",
+    "        nonnegative=False,\n",
+    "        seed=0,\n",
+    "        **HEADER_ALS\n",
+    "    )\n",
+    "    \n",
+    "    model = als.fit(train)\n",
```
miguelgfierro: Here we are using train (60% of the original data) and valid (20%). In the previous method, `TrainValidationSplit` is instantiated to use train, which is splitted internally in 80%-20%. Is this correct? If so, in the first method you should input train+valid.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Customize an objective function\n",
+    "def objective(params):\n",
```
miguelgfierro: I would try to follow here single responsibility. I would suggest:
`def objective(params, model, dataset)` or alternatively `def objective(params, model, train, valid)`. At a high level, this function takes a model and a dataset and find the best parameters.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "`fmin` of `hyperopt` is used for running the trials for searching optimal hyper parameters. In `hyperopt`, there are different strategies for intelligently optimize hyper parameters. For example, `hyperopt` avails Tree of Parzen Estimators (TPE) method for searching optimal parameters, which is more efficient than random search or grid search (details can be found [here](http://hyperopt.github.io/hyperopt/)). It is claimed in the project website that Bayesian based optimization method is planned but so far it has not yet been implemented. \n",
+    "\n",
+    "The following runs the trials with the pre-defined objective function and search space. TPE is used as the optimization method. Totally there will be 10 evaluations run for searching the best parameters."
```
miguelgfierro: In `details can be found [here](http://hyperopt.github.io/hyperopt/)` we are linking to the main page of hyperopt. I couldn't find a description of TPE on their page, which is odd. Maybe we can do a brief description of TPE and link to some source like wikipedia or other?

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "outputs": [],
+   "source": [
+    "model_best_spark = model.bestModel"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 3 Hyperparameter tuning with `hyperopt` in sequential mode"
```
miguelgfierro: According to the name in here https://github.com/hyperopt/hyperopt/wiki, Hyperopt is Distributed asynchronous algorithm configuration / hyperparameter optimization. I don't understand why we are using it in `sequential mode`

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    ")\n",
+    "    \n",
+    "model_best_hyperopt = als.fit(train)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Tuning prameters against other metrics can be simply done by modifying the `objective` function. The following shows an objective function of how to tune \"precision@k\". Since `fmin` in `hyperopt` only supports minimization while the actual objective of the loss is to maximize \"precision@k\", `-precision` instead of `precision` is used in the returned value of the `objective` function."
```
miguelgfierro: I think this is super powerful. This allows optimizing for rating or for ranking independently

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "Sample the parameters used for model building from the pre-defined space. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 20,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import hyperopt.pyll.stochastic\n",
```
miguelgfierro: In other notebooks we put all the imports in the top cell

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "metadata": {},
+   "source": [
+    "# Conclusions"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In summary, there are mainly three different approaches for running hyperparameter tuning for Spark based recommendation algorithm. \"How-to\" guides and Pros and Cons are tabulated below."
```
miguelgfierro: I really like this format for conclusions, it states pros and cons of each method. Maybe we should follow this pattern in other notebooks?

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+      ]
+     },
+     "execution_count": 11,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "model.explainParams()"
```
gramhagen: the output directly from this model is pretty ugly, but it can be cleaned up a bit and merged with validation metrics like this:
```
for idx, item in enumerate(model.getEstimatorParamMaps()):
    print('Run {}:'.format(idx))
    print('\tValidation Metric: {}'.format(model.validationMetrics[idx]))
    for key, value in item.items():
        print('\t{0}: {1}'.format(repr(key), value))
```

notebooks/02_model/hypertune_spark_deep_dive.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "metadata": {},
+   "source": [
+    "# Hyperparameter tuning (Spark based recommender)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Hyperparameter tuning for Spark based recommender algorithm is important to select a model with the optimal performance. This notebook introduces good practices in performing hyperparameter tuning for building recommender models with the utility functions provided in the [Microsoft/Recommenders](https://github.com/Microsoft/Recommenders.git) repository."
```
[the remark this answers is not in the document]
gramhagen: I might say something like this:
```
Hyperparameter tuning is an important step to selecting a model with optimal performance. 
This notebook provides examples of performing hyperparameter tuning when training recommender algorithms on Spark using utility functions in the [Microsoft/Recommenders](https://github.com/Microsoft/Recommenders.git) repository.
```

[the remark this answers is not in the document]
yueguoguo: Much better than Grammarly 👍  Thanks both

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: Removed.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: Added reference.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: Ah this was not made clear. `hyperopt` does not support distribution on Spark.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: Agree and this is what we want. Support custom loss function is a very important feature to have.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: We changed the deep dive notebooks structure a little bit. E.g., in the reco algorithm notebooks, we put theoretical introduction at the beginning suggested by @wutaomsft.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: Good idea. If there is no comparative study, a general conclusion is great to have as key takeaways of the notebook.

yueguoguo:
> looks pretty good to me, there is also a package spark_sklearn that supports sklearn style grid-search, it might be worth considering updating this (in the future) if we can get that version working as I suspect it could simplify some of the pyspark implementation: https://github.com/databricks/spark-sklearn/blob/master/python/spark_sklearn/grid_search.py

This is nice! 

When doing hyper tuning/optimization, what we expect are generally
* Parameter sampling methods.
* Distributed computing capability.
* Parameter searching algorithm. 
So far Spark native constructs and `spark_sklearn` only support grid search, which, IMHO, do not reduce the efforts of finding optimal hyperparameters. I suspect this is challenging work because of the Spark mechanism which runs the actually distributed workloads with Java. One can customize Spark interfaces (e.g., Spark `Estimator`, `Evaluator`, etc.) to allow this feature.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: 👍

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: Good catch. I modified the split ratio.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: It is a little bit against SRP. I would do the same but there is some restriction in parsing arguments to `objective` when calling `fmin` in `hyperopt` for tuning.

notebooks/02_model/hypertune_spark_deep_dive.ipynb
yueguoguo: I changed the section title and added some text to make it more clear.

yueguoguo:
Discussed with @loomlike about the location of the hyper tune notebooks. We thought it would be good to create a new folder (like now) for hyper tune specific topics. 

Any ideas on this?

miguelgfierro:
@yueguoguo one minor comment about the folder is that parameter tunning would come before o16n, so maybe we should swap the folder numbers

yueguoguo:
> @yueguoguo one minor comment about the folder is that parameter tunning would come before o16n, so maybe we should swap the folder numbers.

@miguelgfierro make sense.

Issue #189, opened 2018-11-01
Set the GPU environment 


miguelgfierro:
related to https://github.com/Microsoft/Recommenders/issues/188

loomlike:
related to #

miguelgfierro:
- [ ] add setup
- [ ] add the tests

PR #408, opened 2019-01-09
Gpu env

miguelgfierro:
### Description

Add GPU env to support new GPU algos

### Releated Issues
#189

### Checklist:

- [ ] I have added tests.


scripts/generate_conda_file.sh
```diff
@@ -90,14 +90,12 @@ ${pyspark}- pyspark==2.3.1
 - fastparquet>=0.1.6
 ${pyspark}- pyarrow>=0.8.0
 - pip:
-  - idna==2.7
```
miguelgfierro: @yueguoguo  @loomlike I think we are not using idna or tffm at the moment, am I right?

yueguoguo: `idna` is a dependency package required by some other packages. Here, the explicit addition of `idna` with a fixed version of 2.7 is because there are conflicts of package versions for certain packages. The consequence of having no fixed version number of `idna` is that it would produce errors in setting up the conda environment in the CI/CD machine. So let's keep it.

miguelgfierro: good, I put the library back

PR #411, opened 2019-01-10
FastAI Recommender

danielsc:
### Description
Adding recommender based on fast.ai library

### Motivation and Context
Fast AI is an interesting library that uses Pytorch under the cover. This is an interesting addition to the current lineup of algorithms in this repo.

### Checklist:

- [x] I have added tests.


notebooks/00_quick_start/fastai_recommendation.ipynb
```diff
@@ -0,0 +1,788 @@
...
+      ]
+     },
+     "execution_count": 52,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# Mean Squared Error\n",
+    "from sklearn.metrics import r2_score, explained_variance_score\n",
```
yueguoguo: We have rating metrics available in the reco_util functions which can be used directly. See [here](https://github.com/Microsoft/Recommenders/blob/master/reco_utils/evaluation/python_evaluation.py) for details.

notebooks/00_quick_start/fastai_recommendation.ipynb
```diff
@@ -0,0 +1,788 @@
...
+       ")"
+      ]
+     },
+     "execution_count": 38,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "learn = collab_learner(data, n_factors=40, y_range=[0,5.5], wd=1e-1)\n",
```
miguelgfierro: I think the range is 1-5 for movielens 100k

notebooks/00_quick_start/fastai_recommendation.ipynb
miguelgfierro: yeah +1 to this

danielsc: ok -- did that

notebooks/00_quick_start/fastai_recommendation.ipynb
danielsc: Although ratings can only range from 1-5, I am are setting the range of possible ratings from 0 to 5.5 -- that will allow the model to predict values to both sides of 1 and 5, which improves accuracy.
I added that as an explanation to the cell above

PR #415, opened 2019-01-11
Test precision at k

miguelgfierro:
### Description

Added some examples to test precision at k is normalized

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/tests/unit/test_python_evaluation.py
+++ b/tests/unit/test_python_evaluation.py
@@ -95,7 +95,6 @@ def python_data():
                 5,
             ],
         }
-
     )
 
     return rating_true, rating_pred, rating_nohit
@@ -173,7 +172,6 @@ def test_python_map_at_k(python_data, target_metrics):
 
 def test_python_precision(python_data, target_metrics):
     rating_true, rating_pred, rating_nohit = python_data
-
     assert (
         precision_at_k(
             k=10,
@@ -185,6 +183,33 @@ def test_python_precision(python_data, target_metrics):
     )
     assert precision_at_k(rating_true, rating_nohit, k=10) == 0.0
     assert precision_at_k(rating_true, rating_pred, k=10) == target_metrics["precision"]
+
+    # check normalization
+    single_user = pd.DataFrame(
+        {"userID": [1, 1, 1], "itemID": [1, 2, 3], "rating": [5, 4, 3]}
+    )
+    assert (
+        precision_at_k(
+            k=3,
+            rating_true=single_user,
+            rating_pred=single_user,
+            col_prediction="rating",
+        )
+        == 1
+    )
+    same_items = pd.DataFrame(
+        {
+            "userID": [1, 1, 1, 2, 2, 2],
+            "itemID": [1, 2, 3, 1, 2, 3],
+            "rating": [5, 4, 3, 5, 5, 3],
+        }
+    )
+    assert (
+        precision_at_k(
+            k=3, rating_true=same_items, rating_pred=same_items, col_prediction="rating"
+        )
+        == 1
+    )
 
 
 def test_python_recall(python_data, target_metrics):
--- a/tests/unit/test_python_evaluation.py
+++ b/tests/unit/test_python_evaluation.py
@@ -184,7 +184,7 @@ def test_python_precision(python_data, target_metrics):
     assert precision_at_k(rating_true, rating_nohit, k=10) == 0.0
     assert precision_at_k(rating_true, rating_pred, k=10) == target_metrics["precision"]
 
-    # check normalization
+    # Check normalization
     single_user = pd.DataFrame(
         {"userID": [1, 1, 1], "itemID": [1, 2, 3], "rating": [5, 4, 3]}
     )
@@ -209,6 +209,15 @@ def test_python_precision(python_data, target_metrics):
             k=3, rating_true=same_items, rating_pred=same_items, col_prediction="rating"
         )
         == 1
+    )
+
+    # Check that if the sample size is smaller than k, the maximum precision can not be 1
+    # if we do precision@5 when there is only 3 items, we can get a maximum of 3/5.
+    assert (
+        precision_at_k(
+            k=5, rating_true=same_items, rating_pred=same_items, col_prediction="rating"
+        )
+        == 0.6
     )
 
 
--- a/tests/unit/test_python_evaluation.py
+++ b/tests/unit/test_python_evaluation.py
@@ -1,12 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-"""
-Test evaluation
-"""
 import pandas as pd
 import pytest
-
 from reco_utils.evaluation.python_evaluation import (
     rmse,
     mae,
--- a/tests/unit/test_spark_evaluation.py
+++ b/tests/unit/test_spark_evaluation.py
@@ -155,7 +155,7 @@ def test_spark_recall(spark_data, target_metrics):
 
 
 @pytest.mark.spark
-def test_spark_precision(spark_data, target_metrics):
+def test_spark_precision(spark_data, target_metrics, spark):
     df_true, df_pred = spark_data
 
     evaluator = SparkRankingEvaluation(df_true, df_pred, k=10)
--- a/tests/unit/test_spark_evaluation.py
+++ b/tests/unit/test_spark_evaluation.py
@@ -184,7 +184,7 @@ def test_spark_precision(spark_data, target_metrics, spark):
         }
     )
     df_same = spark.createDataFrame(same_items)
-    evaluator3 = SparkRankingEvaluation(df_same, df_same, k=5, col_prediction="rating")
+    evaluator3 = SparkRankingEvaluation(df_same, df_same, k=3, col_prediction="rating")
     assert evaluator3.precision_at_k() == 1
 
 
--- a/tests/unit/test_spark_evaluation.py
+++ b/tests/unit/test_spark_evaluation.py
@@ -166,6 +166,27 @@ def test_spark_precision(spark_data, target_metrics):
     )
     assert evaluator1.precision_at_k() == target_metrics["precision"]
 
+    # check normalization
+    single_user = pd.DataFrame(
+        {"userID": [1, 1, 1], "itemID": [1, 2, 3], "rating": [5, 4, 3]}
+    )
+    df_single = spark.createDataFrame(single_user)
+    evaluator2 = SparkRankingEvaluation(
+        df_single, df_single, k=3, col_prediction="rating"
+    )
+    assert evaluator2.precision_at_k() == 1
+
+    same_items = pd.DataFrame(
+        {
+            "userID": [1, 1, 1, 2, 2, 2],
+            "itemID": [1, 2, 3, 1, 2, 3],
+            "rating": [5, 4, 3, 5, 5, 3],
+        }
+    )
+    df_same = spark.createDataFrame(same_items)
+    evaluator3 = SparkRankingEvaluation(df_same, df_same, k=5, col_prediction="rating")
+    assert evaluator3.precision_at_k() == 1
+
 
 @pytest.mark.spark
 def test_spark_ndcg(spark_data, target_metrics):
--- a/tests/unit/test_spark_evaluation.py
+++ b/tests/unit/test_spark_evaluation.py
@@ -166,7 +166,7 @@ def test_spark_precision(spark_data, target_metrics, spark):
     )
     assert evaluator1.precision_at_k() == target_metrics["precision"]
 
-    # check normalization
+    # Check normalization
     single_user = pd.DataFrame(
         {"userID": [1, 1, 1], "itemID": [1, 2, 3], "rating": [5, 4, 3]}
     )
@@ -186,6 +186,11 @@ def test_spark_precision(spark_data, target_metrics, spark):
     df_same = spark.createDataFrame(same_items)
     evaluator3 = SparkRankingEvaluation(df_same, df_same, k=3, col_prediction="rating")
     assert evaluator3.precision_at_k() == 1
+
+    # Check that if the sample size is smaller than k, the maximum precision can not be 1
+    # if we do precision@5 when there is only 3 items, we can get a maximum of 3/5.
+    evaluator4 = SparkRankingEvaluation(df_same, df_same, k=5, col_prediction="rating")
+    assert evaluator4.precision_at_k() == 0.6
 
 
 @pytest.mark.spark


anargyri:
Could you also try precision@10 in order to test the case k > sample size per user?

tests/unit/test_python_evaluation.py
```diff
@@ -186,6 +184,33 @@ def test_python_precision(python_data, target_metrics):
...
+        == 1
+    )
+    same_items = pd.DataFrame(
+        {
+            "userID": [1, 1, 1, 2, 2, 2],
+            "itemID": [1, 2, 3, 1, 2, 3],
+            "rating": [5, 4, 3, 5, 5, 3],
+        }
+    )
+    assert (
```
WessZumino: if k=|D|_i, where  |D|_i is the size of the rated elements, then the normalisation is correct by definition. You can check that the test fails if you use $k=10$. To be fair, this is not quite what may happen in the real case. In that case the predicted set has always k elements by definition while the test set may have less than k.

miguelgfierro: Can you elaborate ’You can check that the test fails if you use $k=10$.'?

WessZumino: Sorry, I meant if you take your code and just change to k=10 the test fails. I tried it explicitly in your code but  it is a trivial check. The less trivial one is the one Andreas mentioned, i.e. having unequal numbers of entries in the test set, with some of them having |D|_i < 10 for some i. Then the predicted set should have 10 rated elements per user. This is what happens in the real dataset.

WessZumino: For example, in the 100k Movielens dataset, with a 25% split data in the test set, 30% of the users have less than 10 ratings. Since precision @10 uses a "hit" normalisation factor of 1/10, it is clear that this 30% of users will lower the maximum precision. It is intrinsic in the definition of these metrics, making them dataset dependent. My suggestion was just to report the reference scale together with the absolute value of the metrics, I am not suggesting to not use these metrics.

miguelgfierro: So what I understand from your latest comment is that you think we should inform the user that in the case of 100k Movielens, where 30% user have less ratings than 10, they will never get the maximum precision. Is that what you are saying?

The main point of this PR is to check whether or not the precision at k is normalized. We saw that it is. As Andreas explain, in some cases (when k> sample size) the maximum precision can be less than 1, but that doesn't mean that the metric is not normalized. @WessZumino do you agree with this?

anargyri: > So what I understand from your latest comment is that you think we should inform the user that in the case of 100k Movielens, where 30% user have less ratings than 10, they will never get the maximum precision. Is that what you are saying?
> 
> The main point of this PR is to check whether or not the precision at k is normalized. We saw that it is. As Andreas explain, in some cases (when k> sample size) the maximum precision can be less than 1, but that doesn't mean that the metric is not normalized. @WessZumino do you agree with this?

No, this is not what should happen. If k > sample size the maximum precision equals 1 and is achieved when the ground truth and the recommendations are identical for each user.

WessZumino: Sorry for the late answer. Well, the usual meaning of normalisation is that the sum over all realisations is equal to 1. So if the maximum precision can be less than one, it  is like saying that the probability of getting all right is less than one, hence that the probability is not normalised. 
Note that ranking metrics are, mathematically, distributions evaluated on empirical data, and therefore follow the two fundamental laws of probability: 

\sum_i p_i = 1 (normalisation) 
p_i \ge 0 for all i (non-negativity) 

It is not about informing the user that they will never achieve 100%, it is about specifying a scale. Apart from pure math, all numbers we use in science and engineering are not pure, meaning they have meaning only if defined relative to another number that serves as a reference ( a scale). For convenience, this number is often chosen to be 1, but this is arbitrary. Saying you achieved 0.6 does not mean anything if you do not introduce a scale, e.g. 0.6/1 or 0.6/0.9. 

I have created a notebook to make this as clear as possible 

https://github.com/Microsoft/Recommenders/blob/test_normalizations/notebooks/03_evaluate/test_normalizations.ipynb

In my opinion, it would make the repo more scientifically rigorous to give the reference scale together with the ranking metrics. In the notebook I have introduced a very simple function that returns the maximum achievable precision @k .

anargyri: Thanks for the notebook and the explanation. However, the issue is not mathematical in the current context (offline evaluation for recommendations). The definition of precision involves the count of "true positives" or "relevant" or "useful" recommendations. Defining what these terms mean is possible in more than one (mathematically correct) ways, depending on the use case.

The reason for this is the items that are recommended but do not appear in the test data. There is not enough information about whether these items are relevant or not, because there can be several reasons why they do not appear in the data:
- the user did not select them,
- they were not sampled when generating the test data set,
- they were not selected by the user because they were not shown to him/her etc. 
In an online setting this information is known, but for offline evaluation it is not possible to make a generic assumption as to which of the above applies to each user-item.

So, it is up to the practitioners who have more information about the use case to make appropriate assumptions about this, as they see fit. For example, the data scientist could assume that 
1. all recommended items that do not appear in the test set are not relevant
2. all recommended items that do not appear in the test set are relevant
3. some are relevant, some are not.
In case 1, you obtain the definition of precision that we and Spark MLLib use. In case 2, you obtain the definition that you propose. Case 3 requires that the data scientist assumes a relevance Boolean value for every item and every user.

It is clear that all assumptions can be problematic and there may be a large degree of uncertainty. This is inevitable because of the limitations of what information is available in the offline setting.

The scope of this PR is just case 1 and Miguel verified that we get the expected behavior. If you want you could add case 2 in a new issue, as a new method or with a keyword argument in the existing method (@miguelgfierro @yueguoguo which is more pythonic?) You could even implement case 3 but that would require passing the entire user-item relevance matrix as an input argument.

miguelgfierro: Really great explanation, all this can be related to the concept of data is missing not at random, that we discussed before. In the paper https://dl.acm.org/citation.cfm?id=1835895 the authors define a metric that can cope with this problem

miguelgfierro: Copying here my comments from teams:
I think this discussion is great, thanks Mirco for taking the time to review this. In regards with the pandas/numpy, something that we discussed before was to have the low level computation in numpy, even though the high level interface uses pandas. If the internal code is faster with numpy, why not using that.
Regarding the reference scale on precision, I think it is very interesting to show more clarity about this. There is a fundamental difference of traditional machine learning systems and a recommendation systems, which is the uncertainty.
Let's take precision at 10 for Alice. Alice has only ranked the 7 Harry Potter movies with 5 stars, so the maximum precision at 10 would be 0.7, right? Now, does that mean that this value is a wrong measurement? what the literature says is no. First, because we are fixing the k, and second, because we don't know the ground truth of all the other items. Following the example, if the reco systems recommends the 3 books of Lord of the Rings to Alice, she might like them, right?

anargyri: Yes, we could implement the measures defined in the paper too.

gramhagen:
related to this, was my comment about recall@k, there are similar edge cases here to consider, if k < # relevant documents, max(recall@k) per user would be k/(# relevant items), which makes the metric challenging particularly when averaging over users (with different relevant item counts) as well as interpreting 'changes' in performance if k is altered.

tests/unit/test_python_evaluation.py
yueguoguo: In terms of implementation, it would be great to have them coded by following the principles we have detailed in the [contributing guidelines](https://github.com/Microsoft/Recommenders/blob/master/CONTRIBUTING.md). For instance, if it is a new metric, and it is something additional to the existing ones, it should be named differently from the existing ones. In this case, it can be called `normalized_precision_at_k` or things similar.

WessZumino: @anargyri thanks for the detailed explanation. Here are few comments: 

> The scope of this PR is just case 1 and Miguel verified that we get the expected behavior. 

I thought the scope of this PR was to show that @k measures are normalised even when k > the test set. At least this is what I have discussed on Teams. That no problem exists when k < = test set has never been under discussion and it is what Miguel verified here or it could be seen directly by looking at the equations. 

I agree about everything else you said; as for the lack of information, we can only discuss what we measure. If our information is incomplete, then we can only asses the validity of the model on this partial information and update our assessment when new information is available. 

However, the scenario of the at k metrics looks conceptually different to me: 

We propose 10 movies but only 6 have been seen/rated by the user. If my model predicts all 6 movies in the test correctly, why is it penalised with a score of 6/10 instead of 6/6? The best I can do is to asses my model on the data I have, not those I could have. In this case, my model could be very good, but I am penalising it because I do not have enough test data to verify its predictions. 

Given the three definitions you proposed, I would say that I am proposing to adopt point 1  

> all recommended items that do not appear in the test set are not relevant

Being not relevant they should not penalise the model. Whereas you are proposing point 2

> all recommended items that do not appear in the test set are relevant

Nevertheless, I agree that this is a matter of definitions. This is why I have only proposed to report the reference scale together with the provided number. As I mentioned before, a ranking number without a ranking scale does not make much sense by definition.

anargyri: In fact, the case k > test set has already been tested by Miguel in lines 212, 193 in the two files.

There has been discussion in the literature about the topic of evaluating precision and recall; a brief summary can be found in [_Evaluating Collaborative Filtering Recommender Systems, HERLOCKER, KONSTAN, TERVEEN, and RIEDL, 2004_], pages 23-24. As they say, precision for recommender systems is subjective to a large extent and hence has been defined in multiple ways.

I think what you propose (that we report a maximum possible for the current definition we have) is a good idea. Feel free to create a new issue or PR.

I also think it will be good to implement your preferred definition for precision as a separate function. In this way, we will have two measures of precision
- one "pessimistic" (the lower bound), i.e. the one we currently have,  
- the other "optimistic" (the upper bound). 

The true precision (which estimates the probability that the user will click on the recommendations produced by the algorithm) lies somewhere in between the two measures. But we don't know where exactly, unless we are doing online evaluation. So, if the practitioner has access to both these measures, then they may be able to better compare the results from offline evaluation to those from A/B testing.
@yueguoguo @miguelgfierro any thoughts?

anargyri:
> ok guys, to try to settle this. My proposal is the following to create a notebook, maybe more general to what Mirco has already created, and focussed on the issue with the missing data.
> We can directly call it: `Data missing not at random`. In a first section, we can talk about the observations noted by Mirco, and then that will give us a lead to talk about MNAR, we could explain the concept and add the toy example in (Steck 2010). Then maybe we could give some pointers to how this is solved, i.e all rank functions of Steck, Inverse Propensity Scoring as defined in (Schnabel et al, 2016) and others, or give some pointers to LightFM (even we can get Maciej collaboration for this in the future).
> 
> Please, either like this or propose a different solution

See also my other suggestion somewhere in the comments above. The two proposals are parallel, so we could implement both.

tests/unit/test_python_evaluation.py
miguelgfierro: @anargyri is `the other "optimistic" (the upper bound).` something seen in literature or in libraries

yueguoguo: Sorry guys late to the party... @miguelgfierro and @anargyri Section 8.4 "Evaluation of ranked retrieval results", page 161 of https://nlp.stanford.edu/IR-book/pdf/08eval.pdf, the author suggests R-precision as an alternative. This might be the easiest approach - one can fix or threshold the number of relevant documents in the ground truth. Actually, we have this implemented in the evaluator as `min_rating` filter.

anargyri: I think this type of measure is a bit different, because it depends on k in a weaker way. It makes sense for information retrieval, but I am not sure how well it applies to recommenders. What they do, if I understand it, is that if a user has seen 1000 items and our top-10 recommendations contain 3 of them then R-precision equals 3/1000 and also R-recall equals 3/1000. It doesn't matter whether these 3 items are the top 3 ranked ones among the 1000, or the bottom 3 ones; the R-precision is the same. 

If you restrict the set to the top 10 among the 1000 then it becomes what we have.

Also note that if we report R-precision, then precision and recall will always be the same.

About implementation, we may need to change the notebooks in order to produce all the test data, not just the top k.

anargyri: See also the discussion in the reference I gave above (_Herlocker et al._). They talk about R-recall and R-precision and cite _Sarwar et al._ but they say 
"While this metric can be useful, it has underlying biases that researchers must be aware of. In particular, the value of this metric depends heavily on the percentage of relevant items that each user has rated. If a user has rated only a small percentage of relevant items, a recommender with high “true recall” may yield a low value for measured recall, since the recommender may have recommended unrated relevant items. Accordingly, this metric should only be used in a comparative fashion on the same dataset; it should not be interpreted as an absolute measure".

anargyri: > @anargyri is `the other "optimistic" (the upper bound).` something seen in literature or in libraries

I haven't seen this in the literature; some people in blogs argue for it (I don't remember the links) but I guess that doesn't say much.  

Probably both the pessimistic and optimistic measures are pretty bad, which partly accounts for the discrepancies between offline and online results seen in practice. Accurate measurement is a complex topic and there are many alternative ways to approach this. E.g. you could assume that the precision per user does not change above the test set size; or you could argue that the test set size is correlated with the probability of the user clicking on the item; and so on.

yueguoguo: > See also the discussion in the reference I gave above (Herlocker et al.). They talk about R-recall and R-precision and cite Sarwar et al. but they say
"While this metric can be useful, it has underlying biases that researchers must be aware of. In particular, the value of this metric depends heavily on the percentage of relevant items that each user has rated. If a user has rated only a small percentage of relevant items, a recommender with high “true recall” may yield a low value for measured recall, since the recommender may have recommended unrated relevant items. Accordingly, this metric should only be used in a comparative fashion on the same dataset; it should not be interpreted as an absolute measure".

Good points. This is what I suggested earlier - we may want to provide alternatives to the existing metrics, but also give explanations and guidance about when to use what, what the limitations are, etc.

anargyri: Sounds good. an additional thing we could check is how high the variance of these metrics is i.e. if you resample the test data what range of values you get. We could also compute bootstrapped metrics in order to get better estimates.

yueguoguo: Good idea. Looks like most of the stuff mentioned here are covered in @WessZumino new notebook?

gramhagen:
Sounds like a good proposal to me, providing information about MNAR would be great and demonstrating some of the biases in the models, as well as deficiencies in the metrics would be very helpful.

tests/unit/test_python_evaluation.py
miguelgfierro: ok, closing this PR and moving info to this issue https://github.com/Microsoft/Recommenders/issues/459

Issue #399, opened 2019-01-04
Unit tests for NCF implementation


yueguoguo:
### *What* is affected by this bug?
Unit tests for NCF core algorithm implementation are missing.

yueguoguo:
Fixed in #417

PR #417, opened 2019-01-13
Unit tests for NCF implementation algorithm  test

AaronHeee:
### Description
Add unit tests for NCF model

### Releated Issues
#399 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/notebooks/02_model/ncf_deep_dive.ipynb
+++ b/notebooks/02_model/ncf_deep_dive.ipynb
@@ -18,7 +18,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 10,
    "metadata": {},
    "outputs": [
     {
@@ -37,6 +37,7 @@
     "sys.path.append(\"../../\")\n",
     "import time\n",
     "import os\n",
+    "import shutil\n",
     "import pandas as pd\n",
     "import numpy as np\n",
     "import tensorflow as tf\n",
@@ -54,7 +55,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 11,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -172,7 +173,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 12,
    "metadata": {},
    "outputs": [
     {
@@ -251,7 +252,7 @@
        "4     166     346     1.0  886397596"
       ]
      },
-     "execution_count": 4,
+     "execution_count": 12,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -267,7 +268,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 13,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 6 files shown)


tests/unit/test_ncf_singlenode.py
```diff
@@ -0,0 +1,163 @@
...
+from reco_utils.common.constants import (
+    DEFAULT_USER_COL,
+    DEFAULT_ITEM_COL,
+    DEFAULT_RATING_COL,
+    DEFAULT_TIMESTAMP_COL,
+)
+
+from tests.ncf_common import python_dataset_ncf, test_specs_ncf
+
+@pytest.mark.parametrize(
```
miguelgfierro: I think here we are missing the gpu test mark: `@pytest.mark.gpu` see this example: https://github.com/Microsoft/Recommenders/blob/staging/tests/unit/test_template_gpu.py. It will be the same in every test that has NCF

notebooks/02_model/ncf_deep_dive.ipynb
```diff
@@ -18,7 +18,7 @@
   },
```
miguelgfierro: Apart from testing the python files, we are also testing the notebooks, here you have an example: https://github.com/Microsoft/Recommenders/blob/staging/tests/unit/test_notebooks_python.py. It uses papermill to run the notebook. You could add the test in the file `https://github.com/Microsoft/Recommenders/blob/staging/tests/unit/test_notebooks_gpu.py`.

Apart from the unit tests for notebooks, we have smoke and integration tests that run every night. Here you have an example of smoke test in surprise https://github.com/Microsoft/Recommenders/blob/staging/tests/smoke/test_surprise_svd.py and integration tests: https://github.com/Microsoft/Recommenders/blob/staging/tests/integration/test_surprise_svd.py. As you can see, this tests run the notebook and make sure that the code retrieves reasonable metrics, the difference between smoke and integration is just the size of the database we use. Here you can find more details about the difference between tests: https://miguelgfierro.com/blog/2018/a-beginners-guide-to-python-testing/

miguelgfierro:
hey @AaronHeee, apart from the comments, I realized that when you are committing, github is not recognizing your user. So basically, your code contributions are not assigned to you. I think it is important that your hard work is reflected on the repo, so any external people that get the repo, can attribute NCF to you. 
The email that you are using for committing is aaronhe@Aaron-He.local, could you try to add it to your configuration? Alternatively, if you type in your command line `git config user.email aaronheee@users.noreply.github.com`, github automatically detects your user and assign it to your commits. 
Let me know if you need help

AaronHeee:
@miguelgfierro Thanks for your comment and help! I have corrected my username for git now. I'm wondering if I should push the branch again to make this work assigned to me correctly? Or are there any other solutions? Thanks.

yueguoguo:
@AaronHeee not necessary. The concern was that the commits should be from you with the proper profile and name 😄

miguelgfierro:
@AaronHeee not sure if the old commit that you have done have the correct name, do you mind to try to do any change in the code to check that now the assignment is correct

tests/unit/test_ncf_singlenode.py
AaronHeee: Thanks, fix it!

notebooks/02_model/ncf_deep_dive.ipynb
yueguoguo: @miguelgfierro talked with @AaronHeee that we will develop a quick-start notebook and then add smoke/integration tests - similar to what we did with als, surprise-svd, etc. My suggestion is this can go to another PR. What do you think?

Issue #412, opened 2019-01-10
Add more parameter values in hyperparameter searching


yueguoguo:
Hyperparameter tuning notebook can add more parameters to fairly compare the mentioned approaches.

PR #428, opened 2019-01-19
Spark model hypertune

yueguoguo:
### Description
Update with customized Spark evaluators. 

### Motivation and Context
* To show how one can customize Spark constructs for tuning parameters with the native Spark native validation modules.
* Re-run the comparison with more trials.
* Revise some text.

### Releated Issues
#412 

### Checklist:

- [ ] I have added tests.


notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -28,7 +28,7 @@
   },
```
miguelgfierro: We added in the rest of the notebooks an MS header on the fisrt cell
```
<i>Copyright (c) Microsoft Corporation. All rights reserved.</i>
<i>Licensed under the MIT License.</i>
```
Please note that you have to add `<i></i>` in both lines, if not, it is not rendered correctly on Github.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -76,7 +77,7 @@
...
   {
    "cell_type": "code",
-   "execution_count": 60,
+   "execution_count": 37,
    "metadata": {},
    "outputs": [
     {
```
miguelgfierro: About this:
```
%env PYSPARK_PYTHON=/anaconda/envs/recommender/bin/python
%env PYSPARK_DRIVER_PYTHON=/anaconda/envs/recommender/bin/python
```
In the other spark notebooks we are not adding this since it is explained in the setup. In case you still think it is important to have it, maybe we should add a comment saying that you have to add the executor of your environment.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -301,53 +309,113 @@
...
-   "execution_count": 69,
+   "execution_count": 46,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
       "Run 0:\n",
-      "\tValidation Metric: 0.9515946345930347\n",
```
miguelgfierro: Here what is exactly the metric, is it RMSE?

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -750,15 +1252,15 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 79,
+   "execution_count": 66,
    "metadata": {},
    "outputs": [],
    "source": [
     "import hyperopt.pyll.stochastic\n",
```
miguelgfierro: `import hyperopt.pyll.stochastic` should go in the top cell

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -650,7 +1152,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 78,
+   "execution_count": 65,
    "metadata": {},
    "outputs": [],
    "source": [
```
miguelgfierro: The first time I read the notebook, the title `5 Hyperparameter tuning with `hyperopt` sampling methods` was weird to me, since the previous one was `4 Hyperparameter tuning with hyperopt`. Then I came back to 4 and realized that there we were using TPE. 
Maybe we could change 4 to be something like `4 Hyperparameter tuning with hyperopt TPE methods`?. Another option could be to have a big section of hyperopt, and then 2 subsections with the two methods.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -895,7 +1434,7 @@
   },
```
miguelgfierro: Would it be possible to compare as well the model with the custom p@k loss function?

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -966,42 +1505,42 @@
...
+       "0     spark  0.916510  0.728810            0.350507   0.327230   41.124152\n",
+       "0  hyperopt  0.931035  0.745581            0.338503   0.305381  283.482648\n",
+       "0    sample  0.971059  0.767857            0.265452   0.244376  580.297325"
       ]
      },
-     "execution_count": 86,
+     "execution_count": 75,
      "metadata": {},
      "output_type": "execute_result"
     }
```
miguelgfierro: So looking at the results, I get the explanation of the times. However, in terms of performance, basic grid search is better than TPE. I would love to understand this better.

In the paper, TPE outperforms manual and random hyperparameters, and even gaussian processes, however, in the notebook's results there is no improvement. 

We are saying: `This may be owing to the limited number of iterations used in the TPE searching algorithm. ` Can we test this?

Are there any examples with hyperopt where TPE outperforms manual grid search?

Adding @anargyri @eisber @gramhagen @loomlike for discussion.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -574,12 +1076,12 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 76,
+   "execution_count": 63,
    "metadata": {},
```
miguelgfierro: In the rest of the notebooks we are placing all imports in the top cell
```
from pyspark.ml.evaluation import Evaluator
from pyspark.mllib.evaluation import RegressionMetrics, RankingMetrics
```

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -531,39 +1033,39 @@
...
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 75,
+   "execution_count": 62,
    "metadata": {},
    "outputs": [
     {
      "data": {
```
miguelgfierro: About this:
`For example, hyperopt avails Tree of Parzen Estimators (TPE) method for searching optimal parameters, which is more efficient than random search or grid search (details can be found here)`.
Maybe we should add a short explanation of what TPE does, or what are the benefits claimed?

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
yueguoguo: Removed. Good catch

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
yueguoguo: This was also what I have discussed with Andreas. I am doing the testing on Databricks where the size of datasets, # iterations, etc. are variables to test scalability and other things of our interest. Would love to have your thoughts on this.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
yueguoguo: I did not do that because p@k takes much longer time than rating metrics like rmse. Again, I am thinking to leave scalable discussions onto Azure Databricks notebooks, instead of putting everything into the current one. It is just for illustration purpose. What do you think?

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
yueguoguo: Good idea. Added.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
miguelgfierro: yeah, it's curious

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
miguelgfierro: yeah, makes sense

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
yueguoguo: Agree. I can put those studies into the repo. Maybe after our setup of an Azure Databricks DevOps pipeline?

anargyri: The faster times for Spark may be due to the parallelism. It would be interesting to try the same tuning experiment with hyperopt on mongodb. 

However, even sequential hyperopt has an advantage because it is doing "intelligent" search of the space and not exhaustive grid search. So, there is a trade off between smart search and parallelism. For a small number of hyperparameters (2 in this notebook), parallelism will prevail. As the number of hyperparameters increases, I would expect hyperopt to catch up and at some point be faster than Spark grid search.

One way to show this would be to tune the `alpha` and `nonnegative` hyperparameters of ALS, so that you search a 4d space of hyperparameters. We would probably see better times for hyperopt relative to Spark.

miguelgfierro: what puzzles me is that even though

> it is doing "intelligent" search of the space and not exhaustive grid search.

The metrics are not better in the "intelligent search"

anargyri: They should not be better if exhaustive search was done completely. It is better in cases when it is not possible to search the hyperparameter space in an exhaustive way.

miguelgfierro: in this case we are not doing a complete search

anargyri: Actually it is not clear if ALS finds a good set of hyperparameters, because we don't see the precision in the output. 
Also note that we optimize precision but we report regression metrics. @yueguoguo shouldn't we report precision too?

yueguoguo: It is hard if not impossible to find a "complete search" without a prior knowledge of the parameter range. Regardless random search or grid search, the parameter space is predefined by human and the two approaches just find the optimal ones in different ways - the random search can benefit from some "intelligent" method (e.g.., Bayesian method, early-stop policy, etc.) to reduce the # of iterations, while the second does nothing "intelligent" but to finish the whole iterations and pick the optimal from all.

I think I need to rephrase a little bit in the notebook that this is NOT for comparison. Instead, it is just for providing users with information about how different approaches work and how to choose by a given criterion. A scalable version of the notebook (on Databricks) makes more sense for comparison. What do you think?

yueguoguo: > Also note that we optimize precision but we report regression metrics. @yueguoguo shouldn't we report precision too?

Precision is meant to demonstrate that we can customize Spark constructs to allow any metrics of interest. It is not for comparison. I should have made it more clear in the notebook text.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -406,13 +471,429 @@
...
+    "    .addGrid(alstopk.rank, [10, 20]) \\\n",
+    "    .addGrid(alstopk.regParam, [0.1, 0.01]) \\\n",
+    "    .build()\n",
+    "\n",
+    "tvs = TrainValidationSplit(\n",
+    "    estimator=alstopk,\n",
+    "    estimatorParamMaps=paramGrid,\n",
+    "    # A regression evaluation method is used. \n",
+    "    evaluator=PrecisionAtKEvaluator(),\n",
+    "    # 80% of the data will be used for training, 20% for validation.\n",
```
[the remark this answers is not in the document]
yueguoguo: Good catch!

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
anargyri: I think it becomes a bit confusing because you call the dataframe "comparison" and many people will focus on the table and view it as some type of comparison. 
IMO I would like to avoid giving the impression that if you tune by optimizing precision you also optimize RMSE, MAE etc. as a result. We know that this is not correct, but some people may not be aware.

yueguoguo: Good point. I will rephrase this part to make it sound non-opinionated.

Issue #373, opened 2018-12-17
Add XDeepFM


yueguoguo:
Add xDeepFM implementation / notebooks.

yueguoguo:
Related to #339

nikhilrj:
Closed with #453

Issue #218, opened 2018-11-12
Add FM with Tensorflow


miguelgfierro:
Add FM from deeprec project

miguelgfierro:
@Leavingseason do you have a vanilla FM in deeprec?

miguelgfierro:
closing because of old issue, already added a FM: https://github.com/microsoft/recommenders/pull/909

PR #430, opened 2019-01-21
Deeprec/first

Leavingseason:
### Description

Add a deeprec module, including the  xDeepFM and DKN model.

### Releated Issues
https://github.com/Microsoft/Recommenders/issues/218
https://github.com/Microsoft/Recommenders/issues/373
https://github.com/Microsoft/Recommenders/issues/188

### Checklist:

- [x] I have added tests.
- [ ] I have updated the documentation accordingly. TO DO


tests/unit/test_deeprec_model.py
```diff
@@ -0,0 +1,40 @@
...
+import pytest
+import os
+from reco_utils.recommender.deeprec.deeprec_utils import *
+from reco_utils.recommender.deeprec.models.xDeepFM import *
+from reco_utils.recommender.deeprec.models.dkn import *
+from reco_utils.recommender.deeprec.IO.iterator import *
+from reco_utils.recommender.deeprec.IO.dkn_iterator import *
+
+@pytest.mark.gpu
+@pytest.mark.deeprec
```
miguelgfierro: This is not needed: `@pytest.mark.deeprec`. The CICD system will take directly everything that has the mark gpu and execute it

tests/unit/test_deeprec_utils.py
```diff
@@ -0,0 +1,58 @@
...
+@pytest.fixture
+def resource_path():
+    return os.path.dirname(os.path.realpath(__file__))
+
+
+@pytest.mark.parametrize("must_exist_attributes", [
+    "FEATURE_COUNT", "data_format", "dim"
+])
+@pytest.mark.gpu
+@pytest.mark.deeprec
```
miguelgfierro: same comment as before, `@pytest.mark.deeprec` is not needed

examples/try_SAR.py
```diff
@@ -0,0 +1,104 @@
+
```
miguelgfierro: There is already a quick example of SAR as a notebook, https://github.com/Microsoft/Recommenders/blob/staging/notebooks/00_quick_start/sar_single_node_movielens.ipynb, which is super similar to this file

examples/try_deeprec.py
```diff
@@ -0,0 +1,28 @@
+
```
miguelgfierro: Could you please transform this python file in a notebook as a quickstart, to follow the structure that we already have. Similar to this: https://github.com/Microsoft/Recommenders/blob/staging/notebooks/00_quick_start/sar_single_node_movielens.ipynb

examples/generate_dataset.py
```diff
@@ -0,0 +1,28 @@
+
```
miguelgfierro: Following the comment below, to follow the structure that we already have, could you please add a quickstart notebook? similar to https://github.com/Microsoft/Recommenders/blob/staging/notebooks/00_quick_start/sar_single_node_movielens.ipynb

tests/resources/deeprec/output.txt
```diff
@@ -0,0 +1,7 @@
+2.316978e-06
```
miguelgfierro: All the resource files (npy, txt and other data), could you please upload it a blob? We don't want the repo to be super heavy. The blob we use is called recodatasets, I just created a folder called deeprec. I'll send you the keys privately

examples/try_SAR.py
anargyri: Yes, it  seems this file can be deleted.

examples/generate_dataset.py
anargyri: Maybe you don't need this file. In the notebooks, we don't use synthetic data, but public datasets such as Movielens. In this case we could make a notebook with Movielens or Criteo.

examples/try_deeprec.py
anargyri: Also could you use a public data set such as Movielens or Criteo instead of a synthetic one? 
You could choose to do a quick start notebook or a deep dive one (see Recommenders/notebooks/02_model/). In both cases the data size should not be very large, so that the notebook runs in a few minutes (for Criteo you could subsample the data). The difference is that in the deep dive notebooks you explain in more detail what happens at each step and how the algorithm works under the hood.

tests/unit/test_deeprec_model.py
Leavingseason: I prefer to add this tag, because for my offline running, I only need to run some specific part of test cases, and offline run the entire test cases are too time consuming

examples/try_SAR.py
Leavingseason: Sure I can remove it. 
I just wonder why you prefer a notebook rather than a direct end2end example.

tests/resources/deeprec/output.txt
Leavingseason: Can you provide some examples for blob usage? I am not very clear what is it

tests/unit/test_deeprec_utils.py
Leavingseason: I prefer to add this tag, because for my offline running, I only need to run some specific part of test cases, and offline run the entire test cases are too time consuming

reco_utils/recommender/deeprec/IO/dkn_iterator.py
```diff
@@ -0,0 +1,163 @@
...
+                    candidate_news_index_batch = []
+                    candidate_news_val_batch = []
+                    click_news_index_batch = []
+                    click_news_val_batch = []
+                    candidate_news_entity_index_batch = []
+                    click_news_entity_index_batch = []
+                    label_list = []
+                    impression_id_list = []
+                    cnt = 0
+            # if cnt > 0:
```
yueguoguo: Remove commented codes.

tests/unit/test_deeprec_utils.py
```diff
@@ -0,0 +1,58 @@
...
+def test_DKN_iterator(resource_path):
+    data_file = os.path.join(resource_path, r'../resources/deeprec/dkn/final_test_with_entity.txt')
+    yaml_file = os.path.join(resource_path, r'../resources/deeprec/dkn/dkn.yaml')
+    hparams = prepare_hparams(yaml_file, wordEmb_file='', entityEmb_file='')
+    iterator = DKNTextIterator(hparams, tf.Graph())
+    assert iterator is not None
+    for res in iterator.load_data_from_file(data_file):
+        assert isinstance(res, dict)
+
+# if __name__ == '__main__':
```
yueguoguo: Remove commented codes if not needed.

tests/resources/deeprec/output.txt
miguelgfierro: A good example is on this notebook: https://github.com/Microsoft/Recommenders/blob/staging/notebooks/02_model/ncf_deep_dive.ipynb

The diagram with the architecture of NCF is actually hosted here: https://recodatasets.blob.core.windows.net/images/NCF.svg. 

Another example are the movielens datasets, we load them from an external storage, you can see the code here: https://github.com/Microsoft/Recommenders/blob/staging/reco_utils/dataset/movielens.py#L386

tests/unit/test_deeprec_model.py
miguelgfierro: ok sounds good

examples/try_SAR.py
miguelgfierro: It is just the way the repo has been originally designed. We found that there is a high interest from our users to see notebooks instead of pure python files. One reason might be that in a notebook, apart from code you can add formatted comments and images.

tests/resources/deeprec/output.txt
Leavingseason: Got it, thanks!

examples/try_deeprec.py
anargyri: @Leavingseason For this notebook, do you think you can use Movielensm or do you prefer Criteo instead? If the latter, one of the team should write a dataset loader for Criteo first.

Leavingseason: @anargyri deeprec uses a different type of dataloaders, I put them under the IO folder. I can try using a small sample of Criteo dataset.

examples/generate_dataset.py
Leavingseason: I will remove this script to avoid too much workload on notebook...

tests/resources/deeprec/output.txt
Leavingseason: I separate data files to blob storage,  you can see the commit here: https://github.com/Microsoft/Recommenders/pull/430/commits/4eb799c52114985ace545f3849ea396477f8db9e

notebooks/02_model/ncf_deep_dive.ipynb
```diff
@@ -26,21 +26,8 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": 1,
+   "cell_type": "markdown",
```
anargyri: This notebook must have been changed by mistake.

notebooks/02_model/baseline_deep_dive.ipynb
```diff
@@ -10,8 +10,10 @@
    ]
   },
   {
-   "cell_type": "markdown",
+   "cell_type": "code",
```
anargyri: This notebook too was changed by mistake.

examples/try_deeprec.py
miguelgfierro: Criteo would be great, not sure if @yexing99 has worked with it. @wutaomsft FYI

anargyri: The approach we follow in the repo is to avoid external dataset loaders and instead use data set loaders common to all algorithms, as in https://github.com/Microsoft/Recommenders/blob/master/reco_utils/dataset/movielens.py.
These return a pandas dataframe with the specified columns. This dataframe can then be manipulated with the external libraries and algorithms.

So, I suggest a similar approach here, in which you take the pandas dataframe that we provide and you apply the data transformations from your code to it. It may look like a detour, but this enables everyone to easily apply DeepRec on any new dataset they have: they just need to write a new loader under reco_utils/dataset/ and they don't need to modify your code. Moreover, this loader will be reusable by the other algorithms we have in the repo.

Leavingseason: I don't think so. DeepRec contains different types of data, e.g. FFM format for xdeepfm,  DKN format which requires aligned news words and news entities. It's hard to use dataframe to store them. Besides the data format, in feature-based algos the feature files are usually large (in production, may be dozens of GB). We can never try to load the entire feature file into memory via dataframe .

reco_utils/recommender/deeprec/IO/dkn_iterator.py
```diff
@@ -0,0 +1,163 @@
...
+
+            else:
+                raise ValueError("data format is wrong")
+
+        return label, candidate_news_index, candidate_news_val, click_news_index, click_news_val,\
+            candidate_news_entity_index, click_news_entity_index, impression_id
+
+
+    def load_data_from_file(self, infile):
+        candidate_news_index_batch = []
```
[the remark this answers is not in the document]
Leavingseason: Sure. I will add docstrings later.

examples/try_deeprec.py
anargyri: I see. Is FFM format the same as in the FFM paper and in https://github.com/guestwalk/libffm ?
Would it make sense to have a loader for FFM, one for DKN etc.? I am thinking that having the data in sparse FFM format may be useful in its own right, e.g. the user can do some data preprocessing, visualization, input the data to an FFM etc.

Leavingseason: Yes, the format is almost the same as in your link. In fact we are indeed using an iterator for FFM, and an iterator for DKN.  We hope that similar feature-aware models can share the same type of loader (iterator), but not a unified loader which can take all types of feature-aware data (for example, not good for one loader which can take both FFM and DKN format)

anargyri:
Hi Jianxun,
It would be good to have these files merged into the repo even only with synthetic data, as they currently are. The only thing that is missing from this PR is a notebook that showcases the algorithm (just xDeepFM for now). This would be something like the notebooks we have e.g. https://github.com/Microsoft/Recommenders/blob/master/notebooks/02_model/surprise_svd_deep_dive.ipynb in which you would explain briefly what the algorithm does in a way similar to the paper and cite the paper.

miguelgfierro:
hey @anargyri Jianxun is working on this PR https://github.com/Microsoft/Recommenders/pull/453. The current one was closed to not have all the history of files

Issue #434, opened 2019-01-22
[BUG] check smoke and integration tests for SAR and fastai


miguelgfierro:
### *What* is affected by this bug?
SAR and fastai nightly builds are failing https://msdata.visualstudio.com/AlgorithmsAndDataScience/_build/results?buildId=2200124

### In *which* platform does it happen?
* *Azure Data Science Virtual Machine.*

### Expected behavior (i.e. solution)
*For example:*
* The tests  should pass successfully.

PR #435, opened 2019-01-22
add additional packages required by fastai

danielsc:
### Description
Fastai requires dataclasses to be pip installed if built on python < 3.7

### Motivation and Context
fixes fastai part of https://github.com/Microsoft/Recommenders/issues/434

### Checklist:

- [x] I have added tests.
- I ran nightly tests on the change: https://msdata.visualstudio.com/AlgorithmsAndDataScience/_build/results?buildId=2201874

### DIFF
--- a/scripts/generate_conda_file.sh
+++ b/scripts/generate_conda_file.sh
@@ -108,7 +108,8 @@ ${pyspark}- pyarrow>=0.8.0
 ${gpu}  - numba>=0.38.1 
   - gitpython>=2.1.8
   - pydocumentdb>=2.3.3
-  - azureml-core>=0.1.74
+${gpu}  - nvidia-ml-py3>=7.352.0
+  - dataclasses>=0.6
 EOM
 
 echo "Conda file generated: " $CONDA_FILE
--- a/scripts/generate_conda_file.sh
+++ b/scripts/generate_conda_file.sh
@@ -108,7 +108,7 @@ ${pyspark}- pyarrow>=0.8.0
 ${gpu}  - numba>=0.38.1 
   - gitpython>=2.1.8
   - pydocumentdb>=2.3.3
-${gpu}  - nvidia-ml-py3>=7.352.0
+  - nvidia-ml-py3>=7.352.0
   - dataclasses>=0.6
 EOM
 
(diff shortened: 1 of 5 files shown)


scripts/generate_conda_file.sh
```diff
@@ -108,7 +108,8 @@ ${pyspark}- pyarrow>=0.8.0
 ${gpu}  - numba>=0.38.1 
   - gitpython>=2.1.8
   - pydocumentdb>=2.3.3
-  - azureml-core>=0.1.74
+${gpu}  - nvidia-ml-py3>=7.352.0
+  - dataclasses>=0.6
```
miguelgfierro: a question, are you considering that our user will use fastai in a non-gpu environment? As of now, all TF code is considered to go on a gpu env

danielsc: Fast ai should run without a GPU, but it will be slower (the same should be true for TF). I wonder if we want to narrow down the audience for these algorithms to just users that have a GPU. Often times, I would like to try out a github repo on my Mac, just to see how/if it works. What are your thoughts?

danielsc: on a related note: do we have a way to run tests on GPU hardware in automation?

miguelgfierro: for question 1, if I'm doing deep learning I always use a gpu machine, I don't know if I'm the typical user though.
for q2, we have a test pipeline on gpu (in fact our test machine is a gpu machine)

danielsc: ok -- I will put fastai behind the ${gpu} tag then

danielsc: looking at the files, we are installing the non-gpu version of tensorflow if the user does not select gpu

tests/smoke/test_template_gpu.py
```diff
@@ -3,10 +3,37 @@
...
 
 
+TOL = 0.05
+
 @pytest.mark.smoke
 @pytest.mark.gpu
 def test_gpu_vm():
     assert get_number_gpus() >= 1
 
+
```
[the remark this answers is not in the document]
danielsc: moved things into separate file and also added a unit test that runs the notebook.

Issue #441, opened 2019-01-23
Bug on hypertune_spark_deep_dive notebook dataload


loomlike:
Movielens dataloader uses default column names in the repo. To use custom names, should pass either header or schema.

### *What* is affected by this bug?
* *Run of a notebook.*

### In *which* platform does it happen?
* Spark

### *How* do we replicate the issue?
* Run the notebook

### Expected behavior (i.e. solution)
Easily fix by changing the data loading line as:
`load_spark_df(spark, size='100k', header=(COL_USER, COL_ITEM, COL_RATING))`

loomlike:
Also need to check if we have a test function for this notebook or we intentionally not added the test for this? @yueguoguo @miguelgfierro

yueguoguo:
Fixed in #442

PR #444, opened 2019-01-24
Spark hypertune 

yueguoguo:
### Description
Notebook for Spark hyperparameter tuning.

### Motivation and Context
* Fixed bugs.
* Updated text to make it clear.

### Releated Issues
#441 

### Checklist:

- [ ] I have added tests.


notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -166,7 +166,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 6,
    "metadata": {},
    "outputs": [],
    "source": [
```
anargyri: Better say "runs [hyperparameter tuning"  rather than "is supposed to run hyperparameter tuning".

anargyri: "It is worth mentioning that tuning hyperparameter with `hyperdrive` in Azure Machine Learning services will be a feature in the future version.". This contradicts the previous paragraph, which says how to do tuning with AML.

yueguoguo: 👍

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -984,7 +929,7 @@
   },
```
anargyri: You use loguniform, not lognormal.
The following is not correct:
"In the ALS algorithm, the two hyper parameters, rank and reg, affect model performance in a way that\n",
    "* The higher the rank, the better the model performance but also the higher risk of overfitting.\n",
    "* The reg parameter is subject to rank and it prevents overfitting in certain way. \n",

Model performance (on test or validation data) will improve and then deteriorate, but it could go up and down more times. You don't know what will happen in advance. The reg. parameter prevents overfitting but is not related to the rank hyperparameter.

notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
```diff
@@ -1513,57 +1406,57 @@
...
     "\n",
-    "The three approaches use the same RMSE loss. In this measure, the native Spark construct performs the best. The `hyperopt` based approach performs the second best. This may be owing to the limited number of iterations used in the TPE searching algorithm. Note the differences in the RMSE metrics may also come from the randomness of the intermediate steps in parameter tuning process. In practice, multiple runs are required for generating statistically robust comparison results. We have tried 5 times for running the same comparison codes above. The results aligned well with each other in terms of objective metric values and elapsed time. "
```
anargyri: This statement is debatable. What happens is that all three methods perform the same since they get roughly the same validation RMSE (0.94). But they get different RMSE on the test data set. So, the methods do what they are supposed to do but they probably find a bit different hyperparameter values, which lead to some difference on the test data. Or it could be the randomness as you say. So, I don't think that in this experiment hyperopt is worse than Spark. If you tried it with a different dataset you could get better RMSE on test data with hyperopt. It is a matter of statistical variation.

What you can say for sure is that to get very similar statistical performance, spark takes much less time than hyperopt.

yueguoguo: This is a very good point. Indeed, the overall performance may be from dimensionality of hyperparameters, dataset, randomness of experimentation, etc.

Issue #424, opened 2019-01-16
Add an experimentation example with AML


miguelgfierro:
This article https://azure.microsoft.com/en-gb/blog/experimentation-using-azure-machine-learning/ from @danielsc can be used as a base

loomlike:
I'm working on AML + hyperdrive for wide-deep, excluding auto-ml (don't think we --recommender-- can use automl). Basically I have two notebooks around 75% ready, one for wide-deep and the other one for aml+hyperdrive for wide-deep tuning

miguelgfierro:
closing this since it has been solved on #448 and on #546

Issue #374, opened 2018-12-17
Add Wide&Deep


yueguoguo:
Add Wide & Deep implementation / notebook.

loomlike:
Related to #339

miguelgfierro:
fixed in https://github.com/Microsoft/Recommenders/pull/448

Issue #339, opened 2018-12-07
DNN algo


loomlike:
Add DNN-based algo, e.g.:
* wide deep model
* neural-CF

nikhilrj:
Closing as this is a parent issue - let's track the individual children instead

Issue #233, opened 2018-11-13
Explore capability of utilizing Azure services for a recommendation system


loomlike:
When we want to run reco models on big data, some of Azure services, like Storage or AML will be useful as well as Spark on Databricks.
Maybe good to have a quickstart notebook showing how our recommender repo can be setup with those services and utilized.

PR #448, opened 2019-01-25
1) wide deep model and 2) AzureML-hyperdrive

loomlike:
### Description
This PR covers two notebooks: 1) wide-deep quickstart and 2) aml-hyperdrive, along with the utils for them.

The notebook cells' results are somewhat too verbose (especially wide-deep quickstart's training logs). Maybe we want to cut them out and attach TensorBoard screen-shot instead?

Also, please note, this PR doesn't include unit-tests of those two notebooks, tensorflow-utils and azure-ml training script. Wanted to discuss how to test them. For tf_utils, maybe make a very simple model and test w/ the util functions. For AzureML-related unit-test, will sync up w/ @jreynolds01.

Since the changes in this PR are not small (sorry about that!), would like to ask some of the reviewers to focus on specific codes to make your life easy!:
@eisber Could you review aml training script? (reco_utils/aml/wide_deep_training.py)
@yueguoguo Want to get your feedback especially on aml hyperdrive notebook as you already made a great hyper-parameter tuning notebook! (notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb)
@jreynolds01 You may give me a good feedback for AML service connection part in the aml hyperdrive notebook. I'm considering checking the aml's config.json file first and then using parameters as a fall-back mechanism.
@miguelgfierro, @anargyri  Want you to look at the wide-deep quickstart notebook to check model codes as well as the contents.

I asked those specific files just for your convenience, so please feel free to comments on other codes too.
All the other contributors' feedback is also welcome!

### Motivation and Context
Add deep neural networks recommender examples
Show examples of AzureML for hyper-parameter tuning

### Releated Issues
#374 #339 #424 #297 #233 

### Checklist:

- [ ] I have added tests.


loomlike:
Just found out AML's output widget (which is really cool) is not showing at github. I will attach screen-shot to the notebook later when I address comments.

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
```diff
@@ -0,0 +1,619 @@
...
+    "Licensed under the MIT License.</i>\n",
+    "<br><br>\n",
+    "# Recommender Hyperparameter Tuning w/ AzureML"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this notebook, we show how to hyperparameter tune a recommender model by utilizing **Azure Machine Learning service*** ([AML or AzureML](https://azure.microsoft.com/en-us/services/machine-learning-service/)) in the context of movie recommendation. Note, to use AML, you will need Azure subscription.\n",
```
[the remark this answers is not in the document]
loomlike: @yueguoguo Actually that asterisk is for foot-note reference mark. I will change it to superscript, something like "[<sup>*</sup>](NOTE_LINK)".

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
```diff
@@ -0,0 +1,619 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Found the config file in: C:\\Users\\jumin\\git\\Recommenders\\notebooks\\04_model_select_and_optimize\\aml_config\\config.json\n",
```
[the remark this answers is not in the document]
loomlike: That result string comes from AML function call...and does include non-generic contents. I probably want to clean up this cell w/ `IPython.display.clear_output()`

reco_utils/common/tf_utils.py
```diff
@@ -0,0 +1,189 @@
...
+    tf.estimator.inputs.pandas_input_fn cannot handle array/list column properly.
+    If the df does not include any array/list data column, one can simply use TensorFlow's pandas_input_fn.
+
+    For more information, see (https://www.tensorflow.org/api_docs/python/tf/estimator/inputs/numpy_input_fn)
+
+    Args:
+        df (pd.DataFrame): Data containing features
+        y_col (str): Label column name if df has it.
+        batch_size (int): Batch size for the input function
+        num_epochs (int): Number of number of epochs to iterate over data. If None will run forever.
```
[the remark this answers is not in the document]
loomlike: Good catch!

reco_utils/common/tf_utils.py
```diff
@@ -0,0 +1,189 @@
...
+            )['average_loss']
+
+            eval_log.append("average_loss = {}".format(loss))
+            self._write_simple_value_summary('evaluation/average_loss', loss)
+
+            if self.eval_fn is not None:
+                predictions = list(itertools.islice(
+                    self.model.predict(input_fn=pandas_input_fn(
+                        df=self.eval_df,
+                        batch_size=10000,
```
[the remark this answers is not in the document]
loomlike: Added.

tests/unit/test_aml_wide_deep_training.py
```diff
@@ -0,0 +1 @@
+# TODO except tf.train.NanLossDuringTrainingError:
```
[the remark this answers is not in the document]
loomlike: Oh yeah. Forgot to exclude from commit. Anyway I'm working on that now.

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
```diff
@@ -0,0 +1,619 @@
...
+    "If you are using a [Data Science Virtual Machine (DSVM)](https://docs.microsoft.com/en-us/azure/machine-learning/service/how-to-configure-environment#dsvm) or [Azure Notebook](https://notebooks.azure.com/), `azureml-sdk` is already installed in it so you don't need to install the package.\n",
+    "\n",
+    "To install AML Python SDK*, run\n",
+    "```\n",
+    "pip install --upgrade azureml-sdk[notebooks]\n",
+    "```\n",
+    "\n",
+    "More info about setting up AML environment can be found from [this link](https://docs.microsoft.com/en-us/azure/machine-learning/service/how-to-configure-environment).\n",
+    "\n",
+    "> \\* AML has a Databricks sdk `azureml-sdk[databricks]` but it doesn't support hyperparameter tuning on Databricks for now."
```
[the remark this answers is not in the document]
loomlike: Turns out our conda envs already include azureml-sdk, so I removed this entire prerequisite cell.

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
```diff
@@ -0,0 +1,619 @@
...
+    "  \n",
+    "#### Create a workspace\n",
+    "1. Sign in to the [Azure portal](https://portal.azure.com) by using the credentials for the Azure subscription you use.\n",
+    "2. Select **Create a resource** menu, search for **Machine Learning service workspace** select **Create** button.\n",
+    "3. In the **ML service workspace** pane, configure your workspace with entering the *workspace name* and *resource group* (or **create new** resource group if you don't have one already), and select **Create**. It can take a few moments to create the workspace.\n",
+    "  \n",
+    "<br>\n",
+    "  \n",
+    "#### Make a configuration file\n",
+    "To configure this notebook to communicate with your workspace easily, create a *.\\aml_config\\config.json* file with the following contents:\n",
```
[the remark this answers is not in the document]
loomlike: I'm planning to give another option of using constants like:
```
subscription_id = '<subscription-id>'
resource_group  = '<resource-group>'
workspace_name  = '<workspace-name>'

try:
    ws = Workspace(subscription_id = subscription_id, resource_group = resource_group, workspace_name = workspace_name)
except:
    # try to find config file
```
where we also can parameterize w/ papermill for testing. Will this work?

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
```diff
@@ -0,0 +1,1388 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Tensorflow Version: 1.12.0\n"
+     ]
+    },
+    {
+     "data": {
+      "text/plain": [
```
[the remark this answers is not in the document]
loomlike: It's tricky since if someone edit the contents of this notebook in the future and run the cells, the output values will be changed and not aligned w/ the HW description anymore... any suggestions?

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
```diff
@@ -0,0 +1,1388 @@
...
+     "output_type": "stream",
+     "text": [
+      "Num epochs:  26\n",
+      "Embedding 6040 users to 32-dim vector\n",
+      "Embedding 3706 items to 16-dim vector\n"
+     ]
+    }
+   ],
+   "source": [
+    "\"\"\" Hyper parameters\n",
```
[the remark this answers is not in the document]
loomlike: Good point. will add comments as well as description cell.

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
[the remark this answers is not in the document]
yueguoguo: Good point. Maybe we can add the info at the begging of the notebook about the GPU information.

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
```diff
@@ -0,0 +1,1388 @@
+{
```
miguelgfierro: <p>Maybe it would be worthwhile to add a small paragraph describing what is Wide&amp;Deep at a high level</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

miguelgfierro: <p>In the other notebooks, we have one cell for all the imports and prints of versions</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

miguelgfierro: <p>how many genres are there in total?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

miguelgfierro: <p>Do you think that if we wrap here the one-hot encoding in a function, a user can reuse that function? or is it too specific?</p><p><br></p><p><br></p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

miguelgfierro: <p>the notation `_items` is  weird in a notebook. Adding `_` to a name typically means that this variable is private, but in a notebook there is no such thing.  I would recommend to rename `_items` to either `items` or a different name (without `_`)</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

miguelgfierro: <p>I'm curious, would the splitter work if we don't drop the column `Genres_string`?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

miguelgfierro: <p>mmm interesting, in reviewnb this paragraph is not rendered exactly the same as in the notebook. </p><p>For the list, would it make more sense to add : instead of /? </p><ol><li>Wide: cross...</li><li>Deep: lower...</li></ol>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

miguelgfierro: <p>Can you move these parameters to the top? like in the other notebooks? We will use this to control the parameters in the unit tests. I have to write some documentation on how to add parameters with papermill, so the notebooks can be executed in the tests</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

miguelgfierro: <p>I would suggest to move LOG_STEP and the model directory together with the rest of the parameters</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

miguelgfierro: <p>I'm curious, why MAE and not RMSE? is what they recommend in the paper?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

miguelgfierro: <p>I think we should comment this. We have a loss function MAE and we are optimizing with respect to precision at k. So basically, our model should have a good precision at k. If we are looking to optimize a different metric, such us recall, we can just add here recall and the results will be different. </p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

miguelgfierro: <p>Could we have less verbose logs?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

miguelgfierro: <p>Is there a way to show a screenshot of tensorboard in the notebook?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

miguelgfierro: <p>I think we should explain why we are exporting the model</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

miguelgfierro: <p>As before, this is too verbose</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

miguelgfierro: <p>As before, this is too verbose</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

miguelgfierro: <p>As before, this is too verbose</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
```diff
@@ -0,0 +1,619 @@
+{
```
miguelgfierro: <p>One question here, is this agnostic to the model? Meaning, can I use any model (TF, LightGBM, sklearn) with this notebook?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

miguelgfierro: <p>I would remove the mention to databricks, here we are not using it. Also, it's dangerous to say things like, this is not supported for now, because, in a month this feature can be added and we will forgot that we wrote this line here</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

miguelgfierro: <p>Can you move all the parameters to the same cell? We will tag this cell in papermill. Also, could you add as variables the parameters of the VM? we will tune it with papermill</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

miguelgfierro: <p>Again, I would not mention: "only supported choice". This could change and we will forget about it</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

miguelgfierro: <p>how can this be extended to other models? I wonder if this notebook is more focused on TF models or is generic</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

miguelgfierro: <p>I would add a conclusion here. The objective of using hyperparameter tuning is to have show that the performance metrics are better than a non-optimized selection of parameters. I think we should do a comparison here on the metrics obtained on this notebook with the metrics obtained in the plan quick start notebook of W&amp;D</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/448/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

miguelgfierro:
hey @loomlike the code is really good, congratulations man. I commented only part of the all the files you showed. I will do another pass after you reply to some questions that I have. 

Great work!

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
bethz: <p>As I'm new, a couple of questions:</p><p>Can we refer to it only as AzureML? </p><p>It had been mentioned to me that AML should be avoided as a reference? </p><p>Please inform me so I follow the convention going forward. </p><p>thanks!</p><p><br></p>

miguelgfierro: That makes sense. The standard way we would take this action is to create an issue and then a person assigns himself or herself to achieve the tasks. Beth, would you be willing to take this task?

loomlike: @miguelgfierro Good point. Will add.
@bethz That is interesting... I heard no one says 'AzureML' but all say AML. I referred AzureML along with AML on the notebook only because the package name is `azureml`. But if what you say is true, we should definitely follow that. Can you confirm which is the right way to refer?

loomlike: @miguelgfierro hmm.. somehow ReviewNB inserts an extra line between lines in the comments and doesn't support Markdown. We still can use ReviewNB as a complementary tool though. Super easy to read.

loomlike: Changing

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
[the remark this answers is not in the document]
loomlike: Adding

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
loomlike: Btw, I just kept Genres_string to show how genres(str) converted into int array. No need to keep at all otherwise.

loomlike: Make sense. Also editing the contents into a table per Le's suggestion.

loomlike: Make sense to `NUM_EPOCHS` where we can pass very small number for unit-testing, will do that. But wonder about other hyperparameters. Wouldn't it be better flow to keep them under the modeling section?

loomlike: Moving them to the top where we set parameters

loomlike: No, It is just an example of how to printing out additional metrics via TensorFlow's native functions.

Btw, model's default `average_loss` is Mean Square Error, and thus just `sqrt(average_loss)` is the RMSE--tested w/ our `rmse` function too. That's why I didn't use RMSE and showed MAE instead as an 'additional' metrics example. Nothing deep meaning here haha. Will add more context to the cell.

loomlike: I think this code (along with the previous code setting a metric) gives confusion. This is just showcasing an evaluation hook, which reports the given metric values during the training and records them into TensorBoard.

To make the model optimized into a custom metric other than the default, `loss`, should define own `Estimator`. Would not be too difficult, but rather do that on a deep-dive notebook. 

I think would be better to clean up the codes to be:
- remove the previous MAE part (since we anyway do the separate evaluation at the end)
- change the hook name to `precision_log_hook` so that show its purpose clearly, and
- add explanation what the hook does.

miguelgfierro: > ReviewNB inserts an extra line between lines in the comments
I don't understand this very well. Are you viewing this on github or reviewnb?

bethz: <p><a href="https://github.com/miguelgfierro" target="_blank">@miguelgfierro</a>  <a href="https://github.com/loomlike" target="_blank">@loomlike</a>  Yes, I will confirm which is the right way to refer. I didn't have privs last week but I'll try to create the issue in GitHub now as I think that was fixed. </p>

bethz: <p>I was able to successfully create an <a href="https://github.com/Microsoft/Recommenders/issues/483" target="_blank">issue </a> :-) </p><p><br></p>

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
anargyri: What about using `Workspace.write_config()` instead of doing this manually?

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
loomlike: Changed to AzureML. For scripts, kept aml*

tests/unit/test_aml_wide_deep_training.py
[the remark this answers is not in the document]
loomlike: Added test

notebooks/00_quick_start/wide_deep_model_movielens.ipynb
loomlike: @miguelgfierro I left the above reply from ReviewNB and as you can see, each line has an extra line breaks there... (this comment is writing from github)

loomlike: Added to the top cell, so that aml notebook also can utilize this wide-deep notebook for its model training/testing via papermill's parameterizing.

loomlike: Added more contexts to the markdown cell

loomlike: No easy way. I just pasted screen-shots

loomlike: Added explanation to a markdown cell

loomlike: Changed. Let me know if new notebook looks ok

loomlike: Please see my reply to the previous comment

loomlike: <p>Please see my reply to the previous comment</p>

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
loomlike: <p>Good question (and good point to discuss further). You should change mostly the first part and last part of the notebook since each model will have different parameters. Also will need to change the parsing script (AzureML's entry script).</p><p><br></p><p>But would be nice to generalize more...</p>

loomlike: Make sense. Removed.

loomlike: Made changes.

loomlike: Removed.

loomlike: As mentioned earlier, that is good topic to discuss. Since different models have different set of hyper-params, it is not possible to generalize 100%, but at least can modularize the hyperparameter parsing part so that can reuse other scripts as much as possible.

We will have more clear idea about how much of the codes can be generalized once we start adding more AzureML examples.

loomlike: Added another wide-deep model results trained with known-hyperparameters in google's [movielens example](https://github.com/tensorflow/models/tree/master/official/wide_deep). Also added conclusion.

Btw, our quickstart notebook uses the hyperparameter we found from this notebook.

notebooks/04_model_select_and_optimize/aml_hyperparameter_tuning.ipynb
loomlike: Added that to the codes. Also updated the documentation markdown cells for that.

notebooks/00_quick_start/wide_deep_movielens.ipynb
```diff
@@ -0,0 +1,1225 @@
+{
```
miguelgfierro: @loomlike we are hosting all the images in a blob to reduce the size of the wiki. Ideally in svg. See an example: https://github.com/Microsoft/Recommenders/blob/master/notebooks/02_model/ncf_deep_dive.ipynb

The location and key is on teams, please let me know if you need help on this.

loomlike: Updated!

reco_utils/aml/wide_deep.py
```diff
@@ -0,0 +1,136 @@
...
+)
+
+
+NOTEBOOK_NAME = os.path.join(
+    "notebooks",
+    "00_quick_start",
+    "wide_deep_movielens.ipynb"
+)
+OUTPUT_NOTEBOOK = "wide_deep.ipynb"
+SUPPORTED_METRICS = {
```
anargyri: I think this list of metrics can be difficult to maintain. When we add new metrics someone needs to remember to modify this file too. Do we need SUPPORTED_METRICS at all? 
I think it is reasonable to expect the user of the repo to pass the correct name of the metric.

notebooks/04_model_select_and_optimize/hypertune_aml_wide_and_deep_quickstart.ipynb
```diff
@@ -0,0 +1,1109 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Prepare all the necessary scripts which will be loaded to our Hyperdrive Experiment Run\n",
+    "SCRIPT_DIR = 'aml_script'\n",
```
anargyri: Could we avoid copying the entire reco_utils subtree under a new directory?

anargyri: Btw I tried `os.symlink()` as a potential workaround but I get a "no privilege" error on Windows.

reco_utils/common/tf_utils.py
```diff
@@ -0,0 +1,408 @@
...
+MODEL_DIR = 'model_checkpoints'
+
+
+def pandas_input_fn(
+    df,
+    y_col=None,
+    batch_size=128,
+    num_epochs=1,
+    shuffle=False,
+    num_threads=1
```
miguelgfierro: I'm usually a fan of using all the cores of a machine. I use this function https://github.com/miguelgfierro/codebase/blob/master/python/system/system_info.py#L76. Is there a reason why is it better to use only 1 thread?

reco_utils/common/tf_utils.py
```diff
@@ -0,0 +1,408 @@
...
+
+    Args:
+        name (str): Optimizer name.
+        lr (float): Learning rate.
+        ftrl_l1_reg (float): Ftrl optimizer l1 regularization rate. Ignored for other optimizers.
+
+    Returns:
+        optimizer (tf.train.Optimizer)
+    """
+    if name == 'Adagrad':
```
miguelgfierro: what about sgd with momentum?

reco_utils/common/tf_utils.py
```diff
@@ -0,0 +1,408 @@
...
+        eval_df,
+        every_n_iter,
+        model_dir,
+        batch_size,
+        eval_fns,
+        **eval_kwargs
+    )
+
+
+class Logger(abc.ABC):
```
miguelgfierro: where is this used?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    Args:
+        user_df (pd.DataFrame): User data containing unique user ids and maybe their features.
+        item_df (pd.DataFrame): Item data containing unique item ids and maybe their features.
+        user_col (str): User id column name.
+        item_col (str): Item id column name.
+        user_item_filter_df (pd.DataFrame): User-item pairs to be used as a filter.
+        shuffle (bool): If True, shuffles the result.
+
+    Returns:
+        (pd.DataFrame) All pairs of user-item from user_df and item_df, excepting the pairs in user_item_filter_df
```
miguelgfierro: the format we are using in the other files is something like:
```
pd.DataFrame: All pairs...
```

reco_utils/evaluation/python_evaluation.py
```diff
@@ -10,14 +10,25 @@
...
 
-from reco_utils.common.constants import (
-    DEFAULT_USER_COL,
-    DEFAULT_ITEM_COL,
-    DEFAULT_RATING_COL,
-    PREDICTION_COL,
-    DEFAULT_K,
-    DEFAULT_THRESHOLD,
-)
+try:
```
miguelgfierro: this try except is weird to me, in which case `from reco_utils.common.constants import ...` is going to raise an exception?

tests/unit/test_aml_wide_deep.py
```diff
@@ -0,0 +1,69 @@
+from reco_utils.aml.wide_deep import parse_args
```
miguelgfierro: I haven't seen many times peoples testing argparse, it is a mature library. If you really want to test that function, maybe you could just test the exceptions.

reco_utils/common/tf_utils.py
```diff
@@ -0,0 +1,408 @@
...
+        batch_size=batch_size,
+        num_epochs=num_epochs,
+        shuffle=shuffle,
+        num_threads=num_threads
+    )
+
+    return input_fn
+
+
+def _build_optimizer(name, lr=0.001, ftrl_l1_reg=0.0):
```
miguelgfierro: these function _build_optimizer, _build_wide_columns, build_model, etc are not tf utils, they are the wide and deep algo. I think we should move this to a new folder in recommenders.

reco_utils/common/tf_utils.py
```diff
@@ -0,0 +1,408 @@
...
+            tf.feature_column.numeric_column(
+                item_feat_col,
+                shape=item_feat_shape,
+                dtype=tf.float32
+            )
+        )
+    return deep_columns
+
+
+def build_model(
```
miguelgfierro: Is this tested in unit tests?

tests/unit/test_notebooks_gpu.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: I think we should add as well smoke and integration tests

tests/unit/test_tf_utils.py
```diff
@@ -0,0 +1,148 @@
...
+        steps=train_steps
+    )
+    _clean_up('deep_' + MODEL_DIR)
+
+    # Check if hook logged the given metric
+    assert rmse.__name__ in evaluation_logger.get_log()
+    assert len(evaluation_logger.get_log()[rmse.__name__]) == hook_frequency
+
+
+def _clean_up(path):
```
miguelgfierro: maybe we can move this to a general files, this could be used by other algos. Also, we can try to find a more descriptive name, remove_folders, remove_path...

reco_utils/aml/wide_deep.py
loomlike: @anargyri That is fair point, but at the same time the current script helps to early-fail (raise an error before we train the model) which is useful especially when a user run remote-compute related work, like AzureML.

Would it be make sense to move this checking to the Notebook which actually imports those metrics and uses? There, the code maintenance would be easier as we will update the module import too.

notebooks/04_model_select_and_optimize/hypertune_aml_wide_and_deep_quickstart.ipynb
loomlike: You can simply pass the repo's base path "Recommender/" instead of copying into a temp path and use the temp path:
* pros: our aml_notebook's script loading cell will become super simple
* cons: all the heavy notebooks and not-necessary tests, non-python files will be uploaded

Otherwise, you have to copy each files the script has dependency on and their folders.

Which why do we prefer?

reco_utils/common/tf_utils.py
loomlike: The thread here is for data reading and enqueueing. Here is the document from TF:

> number of threads used for reading and enqueueing. In order to have predicted and repeatable order of reading and enqueueing, such as in prediction and evaluation mode, num_threads should be 1.

For this reason, the default value is  1 as well as `shuffle=False`. A user can set, e.g., `num_threads=num_cores-1` for training. Will change that on wide-deep notebook.

reco_utils/aml/wide_deep.py
anargyri: I think that would just move the problem to another file. But now you raise another issue, i.e. the import statement in this notebook needs to be maintained too. Also RANKING_METRICS, RATING_METRICS appear in the notebook. Now imagine maintaining this for every notebook that uses hyperdrive. So probably it is better to have an `import *` instead (even though that is discouraged by PEP8, but in this case it may be better) and remove RANKING_METRICS, RATING_METRICS.

That said, I agree that an early check that the metric exists is good to have inside the .py script, because it saves time. What could be done (instead of SUPPORTED_METRICS) is a try statement with a call to `callable(metric)` for each of the tracked metrics.

@miguelgfierro any thoughts?

notebooks/04_model_select_and_optimize/hypertune_aml_wide_and_deep_quickstart.ipynb
anargyri: Neither of the two :-) What I am suggesting is to avoid copying a subtree.  
The way you do it is fine but instead of `shutil.copytree` you copy a single file i.e. wide_deep.py.

reco_utils/aml/wide_deep.py
loomlike: I guess your point applies to all notebooks. They all just use pre-selected metrics (currently, we use all metrics) and calculate each, like this:
```
from reco_utils.evaluation.python_evaluation import (rmse, mae, rsquared, exp_var, map_at_k, ndcg_at_k, precision_at_k, recall_at_k, get_top_k_items)
...
eval_map = map_at_k(test, all_predictions, col_prediction='prediction', k=TOP_K)
eval_ndcg = ndcg_at_k(test, all_predictions, col_prediction='prediction', k=TOP_K)
eval_precision = precision_at_k(test, all_predictions, col_prediction='prediction', k=TOP_K)
eval_recall = recall_at_k(test, all_predictions, col_prediction='prediction', k=TOP_K)

print("MAP:\t%f" % eval_map,
      "NDCG:\t%f" % eval_ndcg,
      "Precision@K:\t%f" % eval_precision,
      "Recall@K:\t%f" % eval_recall, sep='\n')
```
If we introduce new metrics, we anyway need to update those all.

But anyway I like to generalize if there is possible way to do that.
The thing is, for AzureML use-case, we have to somehow pass which metrics we want to track. We'll don't want to track all the metrics when we do 100 runs of Hyperdrive, especially when the data is big and thus ranking take so long.

Any thoughts? @anargyri @miguelgfierro

notebooks/04_model_select_and_optimize/hypertune_aml_wide_and_deep_quickstart.ipynb
loomlike: wide_deep.py calls the model notebook, and the model notebook have dependency on our repo... e.g. tf_utils, movielens, pandas_df_utils, python_splitters, python_evaluation, constants. So better to just upload all repo (except tests and notebooks) to AzureML workspace blob.

reco_utils/common/tf_utils.py
loomlike: Adding couple more optimizers in TF including Momentum to the next commit.

reco_utils/common/tf_utils.py
loomlike: This is for training log hook to record our metrics while training by using any loggers.
Example of loggers we may want to use are AzureML's Run.log, papermill.record, etc.
(main use-case for now is AzureML's Run.log, but didn't want to make the log hook dependent to AzureML)

reco_utils/dataset/pandas_df_utils.py
loomlike: Updating in the next commit

reco_utils/evaluation/python_evaluation.py
loomlike: I think I added this part awhile back when I wanted to upload scripts to AzureML workspace as few as possible... so I tried to remove any tiny module dependencies. But now just I'm uploading whole repo. Will remove this codes.

tests/unit/test_aml_wide_deep.py
loomlike: Removing from the next commit.

tests/unit/test_tf_utils.py
loomlike: oh. that function is not needed anymore. I changed all to directly call `shutil.rmtree` in the next commit.

reco_utils/common/tf_utils.py
loomlike: _build_wide_columns is wide-deep model specific, but not other functions. 
build_optimizer and build_model can be used any TF Estimators like regular linear regressor or DNNs (w/ TF high-level API). Do we still want to move them?

reco_utils/common/tf_utils.py
loomlike: Yes, `test_build_model`.

tests/unit/test_notebooks_gpu.py
loomlike: Working on that.

reco_utils/common/tf_utils.py
miguelgfierro: I would say that if it is something specific to W&D, it goes with it. If it is something that can be used by other TF algos, then it should go in a common dir

miguelgfierro: Maybe this can sever you as an example (even though it is work in progress): https://github.com/Microsoft/Recommenders/blob/benchmark_feb/reco_utils/recommender/fastai/fastai_utils.py

Here I'm just refactoring some functions from fastai that we use in the quick start notebook and in the benchmark. I'm doing a minimal thing, not super complicated wrappers over the fastai model

reco_utils/aml/wide_deep.py
anargyri: That is a good point; we do the same elsewhere. Although I don't think that we maintain a list of SUPPORTED_METRICS anywhere; we refer to each metric explicitly by name. 

So I think what you said (moving the check to the notebook from the python script) is a better option. Another option is to just rely on the programmer to pass a valid METRICS list as an argument to the script.

notebooks/04_model_select_and_optimize/hypertune_aml_wide_and_deep_quickstart.ipynb
anargyri: Why? This is not necessary. It works without doing this.

reco_utils/aml/wide_deep.py
miguelgfierro: In the [benchmarks](https://github.com/Microsoft/Recommenders/blob/benchmark_feb/notebooks/03_evaluate/comparison_movielens.ipynb) (which is work in progress atm) I define which metrics go with which algo. I would say this is more explicit (which is one of our practices). So I would say that to move the check to the notebook makes sense

loomlike: Ok, in the next commit (which I'm going to push soon), the python script doesn't check them and expects users pass valid metrics instead. In the notebooks RANKING_METRICS and RATING_METRICS will contain metrics function names, which are expected to be the same as function.__name__ and thus the actual function will be called by locals()['name'] later cells. Will this better? @anargyri @miguelgfierro

The reason of using those lists is to give users option to select which metrics to track when use AzureML hyperdrive. Ranking takes more time, and the more metrics they use the longer the evaluation takes which end up more cost of cluster compute.

I tried to pass function directly from aml_notebook to model notebook, but papermill complains about "cannot serialize function".

reco_utils/common/tf_utils.py
loomlike: Agreed. moved column and model functions to recommender/wide_deep/wide_deep_utils.py and keep rest to the common tf_utils.py

tests/unit/test_notebooks_gpu.py
loomlike: Added

loomlike:
Thanks all reviewers! I pushed new commits. Last one got a test error on asserting the number of logs in the log-hook while training. Seems tf doesn't call the hook at the final iteration of training which ends up wrong assertion in the test code. So added one more iteration to the total iterations to fix it.

anargyri:
Why do you need to copy `wide_deep_movielens.ipynb` under SCRIPT_DIR? You are still using the .py file in the estimator.

loomlike:
@anargyri It is because aml-notebook uses wide_deep_movielens.ipynb. The aml-script (py) you are referring now just parses Hyperdrive arguments and then calls the wide_deep_movielens.ipynb. My old version of the script (I think your svd-aml work is based on that code) does training and testing from the py script, but it has been changed based on the recipe-work meeting feedback, saying minimize additional scripts for azureML and reuse existing scripts for model training / testing.

If you have a py script for the model, including build, fit & evaluate functions, you can just call those from the aml-script. In wide-deep model case, I don't use any wrapper class for the model and thus re-use wide_deep_movielens.ipynb.

anargyri:
The naming of the tuning notebook and the spark tuning notebook that already exists is a bit confusing. They both contain "hypertune" but one uses Hyperdrive whereas the other does not. Could we make the distinction clearer? Any better naming suggestions? @yueguoguo

anargyri:
Great job, Jun Ki!  sorry for the nitpicking.

Issue #426, opened 2019-01-18
Add Vowpal Wabbit based recommender example


nikhilrj:
### *What* is affected by this issue?
Investigate adding a VW based example for news recommendations or movies recommendations w/ item features

### In *which* platform does it happen?
DSVM / Databricks

### Expected behavior (i.e. solution)
I have an end to end example of training / deploying / operationalizing a recommender system using VW.

gramhagen:
dropping commits into gramhagen/vw_example branch

for now I'm showing a few different approaches using VW (linear regression, multinomial logistic regression, and two matrix factorization approaches)

need to add some more text to the notebook under 02_model/vowpal_wabbit_deep_dive.ipynb, but interim suggestions are welcome

nikhilrj:
Fixed with #452 - awesome work Scott!

PR #452, opened 2019-01-25
Adding VW Example Notebook

gramhagen:
### Description
- Added notebook to demonstrate options for using VW to do regression based and matrix factorization based recommendation
- Adjusted .gitignore to ignore some movielens data that was generated
- Adjusted spark and python evaluation methods to use default values defined in constants.py
- Adjusted python get_top_k_items method to not change original dataframe

Still need to add a test for the notebook

### Motivation and Context
Add Vowpal Wabbit based recommender example

### Releated Issues
#426 

### Checklist:

- [ ] I have added tests.


notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+      "EXP_VAR: 0.22847638017481597\n",
+      "CPU times: user 2.45 s, sys: 35.4 ms, total: 2.48 s\n",
+      "Wall time: 4.51 s\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%time \n",
+    "\n",
+    "train_params = 'vw --rank 5 -qui -f {model} -d {data}'.format(model=model_path, data=train_path)\n",
```
eisber: missing a space at -qui (=-q ui) ?

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "RMSE: 1.764643840355633\n",
+      "MAE: 1.4427961357602124\n",
+      "RSQUARED: -1.490189004814928\n",
```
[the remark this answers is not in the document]
yueguoguo: @eisber nope - r^2 can be negative when variance of residuals is larger than the total variance (see [here](https://en.wikipedia.org/wiki/Coefficient_of_determination)).

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<center><h1>Using Vowpal Wabbit for Recommendations</h1>\n",
+    "\n",
+    "<img src=\"https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/logo_assets/vowpal-wabbits-github-logo.png?raw=true\" height=\"30%\" width=\"30%\" alt=\"Vowpal Wabbit\">\n",
+    "</center>\n",
+    "\n",
+    "[Vowpal Wabbit](https://github.com/VowpalWabbit/vowpal_wabbit) is a fast online machine learning library that implements several algorithms relevant to the recommendation use case.\n",
```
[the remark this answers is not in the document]
miguelgfierro: +1 to this

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+    "\n",
+    "In this notebook we demonstrate how to use the VW library to generate recommendations on the [Movielens](https://grouplens.org/datasets/movielens/) dataset.\n",
+    "\n",
+    "Several things are worth noting in how VW is being used in this notebook. By leveraging an Azure Data Science Virtual Machine ([DSVM](https://azure.microsoft.com/en-us/services/virtual-machines/data-science-virtual-machines/)), VW comes pre-installed and can be used directly from the command line. There are also python bindings to be able to use VW within a python environment and even a wrapper conforming to the SciKit-Learn Estimator API. However, the python bindings must be installed as an additional python package with Boost dependencies, so for simplicity's sake execution of VW is done via a subprocess call mimicking what would happen from the command line execution of the model.\n",
```
miguelgfierro: Right now, we are not installing VW, we are assuming that a person is using a DSVM. How difficult is to add an installation option with the code generator that we have? https://github.com/Microsoft/Recommenders/blob/staging/scripts/generate_conda_file.sh

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+    "from reco_utils.evaluation.python_evaluation import rmse, mae, exp_var, rsquared, get_top_k_items"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def to_vw(df, output, logistic=False):\n",
```
miguelgfierro: not sure if we are going to reuse this function. If we are going to do several notebooks about VW, maybe it would be worthwhile to move this file to a new vw folder under `reco_utils/recommender`

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+      "EXP_VAR: 0.2620255308898497\n",
+      "CPU times: user 1.67 s, sys: 47.3 ms, total: 1.72 s\n",
+      "Wall time: 2.63 s\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%time \n",
+    "\n",
+    "train_params = 'vw -f {model} -d {data}'.format(model=model_path, data=train_path)\n",
```
miguelgfierro: I think here it would be good to mention what is the meaning of each argument and also link to where can a user find other arguments

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "So far we have treated the user features and item features independently, but taking into account interactions between features can provide a mechanism to learn more fine grained preferences of the users.\n",
+    "\n",
+    "To generate interaction features use the quadratic command line argument and specify the namespaces that should be combined: '-q ui' combines the user and item namespaces based on the first letter of each.\n",
+    "\n",
+    "When generating interaction terms one thing to consider is the [hash space](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Feature-Hashing-and-Extraction) used for the features. It can be beneficial to increase the size of the space to reduce unwanted collisions."
```
miguelgfierro: I think it would be good to explain briefly what is the hashing mechanism in VW and what are its benefits

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
miguelgfierro: also, it will be probably good to explain what are these parameters

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+    "<h3>Cleanup</h3>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 17,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tmpdir.cleanup()"
```
miguelgfierro: I think a conclusion would be great. Maybe as @eisber suggested to have a final table with all the results

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+   "metadata": {},
+   "source": [
+    "<center><h1>Using Vowpal Wabbit for Recommendations</h1>\n",
+    "\n",
+    "<img src=\"https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/logo_assets/vowpal-wabbits-github-logo.png?raw=true\" height=\"30%\" width=\"30%\" alt=\"Vowpal Wabbit\">\n",
+    "</center>\n",
+    "\n",
+    "[Vowpal Wabbit](https://github.com/VowpalWabbit/vowpal_wabbit) is a fast online machine learning library that implements several algorithms relevant to the recommendation use case.\n",
+    "\n",
+    "The main advantage of Vowpal Wabbit (VW) is that training is done in an online fashion typically using Stochastic Gradient Descent or similar variants, which allows it to scale well to very large datasets. Additionally, it is optimized to run very quickly and can support distributed training scenarios for extremely large datasets.\n",
```
miguelgfierro: Would it be possible to extend this?
`The main advantage of Vowpal Wabbit (VW) is that training is done in an online fashion typically using Stochastic Gradient Descent or similar variants, which allows it to scale well to very large datasets. Additionally, it is optimized to run very quickly and can support distributed training scenarios for extremely large datasets.`

Some ideas, maybe give an example of a online scenario where VW can be good. In which industries or usecases are people using VW.  In short, why would I use VW as an alternative with other methods

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# load movielens data (use the 1M dataset)\n",
+    "df = load_pandas_df('1m')\n",
```
miguelgfierro: I'm curious about this. Are we using 1m because it is so fast that we don't even need to use 100k?

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+    "\n",
```
miguelgfierro: Just FYI @gramhagen I'm working on explaining how to do tests on notebooks with papermil: https://github.com/Microsoft/Recommenders/issues/455

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+   "source": [
+    "import os\n",
+    "from subprocess import run\n",
+    "from tempfile import TemporaryDirectory\n",
+    "\n",
+    "import pandas as pd\n",
+    "\n",
+    "from reco_utils.dataset.movielens import load_pandas_df\n",
+    "from reco_utils.dataset.python_splitters import python_stratified_split\n",
+    "from reco_utils.evaluation.python_evaluation import rmse, mae, exp_var, rsquared, get_top_k_items"
```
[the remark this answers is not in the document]
miguelgfierro: here we are not using pyspark right?

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def to_vw(df, output, logistic=False):\n",
+    "    \"\"\"Convert Pandas DataFrame to vw input format\n",
+    "    Args:\n",
+    "        df (pd.DataFrame): input DataFrame\n",
+    "        output (str): path to output file\n",
```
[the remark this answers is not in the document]
gramhagen: right, gotit

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+    "        tmp = df.reset_index()\n",
+    "\n",
+    "        # we need to reset the rating type to an integer to simplify the vw formatting\n",
+    "        tmp['rating'] = tmp['rating'].astype('int64')\n",
+    "        \n",
+    "        # convert rating to binary value\n",
+    "        if logistic:\n",
+    "            tmp['rating'] = tmp['rating'].apply(lambda x: 1 if x >= 3 else -1)\n",
+    "        \n",
+    "        for _, row in tmp.iterrows():\n",
```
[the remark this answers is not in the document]
gramhagen: ok, np

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+    "       test_params=test_params, \n",
+    "       test_data=test, \n",
+    "       prediction_path=prediction_path)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A similar alternative is to leverage multinomial classification, which treats each rating value as a distinct class. \n",
```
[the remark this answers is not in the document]
gramhagen: ok

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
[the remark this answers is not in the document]
gramhagen: ah, you're both right! I wasn't converting the ground truth or the prediction output to probabilities

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "RMSE: 1.764643840355633\n",
```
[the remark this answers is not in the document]
gramhagen: ok, sounds cool, not sure how to pull in the timing values, unless you think those aren't that useful (they're all very close)

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: apparently it's equivalent, yeah I can add a little more clarity

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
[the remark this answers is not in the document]
gramhagen: hmmm, i'm not sure if a deep dive into simple regression methods is really all that valuable, but i can put in a little more description of what's happening underneath. My objective here was more about demonstrating functionality within the library vs explaining algorithms that are covered elsewhere. In that way it's a bit different from the other deep dive notebooks.

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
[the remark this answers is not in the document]
gramhagen: ok, yes it's just single node

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+   "source": [
+    "When considering different approaches for solving a problem with machine learning it is helpful to generate a baseline approach to understand how more complex solutions perform across dimensions of performance, time, and resource (memory or cpu) usage.\n",
+    "\n",
+    "One of the most basic approaches for a recommendation engine is to simply learn a linear regression model that is trained on examples of ratings as the target variable and corresponding user ids and movie ids as independent features.\n",
+    "\n",
+    "By passing each user-item rating in as an example the model will begin to learn weights based on average ratings for each user as well as average ratings per item.\n",
+    "\n",
+    "VW uses linear regression by default, so no extra command line options are needed beyond specifying where to locate the model and the data.\n",
+    "\n",
+    "This however generates predicted ratings which are no longer integers, so some additional adjustments can be made at prediction time to convert them back to the integer scale of 1 through 5 if necessary. Here this is done in the evaluate function."
```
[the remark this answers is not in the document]
gramhagen: are you asking if the implementation in this notebook supports tuning? If so, no it is not really set up for that. However, it is possible to modify the run_vw wrapper to support such behavior, you would just need to add additional input to inject hyperparameters and format them in the input string correctly.
Although it would be easier to use the python bindings for this, which allows for the same kind of parameter sweeping that can be done with sk-learn models.

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: installation is a bit tricky due to a dependency on boost libraries. these can be installed via conda, however it is quite challenging to ensure the correct version is leveraged during building vw (currently does not work with python3). this is an ongoing issue with vw and is something I'm working with others on in that repo. Once, it's resolved we can add that to the conda yaml generator.

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: imo this is a bit of a workaround that shouldn't be pushed forward for general use

gramhagen: that said it might be worth considering this relative to output format creation for various libraries, if this is something that would be used to convert a dataframe or pipe a file through it could be generalized.

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: I wanted to make sure the results were comparable to what is being reported in the comparison notebook, and i was thinking the 100k metrics might be noisier than 1m. plus it is quite fast, so why not (other than the time needed to load the data)

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: 👍

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
```diff
@@ -0,0 +1,737 @@
...
+    "<i>Copyright (c) Microsoft Corporation. All rights reserved.</i>\n",
+    "\n",
+    "<i>Licensed under the MIT License.</i>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<center><h1>Using Vowpal Wabbit for Recommendations</h1>\n",
```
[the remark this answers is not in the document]
gramhagen: ok, np

eisber:
Are the item ids all numbers? It’s interesting since you also chose a larger -b. Am starting to wonder if you have any overlap? In theory you should be able to control if you use numerics for the namespace, right?

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: ok

gramhagen:
yes, @eisber you're right, all the user and item features are numeric, there are 6040 unique users, and 3706, unique items, considering interactions, that's 22,393,986 features. I believe when you create interactions the features are hashed, so even though I increased the bit size to 25 -> 2^25 = 33,554,432 the performance is still lower than using bit size of 26. So I'm assuming that is due to hash collisions.

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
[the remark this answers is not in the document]
yueguoguo: I see your point. If it is not about the algorithm itself, it would be useful to see the advantage of VW implementation. Similar to SVD in Surprise and ALS in Spark, the two algorithms are the same in nature (matrix factorization), but each has its own pros and cons, which make it suit for different scenarios.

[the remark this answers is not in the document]
miguelgfierro: agree with both, in the case of VW, a deep dive could be focused on what is the suite of algos that VW has, pros&cons, etc, instead of explaining what is logistic regression or linear regression

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: added a bit more detail here

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
miguelgfierro: makes sense

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
miguelgfierro: can you elaborate a little bit on this? what is the exact bottleneck with boost? when this will be solved? are you directly working on VW repo?

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
miguelgfierro: this is cool. oh, I think we are not adding the benchmark results for the new algos. I added a new issue so we don't forget https://github.com/Microsoft/Recommenders/issues/464

miguelgfierro:
FYI, the tests that are failing should pass after this guys is merged https://github.com/Microsoft/Recommenders/pull/463

notebooks/02_model/vowpal_wabbit_deep_dive.ipynb
gramhagen: sure here are the problems:
- current pip install vowpalwabbit works with system installations of boost on python 2 and 3 
   - pip version is slightly outdated
   - requires system level installation, and potential renaming of libraries
- current repo code changed the python build approach and only works with system installations of boost on python 2 (not 3)
  - i'm gathering notes to explain how to replicate this and post an issue
- neither approach works well with conda, but ideally the boost dependencies could be loaded into a conda env, and vw could be build from conda-forge or through pip under conda (now that conda plays a bit nicer with pip)
 
so my hope is that we can work through these issues in the VW repo (I've been contributing to the python bindings there) and resolve the issues there. but as it stands there's no clear timeline

Issue #237, opened 2018-11-14
Script to automatize the installation of Recommenders repo on Databricks


miguelgfierro:
### *What* is affected by this bug?
Automatize the process of setting up the repo in databricks.

zip folder
```
cd Recommenders
zip -r Recommenders.zip .
```
use databricks cli to upload the zip
install conda on the cluster
install the conda dependencies

### *Where* on the platform does it happen?
*For example:*
 *Azure Databricks.*

### Expected behavior (i.e. solution)
./install_databricks

loomlike:
@miguelgfierro I made and tested the script. Only issue of this approach is, it requires databricks CLI installation and onetime-setup of token. I'm working on updating SETUP.md as well. Will make PR once done.

### Expected behavior (i.e. solution)
* Simply run `./scripts/databricks_install.sh <YOUR_CLUSTER_ID>`

miguelgfierro:
looks great!

PR #438, opened 2019-01-23
Jeremr refarch review

jreynolds01:
### Description

Updates reference architecture documents to reflect two things:

1. recommended configuration is no longer available in the UI so I created a notebook that generates it.
2. makes the als_movies_o16n notebook a little bit more general. e.g. users can change prefix with less fear of obscure azure errors based on character restrictions in service names

### Motivation and Context

1. The reference architecture requires a particular version that is now 'deprecated' in azure databricks. This is a workaround until the cosmosdb connector is updated.
2. Ease of use for the naming changes for the notebook.

### Related Issues

#427 

### Checklist:

- [x ] My code follows the code style of this project, as detailed in our [contribution guidelines](../CONTRIBUTING).
- [ ] I have added tests.
- [x ] I have updated the documentation accordingly.

Test needs to be added/discussed for configuration notebook. Filing pull request for review while test is developed.

PR #457, opened 2019-01-28
Databricks installation script

loomlike:
### Description
* Add repo installation sh for Databricks
* Update SETUP.md accordingly

### Motivation and Context
To make repo installation onto Databricks

### Releated Issues
#237 

### Checklist:
- [ ] I have added tests.


SETUP.md
```diff
@@ -134,24 +133,57 @@ SPARK_WORKER_OPTS="-Dspark.worker.cleanup.enabled=true, -Dspark.worker.cleanup.a
...
+
+Prerequisite
+* Install [Azure Databricks CLI (command-line interface)](https://docs.azuredatabricks.net/user-guide/dev-tools/databricks-cli.html#install-the-cli)
+and setup CLI [authentication](https://docs.azuredatabricks.net/user-guide/dev-tools/databricks-cli.html#set-up-authentication).
+
+1. Start a target cluster and copy the target cluster id. Cluster id can be found with following script:
+    ```
+    databricks clusters list
+    
+    <CLUSTER_ID> <CLUSTER_NAME> <STATUS>
```
gramhagen: probably worth re-iterating that STATUS must be 'RUNNING' for this script to work, if it isn't you can start it with ```databricks cluster start --cluster-id <CLUSTER_ID>```

scripts/databricks_install.sh
```diff
@@ -0,0 +1,25 @@
...
+# This script installs Recommenders into Azure Databricks
+
+echo "Preparing Recommenders library file (egg)..."
+zip -r -q Recommenders.egg . -i *.py -x tests/\* scripts/\*
+
+echo "Uploading to databricks..."
+dbfs cp --overwrite Recommenders.egg dbfs:/FileStore/Recommenders.egg
+
+# Cluster id should be passed by the first argument
+# Cluster id can be found in the URL at https://<databricks-instance>/?o=<16-digit-number>#/setting/clusters/$CLUSTER_ID/configuration.
```
gramhagen: This should match the instructions in SETUP.md for retrieving the cluster id through the cli, doing this from the UI could break if databricks changes how the URL works.

scripts/databricks_install.sh
```diff
@@ -0,0 +1,25 @@
...
+CLUSTER_ID=$1
+
+echo "Installing the library onto databricks cluster $CLUSTER_ID..."
+databricks libraries install --cluster-id $CLUSTER_ID --egg dbfs:/FileStore/Recommenders.egg
+
+databricks libraries cluster-status --cluster-id $CLUSTER_ID
```
gramhagen: this spits out a bunch of JSON, is it possible to parse this just to check the status of just the desired library?

scripts/databricks_install.sh
gramhagen: ha, though I see the databricks CLI provides the same suggestion... /shrug

scripts/databricks_install.sh
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# ---------------------------------------------------------
+# This script installs Recommenders into Azure Databricks
+
+echo "Preparing Recommenders library file (egg)..."
+zip -r -q Recommenders.egg . -i *.py -x tests/\* scripts/\*
+
+echo "Uploading to databricks..."
+dbfs cp --overwrite Recommenders.egg dbfs:/FileStore/Recommenders.egg
```
gramhagen: typically when you use the UI, this is placed under dbfs:/FileStore/jars/<hash>-Recommenders.egg

maybe it's ok to exclude the hash info if the goal is to overwrite it with the current version, but maybe it's a good idea to put it in the jars folder too?

loomlike: Yeah, I intentionally removed 'jars' path to distinguish between the libs uploaded via UIs and this, but make sense to put it back since we don't need to distinguish them. I will put  jars path back.

For hash, yes, the goal is to overwrite so let's not use that.

scripts/databricks_install.sh
loomlike: This was intentional to let users check not just Reco egg installation status but to show all the lib installed on the cluster. Parse and just show the lib name+status is possible. But if we want to show all the other libs status too, parsing gets complicated as the JSON has different keys /structure for different types of libs, e.g. pypi and egg. I rather keep the current form if it is okay.

scripts/databricks_install.sh
loomlike: lol yeah. As using cli is easier, will update this to be the same as SETUP.md

scripts/databricks_install.sh
gramhagen: sure, just curious what options there are, but it looks like the cli can't filter or change the output format, so this is good to check them all I guess

scripts/databricks_install.sh
loomlike: Updated the script. Replaced code comments to printing out messages.

scripts/databricks_install.sh
loomlike: Changed

SETUP.md
loomlike: Oh. thought this comment is for script. Will update the SETUP.md right now!

jreynolds01:
Agreed with Graham re: python vs. shell. 

This does a small part of the create and config notebook in the pull request here: https://github.com/Microsoft/Recommenders/pull/438. 

It assumes a supported version of ADB is running on the cluster (4.1). Using the `.sh` and CLI approach to do this is fine, but would like to see that this is extensible to support the full set up, including cluster creation and potentially preparing for operationalization (which includes additional library installation, including the cosmosdb-spark connector).

Re: dependencies, it requires zip is installed (usually, that's available, but I didn't have access to it by default within a clean Ubuntu from a Windows subsystem for linux). I don't know if we want to mention this or not.

yueguoguo:
> Do we want 1 setup script for adb that installs only the reco_utils module and then a second setup script that facilitates the preparation for operationalization. Or do we want to 1 setup script that can install all dependencies end-to-end depending on flags. Which do we think is easier for the user?

The former is preferred by myself. It is more modular and more flexible.

SETUP.md
```diff
@@ -134,24 +133,57 @@ SPARK_WORKER_OPTS="-Dspark.worker.cleanup.enabled=true, -Dspark.worker.cleanup.a
...
+3. Once your cluster has started, go to the Databricks home workspace, then go to your user and press import.
+4. In the next menu there is an option to import a library, it says: `To import a library, such as a jar or egg, click here`. Press click here.
+5. Then, at the first drop-down menu, mark the option `Upload Python egg or PyPI`.
+6. Then press on `Drop library egg here to upload` and select the the file `Recommenders.egg` you just created.
+7. Then press `Create library`. This will upload the zip and make it available in your workspace.
+8. Finally, in the next menu, attach the library to your cluster.
+
+</details>
+
+To make sure it works, you can now create a new notebook and import the utilities from Databricks:
```
yueguoguo: It might be good to mention that the script does not handle dependencies.

jreynolds01:
Currently testing an addition to do the second part of installation to prepare for operationalization.

loomlike:
Just had quick chat w/ @jreynolds01. We can utilize this script within the create_and_configure_(databricks)_cluster notebook. So that this script can serve those who already have Databricks running and want to install our repo on it, while Jeremy's notebook serves more o18n forcused use-cases.

SETUP.md
loomlike: Updated the SETUP

loomlike:
Last commit addresses the bash script issue when using Windows' git-bash

jreynolds01:
@loomlike - can you do a check of the script I just added for o16n?

SETUP.md
```diff
@@ -131,29 +139,131 @@ SPARK_WORKER_OPTS="-Dspark.worker.cleanup.enabled=true, -Dspark.worker.cleanup.a
...
+```{python}
 import reco_utils
 ```
 
-### Troubleshooting for Azure Databricks
+### Troubleshooting Installation on Azure Databricks
+
 * For the [reco_utils](reco_utils) import to work on Databricks, it is important to zip the content correctly. The zip has to be performed inside the Recommenders folder, if you zip directly above the Recommenders folder, it won't work.
 
+## Prepare Azure Databricks for Operationalization
```
jreynolds01: @nikhilrj - please take a look at this SETUP.

nikhilrj: LGTM!

loomlike: @nikhilrj hmmm good point. But basically the install-script does the same thing as the manual steps... Maybe we can explain what the script does instead of having those contents as 'manual installation'?

SETUP.md
```diff
@@ -12,38 +15,41 @@ In this guide we show how to setup all the dependencies to run the notebooks of
...
+Currently, this repository supports the following on a linux DSVM:
 
-Environments supported to run the notebooks on the Linux DSVM:
 * Python CPU
 * Python GPU
 * PySpark
-Environments supported to run the notebooks on Azure Databricks:
-* PySpark
+
+PySpark is the only supported environment on Azure Databricks.
```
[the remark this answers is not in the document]
loomlike: @jreynolds01 We can use PythonCPU as well as GPU. I even read a Databricks doc describing about how to use Tensorflow-gpu on Databricks, where you can set a single-node Databricks gpu-cluster for that.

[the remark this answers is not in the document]
loomlike: https://docs.databricks.com/applications/deep-learning/tensorflow.html This is from Databricks, but Azure Databricks is the same.

SETUP.md
loomlike: Just read your comment below. I agree the SETUP is a bit long now... what's other's thought?

SETUP.md
```diff
@@ -131,29 +139,131 @@ SPARK_WORKER_OPTS="-Dspark.worker.cleanup.enabled=true, -Dspark.worker.cleanup.a
...
+### Repository installation
+You can setup the repository as a library on Databricks either manually or by running an [installation script](scripts/databricks_install.sh). Both options assume you have access to a provisioned Databricks workspace and cluster and that you have appropriate permissions to install libraries.
+
+<details>
+<summary><strong><em>Quick install</em></strong></summary>
+
+This option utilizes an installation script to do the setup, and it requires additional dependencies in the environment used to execute the script.
+
+> To run the script, following **prerequisites** are required:
+> * Install [Azure Databricks CLI (command-line interface)](https://docs.azuredatabricks.net/user-guide/dev-tools/databricks-cli.html#install-the-cli) and setup CLI authentication. Please find details about how to create a token and set authentication [here](https://docs.azuredatabricks.net/user-guide/dev-tools/databricks-cli.html#set-up-authentication). Very briefly, you can install and configure your environment with the following commands.
```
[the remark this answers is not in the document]
jreynolds01: yeah, we know how to do that, but it's a little more complicated given the payloads the CLI needs, so lets hold onto that for next.

SETUP.md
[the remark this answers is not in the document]
jreynolds01: fixed.

SETUP.md
jreynolds01: i think it's long, but i think the benefits of having all setup in 1 place is worth it. I think we should provide manual information in case someone can't run the scripts for some reason.

However, there are a few different ways to do that:
* Say something like see the scripts and comments in the scripts for manual dependencies, and include links to documentation on how to add libraries, etc. The scripts do implement all the steps, so that is in some ways self-documenting
* Another option would be to have a separate SETUP_MANUAL.md file, and use that as a an Appendix of sorts, where we could reference in the default SETUP.md.

I think if we want to clean it up first bullet is probably a good way to do it. I'd say let's go ahead and merge, and have that as a follow-up action.

Issue #460, opened 2019-01-29
Update conda script to take spark version as an optional argument


jreynolds01:
### *What* is affected by this bug?

Goal of this feature request is to make it easy to generate a new conda environment with a different version of spark to test different versions easily.

* Testing
* Running on DSVM

### In *which* platform does it happen?
*For example:*
* *Azure Data Science Virtual Machine.*
* Build Agents for testing

### *How* do we replicate the issue?

The base version of pyspark supported by the conda generation script is 2.3.1. Examine the `scripts/generate_conda_file.sh` file.

### Expected behavior (i.e. solution)

Add an additional argument that is parsed to allow for flexible spark version at run-time.

Issue #427, opened 2019-01-18
Unable to create an appropriately versioned cluster per instructions in reference architecture and als_movie_o16n


jreynolds01:
### *What* is affected by this bug?

* *Creating an appropriate cluster*
* *Running a notebook on that cluster*
* *Unit tests.*

### In *which* platform does it happen?
* *Azure Databricks.*

### *How* do we replicate the issue?

1. *Create a databricks workspace*
2. *Navigate to Clusters*
3. *Click [+Create Cluster]* In the Databricks Runtime Version, there is no longer an option for DB 4.1, Spark 2.3.0. It was deprecated on 2019-01-17. See deprecation schedule [here](https://docs.azuredatabricks.net/release-notes/runtime/databricks-runtime-ver.html#deprecation-policy).

### Expected behavior (i.e. solution)

Workarounds:

- It is still possible to create a cluster by cloning a cluster of the recommended version.
- It is still possible to create a cluster through the API. Happy to do PR with appropriate json for creating with databricks CLI or directly through the REST API.

### Other Comments

Have we tested whether the cosmosdb connector jar works with more current versions of ADB and spark?

yueguoguo:
> It is still possible to create a cluster through the API. Happy to do PR with appropriate json for creating with databricks CLI or directly through the REST API.

This would be awesome. 

> Have we tested whether the cosmosdb connector jar works with more current versions of ADB and spark?

It was mentioned [here](https://github.com/Azure/azure-cosmosdb-spark#requirements) that which of the versions are supported.

jreynolds01:
ok, have an initial draft working here. https://github.com/jreynolds01/Recommenders/blob/jeremr-refarch-review/notebooks/04_operationalize/create_and_configure_cluster.ipynb 

Will file a PR after going through the walkthrough again to make sure that's clear.

jreynolds01:
Do you have a recommended way of writing a test for something like this? I can use papermill to run the notebook, but the biggest issue I don't know how to deal with is in dealing with the secrets associated with the DB workspace.

miguelgfierro:
As of now we don't have an example of testing something on databricks, I know that devops recently added a way to do CICD with it https://docs.azuredatabricks.net/user-guide/notebooks/azure-devops-services-version-control.html. That should allow one to manage secrets

yueguoguo:
[...]

PR #461, opened 2019-01-29
Parameter to control pyspark version in conda

jreynolds01:
### Description
Updated generate_conda_file.sh to take an optional argument to specify the version of spark to install.

### Motivation and Context

We are doing a lot of testing across different versions of spark due to databricks deprecation schedule. This should facilitate that while we also get databricks testing up and running.

https://github.com/Microsoft/Recommenders/issues/460

### Related Issues

https://github.com/Microsoft/Recommenders/issues/427

### Checklist:
- [ ] I have added tests.

### DIFF
--- a/scripts/generate_conda_file.sh
+++ b/scripts/generate_conda_file.sh
@@ -91,8 +91,6 @@ fi
 # $ python -m ipykernel install --user --name my_env_name --display-name "Python (my_env_name)"
 #
 
-name: recommenders
-
 channels:
 - conda-forge
 - pytorch
--- a/scripts/generate_conda_file.sh
+++ b/scripts/generate_conda_file.sh
@@ -9,7 +9,7 @@
 # $ sh generate_conda_file.sh --pyspark
 # For generating a conda file for running python gpu and pyspark:
 # $ sh generate_conda_file.sh --gpu --pyspark
-# For generating a conda file for running python gpu and pyspark a particular version of spark:
+# For generating a conda file for running python gpu and pyspark with a particular version:
 # $ sh generate_conda_file.sh --gpu --pyspark-version 2.4.0
 #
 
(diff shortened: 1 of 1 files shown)


scripts/generate_conda_file.sh
```diff
@@ -9,6 +9,8 @@
 # $ sh generate_conda_file.sh --pyspark
 # For generating a conda file for running python gpu and pyspark:
 # $ sh generate_conda_file.sh --gpu --pyspark
+# For generating a conda file for running python gpu and pyspark with a particular version:
```
yueguoguo: We had a discussion on this to support DSVM - we need to match the Spark version available on DSVM but it is updating every now and then. @maxkazmsft

scripts/generate_conda_file.sh
```diff
@@ -73,6 +91,8 @@ fi
 # $ python -m ipykernel install --user --name my_env_name --display-name "Python (my_env_name)"
 #
 
+name: recommenders
```
miguelgfierro: does this fix the name? we don't want this, since most users will have several environments. In the documentation we give a different name to each environment: reco_bare, reco_gpu, etc

jreynolds01: fixed.

scripts/generate_conda_file.sh
jreynolds01: Do you think the default version to be the version current on DSVM?

yueguoguo: Hmmm yea because we encourage users in the repo to try out notebooks on a Linux DSVM where some tools/software (e.g., Spark) have been pre-installed.

jreynolds01: I think we should keep this PR as consistent with the prior version used (2.3.1), and then have an additional update that tests against a newer version of spark. Changing this code then also updating the spark version is doing two different things. @miguelgfierro - what do you think?

jreynolds01: resolving to take this approach after talking with @miguelgfierro off-line.

PR #476, opened 2019-01-31
Deeprec/second

anargyri:
### Description
Added Readme


anargyri:
> Maybe it would be good to link the papers of NCF, xDeepFM and DKN

Updated.

Issue #465, opened 2019-01-30
Investigate difference between new and 0.1.1 SAR versions when using movielens 10M


miguelgfierro:
### *What* is affected by this bug?
In the SAR quickstart notebook, when using movielens 10M, there is a significat difference in the metrics and in the compute time

For SAR in release 0.1.1, the metrics are:
```
Took 109.8672547340393 seconds for training.
Took 9.996008157730103 seconds for prediction.
Model:	sar_ref
Top K:	10
MAP:	0.101402
NDCG:	0.321073
Precision@K:	0.275766
Recall@K:	0.156483
```
The new SAR in staging:
```
Took 563.519758939743 seconds for training.
Took 8.40031886100769 seconds for prediction.
Model:	sar_ref
Top K:	10
MAP:	0.066775
NDCG:	0.249102
Precision@K:	0.216268
Recall@K:	0.106777
```

### In *which* platform does it happen?
* *Azure Data Science Virtual Machine.*

Steps for both versions:
- [ ] Check the results with 100k, 1M and 20M
- [x] Check that the unit tests are passing

miguelgfierro:
Results of new and old SAR with **ML1M**
For SAR in release 0.1.1, the metrics are:
```
Took 9.09568476676941 seconds for training.
Took 0.3207972049713135 seconds for prediction.
Model:	sar_ref
Top K:	10
MAP:	0.064013
NDCG:	0.308012
Precision@K:	0.277215
Recall@K:	0.109292
```

The new SAR in staging:
```
Took 5.787659168243408 seconds for training.
Took 0.4074389934539795 seconds for prediction.
Model:	sar_ref
Top K:	10
MAP:	0.064013
NDCG:	0.308012
Precision@K:	0.277215
Recall@K:	0.109292
```

gramhagen:
just confirming that we see the same results despite changes to the notebook.
I took a python export of the notebook in staging and modified it to run through all the data set sizes then ran it on current staging as well as release 0.1.1 and the results match what you get. So it's definitely a change in the code itself. I'll look a bit closer on what has changed in the sar model.

gramhagen:
line profiling the two version of the fit() method reveals a difference in the dot() calculation

| Version | Line | Time (us) | Time (%) | Line Contents |
|----------|-----|------------|-----------|----------------|
| Staging | 403 | 582523501.0 | 92.0 | self.scores = self.user_affinity.dot(self.item_similarity) |
[...]

Issue #464, opened 2019-01-29
Add movielens 1M benchmark results of the new algos


miguelgfierro:
### *What* is affected by this bug?
We have results for SAR, ALS and SVD in a table in the readme, we should add the rest of the algos.

gramhagen:
This should be for 1M dataset? That's at least what the text in the README says, if that's not right I can correct that.

miguelgfierro:
oh good catch, I'll rename the title

miguelgfierro:
I'm working on changing the benchmark notebook. As of now it is just running the notebooks with papermill and parsing the results. However, the data splits are not the same across all notebooks. I will remake it to have more consistency

miguelgfierro:
Progress, code finished and metrics checked:
- [ ] SVD
- [ ] ALS
- [ ] SAR
- [ ] fastai
- [ ] ncf
- [ ] rbm

miguelgfierro:
FastAI is giving some weird metrics:
```
Ranking metrics: 
{
    "MAP": 213457.64543650797,
    "Precision@k": 25008.0,
    "Recall@k": 10.0,
    "nDCG@k": 25008.00000005573
}
```

miguelgfierro:
fixed with https://github.com/Microsoft/Recommenders/pull/715

PR #485, opened 2019-02-05
Gramhagen/sar single node cleanup

gramhagen:
### Description
Simplification and cleanup of code in reference implementation of sar single node
- using sparse csr format for user item affinity
- reverting jaccard and lift calculation for speed (hiding numpy warnings)
- using dense np.matrix for item-cooccurrence and computed similarities
- moving set_index functionality inside of fit
   - note: this does introduce an edge case that must be handled if the user tries to score unseen items, currently it predicts a score of 0
- updating notebooks to remove set_index functionality
- updating test code to remove set_index functionality
- removing time logging as this can be done using line_profiler if needed
- removing commented out experimental code in favor of simple clean reference implementation

### Motivation and Context
This addresses issues encountered with movielens results discrepancy (#465) and also makes the code easier to read as well as simplifies running the algorithm (by removing external indexing)

### Checklist:

- [X] I have added tests.


reco_utils/evaluation/python_evaluation.py
```diff
@@ -582,10 +582,6 @@ def get_top_k_items(dataframe, col_user=DEFAULT_USER_COL, col_rating=DEFAULT_RAT
...
-    tmp = dataframe.copy()
-    tmp[col_rating] = tmp[col_rating].astype(float)
-    return (
-        tmp.groupby(col_user, as_index=False)
-        .apply(lambda x: x.nlargest(k, col_rating))
-        .reset_index()
-    )
+    return (dataframe.groupby(col_user, as_index=False)
+            .apply(lambda x: x.nlargest(k, col_rating))
+            .reset_index(drop=True))
```
anargyri: This changes the returned result, doesn't it? (removes the user info)

notebooks/00_quick_start/sar_single_node_movielens.ipynb
```diff
@@ -41,7 +41,7 @@
   },
```
anargyri: <p>This looks better after the refactoring  of the indexing. However, I think it is good to keep the logging of the progress of the algorithm.</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/485/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

notebooks/02_model/sar_single_node_deep_dive.ipynb
```diff
@@ -87,7 +87,7 @@
   },
```
anargyri: <p>Again, may be better to keep logging something here.</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/485/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

notebooks/00_quick_start/sar_single_node_movielens.ipynb
miguelgfierro: <p>mmm can you check this? if we are using the test set, shouldn't we have users close to 600 (like in the previous commit= instead of 1?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/485/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

notebooks/02_model/sar_single_node_deep_dive.ipynb
miguelgfierro: <p>same as before, not sure if we are using the correct test set</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/485/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

reco_utils/evaluation/python_evaluation.py
gramhagen: The returned results will only have the user col and a rating col which has the top-k list. But the original dataframe should be untouched. Is that the desired output format?

notebooks/00_quick_start/sar_single_node_movielens.ipynb
gramhagen: <p>I'm not sure I understand, do you mean surfacing the info level logs from sar_single_node in the notebook? Or just delineating how much time is taking for creating the indices vs the other portion of the algorithm?</p>

reco_utils/evaluation/python_evaluation.py
anargyri: Ah sorry, I see. It drops the useless index.

notebooks/00_quick_start/sar_single_node_movielens.ipynb
anargyri: Not the time, just the info about what part of the algorithm is under progress i.e.
"Collecting user affinity matrix... 
Calculating time-decayed affinities... 
Creating index columns... 
Building user affinity sparse matrix... 
Calculating item cooccurrence... 
Calculating item similarity... 
Calculating jaccard... 
Calculating recommendation scores... 
done training"

reco_utils/common/python_utils.py
```diff
@@ -2,29 +2,50 @@
...
+    """Compute decay factor for a given value based on an exponential decay
+    Args:
+        value (numeric): value to calculate decay factor
+        max_val (numeric): value at which decay factor will be 1
+        half_life (numeric): value at which decay factor will be 0.5
+    Returns:
+        (float): decay factor
+    """
+
+    return np.exp(-np.log(2) * (max_val - value) / half_life)
```
miguelgfierro: I have a nice piece of code for you. Can you try this and see if it is faster on SAR?

```
def exponential_decay(value, max_val, half_life):
    return np.exp(-np.log(2) * (max_val - value) / half_life)

def exponential_decay_ne(value, max_val, half_life):
    return ne.evaluate("exp(-log(2)*(max_val - value) / half_life)")

s = 1000000
aa = np.random.randn(s, 1).astype(np.float32)
e1 = exponential_decay(aa, 5, 2)
e2 = exponential_decay_ne(aa, 5, 2)
np.allclose(e1, e2) #True
%timeit exponential_decay(aa, 5, 2) #11.2 ms ± 22.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
%timeit exponential_decay_ne(aa, 5, 2) #2.27 ms ± 57.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

reco_utils/common/python_utils.py
```diff
@@ -2,29 +2,50 @@
 from scipy.sparse import coo_matrix
 
 
+def exponential_decay(value, max_val, half_life):
+    """Compute decay factor for a given value based on an exponential decay
+    Args:
+        value (numeric): value to calculate decay factor
```
miguelgfierro: We are not using the notation numeric in the rest of the code, we use np.array, float, int...

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -50,533 +28,331 @@ class SARSingleNode:
...
-            log.info("Removing seen items...")
-            scores_dense[common_index[:, 0], common_index[:, 1]] = 0
-
-        final_scores = scores_dense[test_index[:, 0], test_index[:, 1]]
-
-        results = pd.DataFrame.from_dict(
+
+        # get user / item indices from test set
+        user_ids = test[self.col_user].map(self.user2index).values
+        assert not any(
```
miguelgfierro: instead of having an assert, can you throw an exception?

notebooks/00_quick_start/sar_single_node_movielens.ipynb
gramhagen: ok, np

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: why are exceptions preferred? (I'm fine changing it, just curious about the difference)

reco_utils/common/python_utils.py
gramhagen: 👍

reco_utils/common/python_utils.py
```diff
@@ -2,29 +2,50 @@
...
+        max_val (numeric): value at which decay factor will be 1
+        half_life (numeric): value at which decay factor will be 0.5
+    Returns:
+        (float): decay factor
```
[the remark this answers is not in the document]
gramhagen: it occurs to me that if max_val < max(x) this will actually exponentially <i>increase</i> the value which doesn't seem like the desired behavior, should those values be left as is or set as zero?

reco_utils/recommender/sar/sar_singlenode.py
miguelgfierro: it is a good practice that we are following in the rest of the code, that way we can control how to catch them in case it is necessary. 

Normally what we do is, throw exceptions for code and use asserts for tests

reco_utils/common/python_utils.py
miguelgfierro: is it faster?

gramhagen: for ML-1M
numexpr takes: 11.911 ms
numpy takes: 75.956 ms

I wish they would just merge numexpr with numpy!

miguelgfierro: awesome!!!! do you mind to get the metrics for 10M and 20M?

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: ok, np

notebooks/02_model/sar_single_node_deep_dive.ipynb
gramhagen: curious, let me look into it

reco_utils/common/python_utils.py
gramhagen: | Dataset | Impl | Time (ms)|
|-|-|-|
| 1M | numexpr | 47.705 |
| 1M | numpy | 19.273 |
| 10M | numexpr | 264.781 |
| 10M | numpy | 138.637 |
| 20M | numexpr | 383.995 |
| 20M | numpy | 235.256 |

Oops, I think I posted the last results backwards, sorry!

ugh, at this point though it seems like I should revert to the original numpy code

miguelgfierro: mmm, this is interesting, I wonder why this is the behavior. numexpr should be faster than numpy

anargyri:
Btw I have not looked at all at the other methods (`recommend_k_items(), predict()`), so there may be improvements that can be made on these too.

gramhagen:
@miguelgfierro @anargyri do you want to look at anything further here or are we good with the changes?

reco_utils/common/python_utils.py
gramhagen: i wonder if it has something to do with how the data is being passed in, since we're passing in a pandas series numpy should be able to vectorize the computation, maybe numexpr is looping over each value?

miguelgfierro:
@gramhagen I would like to take a look at why numepr is not working as expected. Also, it would be great to take a look at the other methods as @anargyri suggested

gramhagen:
I did look at get_top_k and recommend, there are some small clean ups there, but the implementation for getting top k using np.argpartition is pretty efficient, so while I could make it simpler, I can't make it faster, so I left that part the same.

Based on feedback for remove seen, I think @anargyri is right it makes sense to be able to optionally remove seen items at prediction time so that's a change that would need to happen there.

Let me look a bit more at numexpr and I'll let you know what I find. But I don't want to delay this too long. We can always adjust it later.

anargyri:
> I did look at get_top_k and recommend, there are some small clean ups there, but the implementation for getting top k using np.argpartition is pretty efficient, so while I could make it simpler, I can't make it faster, so I left that part the same.
> 
> Based on feedback for remove seen, I think @anargyri is right it makes sense to be able to optionally remove seen items at prediction time so that's a change that would need to happen there.
> 
> Let me look a bit more at numexpr and I'll let you know what I find. But I don't want to delay this too long. We can always adjust it later.

Let's leave the possible enhancements (remove_seen, top_k) as separate PRs. About the top k, my main concern is whether it still goes OOM for Movielens20M (on a large enough VM); we should be able to fit the whole user-items matrix of scores in memory.

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -50,533 +27,329 @@ class SARSingleNode:
...
-        if not np.issubdtype(float_type, np.floating):
-            raise ValueError(
-                "Only floating point data types are accepted for the rating column. Data type was {} "
-                "instead.".format(float_type)
-            )
+        Args:
+            df (pd.DataFrame): User item rating dataframe
+        """
+
+        # Map a continuous index to user / item ids
```
miguelgfierro: hey @gramhagen can we refactor the data indexing in a different method from fit? I need this for having consistency with the other algos when doing the benchmark https://github.com/Microsoft/Recommenders/issues/464. Ideally, I would need something like this:
```
index_df = index_data(df)
model = SAR()
model.fit(index_df)
```
or  in case the index needs to maintain state from SAR
```
model = SAR()
index_df = model.index_data(df)
model.fit(index_df)
```

if you think this will take a long time or if you don't have time for doing this, we can add an issue and any of us can make this work on a different PR.

gramhagen: what are you using index_df for outside the model?

miguelgfierro: I'm timing all the training method and for the rest of the algos, I'm not taking into account the data transformations they make before training. Here you can see a partial work https://github.com/Microsoft/Recommenders/blob/benchmark_feb/notebooks/03_evaluate/comparison_movielens.ipynb

gramhagen: Hmm, I think it's fair to consider this as part of the algorithm since it is required for training / prediction. Separating it into another method just adds an unnecessary step for the user to remember.

gramhagen:
@anargyri on a DS14_v2 (112GB RAM) I am able to train and run recommend_top_k on ML 20M without running out of memory, (on 56GB this doesn't work though)

reco_utils/common/python_utils.py
gramhagen: so i looked at evaluating those two expressions on synthetic data, numexpr has an initial cost to compile / start up which means it needs a larger data size to out-perform numpy. 

This [thread](https://github.com/pydata/numexpr/issues/301#issuecomment-388097698) seems to indicate that size is in the 100k range, but from my test it's closer to the 100M range, which we're not reaching with these datasets.

I can send the evaluation code if you want to duplicate it (for some reason it won't let me attach the file here), not sure if the versions of numpy and numexpr have anything to do with it Maybe optimizations happening that were not there in older versions of numpy?

numpy_vs_numexpr

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: k, moved this into a separate method (set_index) which is called by fit if it hasn't been called externally, so if you want to time it without including the indexing (which is a pretty small amount of time relatively) you can. And if it's not called fit will take care of it. Everybody wins! =)

miguelgfierro: that idea is really great.

anargyri:
> @anargyri on a DS14_v2 (112GB RAM) I am able to train and run recommend_top_k on ML 20M without running out of memory, (on 56GB this doesn't work though)

Brilliant! this somehow got fixed since I last checked!

reco_utils/recommender/sar/sar_singlenode.py
anargyri: I don't think we use indexing in an algorithm other than SAR. But we do require data in different formats e.g. DeepRec accepts LIBFFM format. I agree with @gramhagen that this indexing is internal and can be hidden from the user. Otherwise, if we think it should be part of the data format and other algos will use it, we should make it available outside the scope of SAR, as a dataset util.

PR #486, opened 2019-02-05
Gramhagen/mock test movielens

gramhagen:
### Description
Mocking movielens unit test download

### Motivation and Context
Current test_movielens.py invokes http download of data files to build movielens dataset, this slows down unit tests and requires internet connectivity, with this change example data is stored locally for ml 100k and ml 1m formats and the unit test mocks out the http download to pass the path to the local test resource.
Note - this data is not real movielens data, it's a small set of synthetic data in the same format so there are no licensing issues with retaining / distributing it.

### Releated Issues
#471 

### Checklist:

- [X] I have added tests.


tests/unit/test_movielens.py
```diff
@@ -1,8 +1,11 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+from unittest.mock import patch
+
 import pytest
 from reco_utils.dataset import movielens
+from reco_utils.dataset.movielens import _load_datafile
```
miguelgfierro: if we are importing a function, we should make it public

tests/unit/test_movielens.py
```diff
@@ -1,8 +1,11 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: so right now we are mocking the download behavior and are not downloading 100k and 1M anymore. How can we check that these datasets still work?

miguelgfierro: One idea could be to check that the files exist (without download them) and then use the smoke and integration tests to actually download them and test them. Maybe we can move 100k and 1M to smoke and 10 and 20M to integration. 

Any other ideas?

gramhagen: do any smoke tests use the 1M dataset? ideally not, in which case I would be fine moving the 100k test which downloads the data to smoke and the larger ones to integration.

the main thing is I want to be able to run the unit test suite around 1 min or less, and I'd like to be able to do it if I don't have internet access.

with this mocking we will ensure that load_pandas_df and load_spark_df is doing the right thing if we give it properly formatted files (at least for 100k, and 1M formats), if that format changes on MovieLens side then this particular test wouldn't catch it (however integration tests should)

miguelgfierro: I checked and as you said we are only using 100k on smoke, so this makes sense:

> moving the 100k test which downloads the data to smoke and the larger ones to integration.

gramhagen: k, reverting changes

miguelgfierro:
@gramhagen for this PR, would it make sense to wait until CELA answers the question of hosting datasets on a blob?

gramhagen:
I was thinking of just reverting this pr to simply the switch to smoke / integration. then we can add another pr to move to blob. I thought I pushed that commit, but apparently not. Let me retry.

miguelgfierro:
ok that makes sense

tests/unit/test_movielens.py
anargyri: About lack of internet access, would it make sense if there was a marker so that you could have the option to run the mock tests with this flag and otherwise run the tests we had before?
Also, why are there 3 samples for the Spark test and 100000 for the Pandas test?

gramhagen:
ok, all set now. I think we should push this in. And we can switch to using blob later if possible.

PR #487, opened 2019-02-05
Readme / Comparison Updates

### LINKED ISSUE #474: Update README
We have to update the README for the new algos that are in the repo. We also have to update the benchmark results.

@gramhagen could you do VW?

I will do FastAI and NCF

The list of the new algos that we have is:
- [x]  VW
- [x]  NCF
- [x]  FastAI
- [x]  xDeepFM
- [x]  DKN
- [x]  RBM

nikhilrj:
### Description
- Update Readme to highlight new 0.2 algos
- Update comparison notebook to run new notebooks

### Motivation and Context
- So a user can easily find all algorithms available in the repo

### Releated Issues
Fixes #474 

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ In addition, we also provide a [comparison notebook](notebooks/03_evaluate/compa
 | [SVD](notebooks/02_model/surprise_svd_deep_dive.ipynb) | 0.010915 | 0.102398 | 0.092996 | 0.025362 | 0.888991 | 0.696781 | 0.364178 | 0.364178 | 
 | [FastAI](notebooks/00_quick_start/fastai_recommendation.ipynb) | 0.023022 |0.168714 |0.154761 |0.050153 |0.887224 |0.705609 |0.371552 |0.374281 |
 | [SAR](notebooks/00_quick_start/sar_single_node_movielens.ipynb) | 0.064013 | 0.308012 | 0.277215 | 0.109292 | N/A | N/A | N/A | N/A 
-| [NCF](notebooks/00_quick_start/ncf_movielens.ipynb) | 0.026364 | 0.157324 | 0.146258 | 0.058225 | NaN | NaN | NaN | NaN
-| [RBM](notebooks/00_quick_start/rbm_movielens.ipynb) | 0.203125 | 0.582515 | 0.490083 | 0.249196 | NaN | NaN | NaN | NaN
+| [NCF](notebooks/00_quick_start/ncf_movielens.ipynb) | 0.026364 | 0.157324 | 0.146258 | 0.058225 | N/A | N/A | N/A | N/A
+| [RBM](notebooks/00_quick_start/rbm_movielens.ipynb) | 0.203125 | 0.582515 | 0.490083 | 0.249196 | N/A | N/A | N/A | N/A
 
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -52,6 +52,26 @@ We provide several notebooks to show how recommendation algorithms can be design
 
 - The [Operationalizion Notebook](notebooks/05_operationalize) demonstrates how to deploy models in production systems.
 
+
+The Quick-Start and Modeling notebooks showcase how to utilize the following algorithms to build a recommender system:
+
+**Algorithms**
+
+| Algorithm | Environment | Description | 
+| --- | --- | --- |
+| **`Classic Recommenders`** |
+| [Surprise/Singular Value Decomposition (SVD)](notebooks/00_quick_start/sar_single_node_movielens.ipynb) | Python |  | 
+| [Alternating Least Squares (ALS)](notebooks/00_quick_start/als_pyspark_movielens.ipynb) | Spark |  |
+| **`Microsoft Recommenders`** |
+| [Smart Adaptive Recommendations (SAR)](notebooks/00_quick_start/sar_single_node_movielens.ipynb) | Python / Spark | |
+| [Vowpal Wabbit Family (VW)](notebooks/02_model/vowpal_wabbit_deep_dive.ipynb) | Python / Online | |
+| **`Deep Learning`** |
+| [eXtreme Deep Factorization Machine (xDeepFM)](notebooks/00_quick_start/xdeepfm.ipynb) | Python / GPU | | 
+| [Deep Knowledge-Aware Network (DKN)](notebooks/00_quick_start/dkn.ipynb) | Python / GPU | | 
+| [Neural Collaborative Filtering (NCF)](notebooks/00_quick_start/ncf_movielens.ipynb) | Python / GPU | | 
+| [Restricted Boltzmann Machines (RBM)](notebooks/00_quick_start/rbm_movielens.ipynb) | Python / GPU | | 
+| [FastAI](notebooks/00_quick_start/fastai_recommendation.ipynb)  | Python / GPU | | 
+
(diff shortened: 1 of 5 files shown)


notebooks/03_evaluate/comparison.ipynb
```diff
@@ -373,16 +466,16 @@
...
-       "      <td>0.049102</td>\n",
-       "      <td>0.050849</td>\n",
-       "      <td>0.018680</td>\n",
-       "      <td>0.948764</td>\n",
-       "      <td>0.742875</td>\n",
-       "      <td>0.283734</td>\n",
-       "      <td>0.288698</td>\n",
-       "      <td>3.484409</td>\n",
-       "      <td>10.361868</td>\n",
+       "      <td>0.005232</td>\n",
```
nikhilrj: @gramhagen would it make sense to include results from a VW algo here?

README.md
```diff
@@ -52,15 +52,39 @@ We provide several notebooks to show how recommendation algorithms can be design
 
 - The [Operationalizion Notebook](notebooks/05_operationalize) demonstrates how to deploy models in production systems.
 
+
+The Quick-Start and Modeling notebooks showcase how to utilize the following algorithms to build a recommender system:
+
+**Algorithms**
+
+| Algorithm | Environment | Description |
```
nikhilrj: Ideally in the description we should put 1 sentence overviews of the strengths and weaknesses of the algorithm, and where the algorithm should be used. This description should NOT just describe the algorithm and dataset we're using it on, but also describe when and where a user should use this algorithm (where does it shine?),

README.md
```diff
@@ -52,15 +52,39 @@ We provide several notebooks to show how recommendation algorithms can be design
...
+
+**Algorithms**
+
+| Algorithm | Environment | Description | 
+| --- | --- | --- |
+| **`Classic Recommenders`** |
+| [Surprise/Singular Value Decomposition (SVD)](notebooks/00_quick_start/sar_single_node_movielens.ipynb) | Python |  | 
+| [Alternating Least Squares (ALS)](notebooks/00_quick_start/als_pyspark_movielens.ipynb) | Spark |  |
+| **`Microsoft Recommenders`** |
+| [Smart Adaptive Recommendations (SAR)](notebooks/00_quick_start/sar_single_node_movielens.ipynb) | Python / Spark | |
```
miguelgfierro: In the setup, we were referring to the environments as:
* Python CPU
* Python GPU
* PySpark
we should have consistency here

README.md
miguelgfierro: I think that makes sense

README.md
```diff
@@ -52,15 +52,39 @@ We provide several notebooks to show how recommendation algorithms can be design
...
+| [Vowpal Wabbit Family (VW)](notebooks/02_model/vowpal_wabbit_deep_dive.ipynb) | Python / Online | |
+| **`Deep Learning`** |
+| [eXtreme Deep Factorization Machine (xDeepFM)](notebooks/00_quick_start/xdeepfm.ipynb) | Python / GPU | | 
+| [Deep Knowledge-Aware Network (DKN)](notebooks/00_quick_start/dkn.ipynb) | Python / GPU | | 
+| [Neural Collaborative Filtering (NCF)](notebooks/00_quick_start/ncf_movielens.ipynb) | Python / GPU | | 
+| [Restricted Boltzmann Machines (RBM)](notebooks/00_quick_start/rbm_movielens.ipynb) | Python / GPU | | 
+| [FastAI](notebooks/00_quick_start/fastai_recommendation.ipynb)  | Python / GPU | |
```
miguelgfierro: I think we should rename FastAI to FastAI EmbeddingDotBias https://docs.fast.ai/collab.html#EmbeddingDotBias, which is the algo that we are using. I'm going to make a modification of the fastai notebook to make this more clear

notebooks/00_quick_start/fastai_recommendation.ipynb
```diff
@@ -27,12 +27,12 @@
      "name": "stdout",
```
miguelgfierro: do we need to run all these notebooks again? In the fastai one, we are not using GPU

notebooks/03_evaluate/comparison.ipynb
miguelgfierro: +1 to that

README.md
```diff
@@ -52,15 +52,39 @@ We provide several notebooks to show how recommendation algorithms can be design
...
+The Quick-Start and Modeling notebooks showcase how to utilize the following algorithms to build a recommender system:
+
+**Algorithms**
+
+| Algorithm | Environment | Description | 
+| --- | --- | --- |
+| **`Classic Recommenders`** |
+| [Surprise/Singular Value Decomposition (SVD)](notebooks/00_quick_start/sar_single_node_movielens.ipynb) | Python |  | 
+| [Alternating Least Squares (ALS)](notebooks/00_quick_start/als_pyspark_movielens.ipynb) | Spark |  |
+| **`Microsoft Recommenders`** |
```
miguelgfierro: One problem I see with the classification of classic, Microsoft and DL recommenders is that xDeepFM and DKN are also MS recommenders, but they are in DL...

README.md
```diff
@@ -52,15 +52,39 @@ We provide several notebooks to show how recommendation algorithms can be design
...
 
 **Preliminary Comparison**
 
 | Algo | MAP | nDCG@k | Precision@k | Recall@k | RMSE | MAE | R<sup>2</sup> | Explained Variance | 
 | --- | --- | --- | --- | --- | --- | --- | --- | --- | 
-| ALS | 0.002020 | 0.024313 | 0.030677 | 0.009649 | 0.860502 | 0.680608 | 0.406014 | 0.411603 | 
-| SAR | 0.064013 | 0.308012 | 0.277215 | 0.109292 | N/A | N/A | N/A | N/A | 
-| SVD | 0.010915 | 0.102398 | 0.092996 | 0.025362 | 0.888991 | 0.696781 | 0.364178 | 0.364178 | 
+| [ALS](notebooks/00_quick_start/als_pyspark_movielens.ipynb) | 0.002020 | 0.024313 | 0.030677 | 0.009649 | 0.860502 | 0.680608 | 0.406014 | 0.411603 |
```
miguelgfierro: guys, does it make sense to add here the training and testing time? There should be an important difference between VW and RBM for example

gramhagen: I think having train time would be good, test time might not be as useful if different metrics are being applied, or if the majority of the time is really just due to slow evaluation code, but if we can get to a equitable way to demonstrate evaluation time that would be good too

miguelgfierro: ok guys, I'm going to work on getting the times of all algos and will do a commit here

miguelgfierro: FYI https://github.com/Microsoft/Recommenders/issues/464#issuecomment-461037795

README.md
nikhilrj: Moved xDeepFM and DKN to MS (still open to how to best show this)

notebooks/03_evaluate/comparison.ipynb
gramhagen: is the current plan to update the comparison notebook so we have apples to apples for all algos?
if you need current results I can provide them

README.md
WessZumino: @gramhagen I have interpreted test time as the time needed to return recommendation on the test set rather then the time needed to evaluate the metrics. In the former case, test time would dependent only on the number of returned recommendation, i.e. 10 elements against all of them I think. The former definition is as you said, it depends on the evaluation algo. Now I am not sure which definition we are using :)

WessZumino: @miguelgfierro just a small note: I noticed that gpu algos needs more time to run the first time they are instantiated. In the case of RBM, the first time you run it takes considerably longer to run than the next n+1 times. I think this is due to TF initializing the GPU.

Concerning the reported metrics, I noticed  that they are lower then the one reported in the notebooks. Is there a reason for that?

Issue #483, opened 2019-02-04
Proper way to reference Azure ML Service


bethz:
Confirm the proper way to reference Azure ML Service, is it:
- Azure ML
- AML

This will affect docs, scripts, notebooks, etc.

bethz:
Update: there have been a number of emails on this topic. AML is definitely not recommended for use.  Currently finalizing what we can use.

miguelgfierro:
hey @bethz any news on this?

bethz:
A few things:
- AzureML should NOT be called AML. Please do NOT use AML in anything public facing documentation.
- It is suggested to introduce AzureML as an abbreviation in any document where you use it, e.g.
                      “This notebook shows how to use Azure Machine Learning services (AzureML) to …..”

Then you can use AzureML to refer to “Azure Machine Learning services” whenever you want.  This explicitly differentiates it from Azure Machine Learning Studio, for instance, which also has experiments, etc.

PR #493, opened 2019-02-07
Staging to Master: new set of algos, hyperparameters tuning and more

miguelgfierro:
### Checklist:

- [x] I have added tests.


WessZumino:
Hi, I noticed that the notebook's readme file, both for quick start and deep dive, does not include RBM description.

miguelgfierro:
great catch @WessZumino, do you want to add the description? if you do a PR to staging it will directly enter on this PR

nikhilrj:
Replace "AML" with "Azure Machine Learning" as per #483

reco_utils/dataset/numpy_splitters.py
```diff
@@ -0,0 +1,92 @@
...
+    Nusers, Nitems = X.shape  # total number of users and items
+
+    for u in range(Nusers):
+        # For each user obtain the index of rated movies
+        idx = np.asarray(np.where(Xtr[u] != 0))[0].tolist()
+
+        # extract a random subset of size n from the set of rated movies without repetition
+        idx_tst = np.random.choice(idx, tst[u], replace=False)
+        idx_train = list(set(idx).difference(set(idx_tst)))
+
```
loomlike: If the purpose is just to pick the rest of the index, we can simply use delete or mask using `idx_tst` instead of call two sets difference here.

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,204 @@
...
+from scipy.sparse import coo_matrix
+import logging
+
+# import default parameters
+from reco_utils.common.constants import (
+    DEFAULT_USER_COL,
+    DEFAULT_ITEM_COL,
+    DEFAULT_RATING_COL,
+    DEFAULT_TIMESTAMP_COL,
+    PREDICTION_COL,
```
loomlike: This is not about this file but, `reco_utils.common.constants`'s `PREDICTION_COL` should be `DEFAULT_...` to align with what we have

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,204 @@
...
+)
+
+# for logging
+log = logging.getLogger(__name__)
+
+
+class AffinityMatrix:
+    """
+
```
loomlike: Good to add what this class is about here, and move the Args description under `__init__`

reco_utils/dataset/sparse.py
```diff
@@ -0,0 +1,204 @@
...
+        self.col_item = col_item
+        self.col_user = col_user
+        self.col_rating = col_rating
+        self.col_pred = col_pred
+
+        # Options to save the model for future use
+        self.save_path = save_path
+
+    def gen_index(self):
+
```
loomlike: If this is a private method, change name to be `_gen_index(self)`

reco_utils/dataset/sparse.py
loomlike: Also seems like this class is not being used from anywhere. Make sure if this is deprecated module or not.

miguelgfierro:
hey @loomlike, regarding your comments, can you directly change the files and add a PR, which can be reviewed by @WessZumino and others? I have done some changes in some of these files, please see https://github.com/Microsoft/Recommenders/pull/504

loomlike:
Forgot to create a new branch and directly pushed to staging... so I just reverted the last commit I made. Will create a new PR for those changes.

gramhagen:
@miguelgfierro I was about to merge this but one of the tests failed. I'm not sure if it's just a fluke, but it's too late for me to dig into it. If you can take a look when you have time and merge to master I'd appreciate it. I'll bump the PR just to see if the tests run through ok, maybe too many things pushed all at once?

PR #497, opened 2019-02-07
Spark MLLib based evaluation

dciborow:
### Description

The following example provides a path for Spark MLLib users to run ranking metrics on ALS.
### Motivation and Context

This path is not provided in the Spark MLLib documentation. Instead that documentation only shows how to use RMSE. 
http://spark.apache.org/docs/latest/ml-collaborative-filtering.html
### Checklist:

- [ ] I have added tests.


notebooks/03_evaluate/evaluation_spark.ipynb
```diff
@@ -0,0 +1,136 @@
...
+        "cell_type": "markdown",
+        "source": ["### 2.2 Ranking metrics"],
+        "metadata": {}
+    }, {
+        "cell_type": "markdown",
+        "source": ["The following code fits the model, and creates the properly formated output dataset for ranking evaluation."],
+        "metadata": {}
+    }, {
+        "cell_type": "code",
+        "source": ["output = RankingAdapter(k=5, recommender=als).fit(ratings).transform(ratings)\n\noutput.show()"],
```
miguelgfierro: I'm curious how these metrics are created and how they compare in terms of speed to the metrics we created. If they are similar, we should think of a way to make them common. 
FYI @yueguoguo

notebooks/03_evaluate/evaluation_spark.ipynb
```diff
@@ -0,0 +1,136 @@
+{
+    "cells": [{
+        "cell_type": "markdown",
+        "source": ["# Spark Evaluation"],
+        "metadata": {}
+    }, {
+        "cell_type": "markdown",
+        "source": ["This notebook uses MMLSpark, please see github for installation details for your platform.\nhttps://github.com/Azure/mmlspark"],
```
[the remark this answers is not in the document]
dciborow: The code included in the first cell is "supposed" to install it into a local env for you, but this would not work on Databricks. I'm also very curious to see how the build server handles it...

`SUBMIT_ARGS = "--packages Azure:mmlspark:0.15 pyspark-shell"
os.environ["PYSPARK_SUBMIT_ARGS"] = SUBMIT_ARGS

from pyspark.sql import SparkSession

spark = (
   SparkSession.builder.appName("Spark Evaluation Metrics")
   .master("local[*]")
   .config("memory", "1G")
   .getOrCreate()
)`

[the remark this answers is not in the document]
dciborow: I was also thinking... the helper method we have... `get_or_create_spark()`, within it goes and starts Spark with the .getOrCreate() method call. Because this gets called within the helper, I can't further customize it... 
I am thinking it would make sense to refactor the method, and pull .getOrCreate() out into the actual notebooks...

dciborow:
@miguel-ferreira before I check, "i have updated the documentation", are there are particular places you think I need to update?

notebooks/03_evaluate/evaluation_spark.ipynb
yueguoguo: @dciborow do these ranking metrics adapter rely on Scala implementation of mmlspark? We may want to discuss more about it. If they are functionally the same with the ones existing in the repo, we should demonstrate the necessity of having them.

miguelgfierro:
This is mostly the documentation of the code, for example if we have a new algo, we have to add docstrings. Also, sometimes we need to add information in the readme, but it depends on the case

notebooks/03_evaluate/evaluation_spark.ipynb
dciborow: yes, all of the math is done in scala side of mmlspark. 
The goal of the notebook is to show Spark Devs who start off with the SparkMLLib guide how to use the advanced ranking metrics.

dciborow:
@miguel-ferreira how do you run the tests to determine if the mmlspark install works as expected?

miguelgfierro:
For the unit tests, it should be something like this: https://github.com/Microsoft/Recommenders/blob/master/tests/unit/test_notebooks_pyspark.py
Ideally, this should also install mmlspark (I think with the env instruction that you added  it should work right?)

nikhilrj:
@gramhagen can you follow up with Dan how to proceed with this PR? We should incorporate metrics we think are useful into the repo itself if that makes sense.

gramhagen:
Long term it would make sense to take a dependency on MML Spark for spark based evaluation. Updates to metrics can be done there and may have wider impact, plus we wouldn't have to port / maintain duplicate functionality. However in the short term the story for integrating MML Spark into this repo is not super straight forward as it requires different approaches locally vs Databricks. 

My thoughts on path forward here:
1. Let's focus on how to consolidate / simplify the installation of MML Spark whether through pip, or setup script, or SETUP.md instructions. 
2. In the mean time if there are specific metric additions needed we should add them to the existing spark and python evaluators.

Closing this PR until we have item 1 addressed then we can revisit replacing our spark based metrics with MML Spark.

PR #499, opened 2019-02-08
Gramhagen/vw notebook test

gramhagen:
### Description
Adding tests for vw notebook

### Motivation and Context
Need tests for all notebooks

### Checklist:

- [X] I have added tests.


tests/integration/test_notebooks_python.py
```diff
@@ -82,3 +82,30 @@ def test_surprise_svd_integration(notebooks, size, expected_values):
     for key, value in expected_values.items():
         assert results[key] == pytest.approx(value, rel=TOL)
 
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "size, expected_values",
+    [
+        ("1m", dict(rmse=0.9555,
```
miguelgfierro: is this guy fast enough to add 10M?

tests/unit/test_notebooks_python.py
```diff
@@ -38,3 +38,9 @@ def test_baseline_deep_dive_runs(notebooks):
 def test_surprise_deep_dive_runs(notebooks):
     notebook_path = notebooks["surprise_svd_deep_dive"]
     pm.execute_notebook(notebook_path, OUTPUT_NOTEBOOK, kernel_name=KERNEL_NAME)
+
+
+@pytest.mark.notebooks
```
miguelgfierro: so I guess vw is working because it is preinstalled on the dsvm, right?

We should think of adding a installer. Not sure if this PR is the best moment though

gramhagen: yeah, i'm close to resolving the issue of installing the python bindings on conda so that would be a preferable way to do this, they way it is being done on the dsvm is just through apt-get install vowpalwabbit

tests/integration/test_notebooks_python.py
gramhagen: the recommendation side is super slow, because of generating all pairs of user item interactions and writing them to disk, there is a utility within vw that will generate recommendations but that isn't pre-installed.

PR #516, opened 2019-02-13
README edits

yueguoguo:
### Description
* Uses "CPU" or "CPU-only" instead of "Python" may be more accurate in describing the environment.
* Updated descriptions of algorithms. 
* Removed repeated text about notebooks.
* Changed order of paragraph. 

### Checklist:

- [ ] I have added tests.


yueguoguo:
Categorization of "Classic Recommenders", "Microsoft Recommenders", and "Deep Learning Recommenders" may not make much sense. For example, there are deep learning recommenders in Microsoft recommenders as well. I removed them in the PR.

README.md
```diff
@@ -61,18 +61,15 @@ The table below lists recommender algorithms available in the repository at the
...
-| [FastAI Embedding Dot Bias (FAST)](notebooks/00_quick_start/fastai_movielens.ipynb)  | Python / GPU | Collaborative Filtering | General purpose algorithm embedding dot biases for users and items  |
+| [Surprise/Singular Value Decomposition (SVD)](notebooks/00_quick_start/sar_movielens.ipynb) | CPU | Collaborative Filtering | Matrix factorization algorithm for predicting explicit rating feedback in small datasets | 
+| [Alternating Least Squares (ALS)](notebooks/00_quick_start/als_movielens.ipynb) | Spark | Collaborative Filtering | Matrix factorization algorithm for explicit or implicit feedback in large datasets, optimized with Spark for scalability and distributed computing capability |
+| [Smart Adaptive Recommendations (SAR)](notebooks/00_quick_start/sar_movielens.ipynb) | CPU / Spark | Collaborative Filtering | Similarity-based algorithm for implicit feedback dataset, supporting both single-node and distributed environment (Spark) |
```
miguelgfierro: We still don't have SAR on Spark

README.md
```diff
@@ -61,18 +61,15 @@ The table below lists recommender algorithms available in the repository at the
...
+| [Vowpal Wabbit Family (VW)](notebooks/02_model/vowpal_wabbit_deep_dive.ipynb) | Online | Collaborative, Content-based Filtering | Fast online learning algorithms, great for scenarios where user features / context are constantly changing, like real-time bidding |
```
miguelgfierro: It is a little bit confusing for me the word online. I would say that online is not an environment, but a deployment method. The environment would be cpu

README.md
```diff
@@ -61,18 +61,15 @@ The table below lists recommender algorithms available in the repository at the
...
+| [eXtreme Deep Factorization Machine (xDeepFM)](notebooks/00_quick_start/xdeepfm_synthetic.ipynb) | CPU / GPU | Hybrid | Deep learning based algorithm for implicit and explicit feedback with user/item features | 
+| [Deep Knowledge-Aware Network (DKN)](notebooks/00_quick_start/dkn_synthetic.ipynb) | CPU / GPU | Content-based Filtering | Deep learning algorithm incorporating a knowledge graph and article embeddings to provide powerful news or article recommendations |
```
miguelgfierro: An open question here around the use of CPU/GPU in the DL algos. All the GPU algos can be run on CPU, but with certain datasets, this would be infeasible (theoretically the Spark algos can be also run on the CPU of the head node). Would it be ok to say only GPU? Maybe also another question, do we want to recommend our users to use these algos on CPU and GPU or only GPU?

nikhilrj: I think it's reasonable to put CPU/GPU here - someone might do a small / toy dataset on a CPU to get the e2e working then use GPUs when they need to train on a larger dataset.

README.md
yueguoguo: Ah yea good catch. Also here I removed the description of SAR for cold start scenarios.

README.md
yueguoguo: Make sense

README.md
yueguoguo: How about changing CPU to CPU-only

miguelgfierro: sounds good

yueguoguo:
@nikhilrj feel free to commit your edits directly into the PR.

loomlike:
@yueguoguo If we want to denote Microsoft Recommenders, what about add as a footnote. E.g.

| Algo | Env | Type | Desc
--- | --- | --- | ---
SAR<sup>*</sup>  | ... | ... | ...
...  | ... | ... | ...

> \* Invented (or developed or introduced or ...) by Microsoft

PR #520, opened 2019-02-13
Added note to readme and ALS notebook when a pyspark environment is necessary

heatherbshapiro:
### Description
Added notes to the getting started section of the Readme to explicitly call out that some of the quickstart notebooks require a different setup process for pyspark. Also added a note to the quickstart notebook itself to point out that a pyspark environment is required.

The ALS notebook is the first notebook in `00_quick_start` and users might not realize why it doesn't work with the basic setup from the readme. It is good to explicitly call this information out so users do not fall down the rabbit hole and get stuck.

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/notebooks/00_quick_start/als_movielens.ipynb
+++ b/notebooks/00_quick_start/als_movielens.ipynb
@@ -24,7 +24,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "**Note**: This notebook requires a PySpark environment to run properly.. Please follow the steps in [Setup.md](https://github.com/heatherbshapiro/Recommenders/blob/heather_update_readme_for_pyspark/SETUP.md#dependencies-setup) to install the PySpark environment."
+    "**Note**: This notebook requires a PySpark environment to run properly. Please follow the steps in [Setup.md](https://github.com/heatherbshapiro/Recommenders/blob/heather_update_readme_for_pyspark/SETUP.md#dependencies-setup) to install the PySpark environment."
    ]
   },
   {
--- a/notebooks/00_quick_start/als_movielens.ipynb
+++ b/notebooks/00_quick_start/als_movielens.ipynb
@@ -13,11 +13,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Running ALS on MovieLens (pySpark)\n",
+    "# Running ALS on MovieLens (PySpark)\n",
     "\n",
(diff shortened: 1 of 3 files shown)


notebooks/00_quick_start/als_movielens.ipynb
```diff
@@ -13,11 +13,18 @@
...
     "\n",
-    "This notebook provides an example of how to utilize and evaluate ALS pySpark ML (DataFrame-based API) implementation, meant for large-scale distributed datasets. We use a smaller dataset in this example to run ALS efficiently on multiple cores of a [Data Science Virtual Machine](https://azure.microsoft.com/en-gb/services/virtual-machines/data-science-virtual-machines/)."
+    "This notebook provides an example of how to utilize and evaluate ALS PySpark ML (DataFrame-based API) implementation, meant for large-scale distributed datasets. We use a smaller dataset in this example to run ALS efficiently on multiple cores of a [Data Science Virtual Machine](https://azure.microsoft.com/en-gb/services/virtual-machines/data-science-virtual-machines/)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Note**: This notebook requires a PySpark environment to run properly. Please follow the steps in [Setup.md](https://github.com/heatherbshapiro/Recommenders/blob/heather_update_readme_for_pyspark/SETUP.md#dependencies-setup) to install the PySpark environment."
```
miguelgfierro: Minor comment, we usually refer to the setup as SETUP.md

notebooks/00_quick_start/als_movielens.ipynb
```diff
@@ -13,11 +13,18 @@
    "cell_type": "markdown",
```
miguelgfierro: I would suggest to add the same comment to ALS deep dive notebook https://github.com/Microsoft/Recommenders/blob/staging/notebooks/02_model/als_deep_dive.ipynb

heatherbshapiro: Added

yueguoguo:
This is great. We may want to add this "notes" or "tags" to each notebook where PySpark is used (e.g., evaluation, splitting, and later some other algorithms need it). 
Do you think it is better to add it in the name of the notebook or elsewhere so that it will be immediately visible to users?

miguelgfierro:
@yueguoguo I don't have a strong opinion on this. Maybe next time a person reviews the notebooks we could ask if the environment is clear

PR #532, opened 2019-02-18
Staging to master, preparing for Release 2019.02

miguelgfierro:
### Description

PR for release https://github.com/Microsoft/Recommenders/projects/4

### Checklist:

- [x] I have added tests.


reco_utils/dataset/split_utils.py
```diff
@@ -159,4 +159,8 @@ def split_pandas_data_with_ratios(data, ratios, seed=1234, resample=False):
 
     splits = np.split(data, [round(x * len(data)) for x in split_index])
 
+    # Add split index (this makes splitting by group more efficient).
```
anargyri: The returned result is a list of numpy arrays, isn't it? The docstring says dataframes.
Also it would be good to mention in the docstring the 'split_index' attribute.

miguelgfierro: I checked and it returns dataframes. 

@yueguoguo looking at this function, I see that there is no control of the input ratios. For example, `ddd = split_pandas_data_with_ratios(df, [0.6, 0.2, 0.5])` will work, it will split 60%, 20% and 20%. The function is created so the last figure is always ignored. 

I'll send a PR to fix this

miguelgfierro: https://github.com/Microsoft/Recommenders/pull/533

anargyri: This is interesting because the numpy docs say that it should return a list of ndarrays 
https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.split.html
But as you said, it returns a list of dataframes if data is a dataframe and a list of ndarrays if data is an ndarray.

About the ratios, the functions in `python_splitters.py` expect them to be renormalized if they don't add up to one.

README.md
```diff
@@ -1,91 +1,74 @@
...
-| [FastAI Embedding Dot Bias (FAST)](notebooks/00_quick_start/fastai_movielens.ipynb)  | Python / GPU | Collaborative Filtering | General purpose algorithm embedding dot biases for users and items  |
-
-In addition, we also provide a [comparison notebook](notebooks/03_evaluate/comparison.ipynb) to illustrate how different algorithms could be evaluated and compared. In this notebook, data (MovieLens 1M) is randomly split into train/test sets at a 75/25 ratio. A recommendation model is trained using each of the collaborative filtering algorithms below. We utilize empirical parameter values reported in literature [here](http://mymedialite.net/examples/datasets.html). For ranking metrics we use k = 10 (top 10 results). We run the comparison on a Standard NC6s_v2 [Azure DSVM](https://azure.microsoft.com/en-us/services/virtual-machines/data-science-virtual-machines/) (6 vCPUs, 112 GB memory and 1 K80 GPU). Spark ALS is run in local standalone mode. 
+| [Smart Adaptive Recommendations (SAR)<sup>*</sup>](notebooks/00_quick_start/sar_movielens.ipynb) | Python CPU | Collaborative Filtering | Similarity-based algorithm for implicit feedback dataset |
```
nikhilrj: Do we have a reason why we chose this order or algorithms?

README.md
```diff
@@ -1,91 +1,74 @@
-# Recommenders 
+# Recommenders
 
 This repository provides examples and best practices for building recommendation systems, provided as Jupyter notebooks. The examples detail our learnings on five key tasks: 
 - [Prepare Data](notebooks/01_prepare_data/README.md): Preparing and loading data for each recommender algorithm
-- [Model](notebooks/02_model/README.md): Building models using various recommender algorithms such as Alternating Least Squares ([ALS](https://spark.apache.org/docs/latest/api/python/_modules/pyspark/ml/recommendation.html#ALS)), Singular Value Decomposition ([SVD](https://surprise.readthedocs.io/en/stable/matrix_factorization.html#surprise.prediction_algorithms.matrix_factorization.SVD)), etc.
+- [Model](notebooks/02_model/README.md): Building models using various classical and deep learning recommender algorithms such as Alternating Least Squares ([ALS](https://spark.apache.org/docs/latest/api/python/_modules/pyspark/ml/recommendation.html#ALS)) or eXtreme Deep Factorization Machines ([xDeepFM](https://arxiv.org/abs/1803.05170)).
```
nikhilrj: Note I made a small change here

README.md
```diff
@@ -1,91 +1,74 @@
...
+| [FastAI Embedding Dot Bias (FAST)](notebooks/00_quick_start/fastai_movielens.ipynb)  |  Python CPU / Python GPU | Collaborative Filtering | General purpose algorithm with embeddings and biases for users and items |
+
+**NOTE** - <sup>*</sup> indicates algorithms invented/contributed to by Microsoft.
 
 **Preliminary Comparison**
 
+We provide a [comparison notebook](notebooks/03_evaluate/comparison.ipynb) to illustrate how different algorithms could be evaluated and compared. In this notebook, data (MovieLens 1M) is randomly split into training/test sets at a 75/25 ratio. A recommendation model is trained using each of the collaborative filtering algorithms below. We utilize empirical parameter values reported in literature [here](http://mymedialite.net/examples/datasets.html). For ranking metrics we use k = 10 (top 10 results). We run the comparison on a Standard NC6s_v2 [Azure DSVM](https://azure.microsoft.com/en-us/services/virtual-machines/data-science-virtual-machines/) (6 vCPUs, 112 GB memory and 1 P100 GPU). Spark ALS is run in local standalone mode.
```
loomlike: "For ranking metrics we use k = 10 (top 10 results)"
more accurate statement would be "(top 10 recommended items)"

reco_utils/dataset/python_splitters.py
```diff
@@ -98,17 +103,26 @@ def python_chrono_split(
...
 
-    num_of_splits = len(ratio)
-    splits = [pd.DataFrame({})] * num_of_splits
+    # Split by each group and aggregate splits together.
+    splits = []
     df_grouped = data.sort_values(col_timestamp).groupby(split_by_column)
     for name, group in df_grouped:
         group_splits = split_pandas_data_with_ratios(
             df_grouped.get_group(name), ratio, resample=False
         )
```
loomlike: "resample=False" this part is confusion. Is it "re-sampling" or "shuffling"?

This is the codes in split_pandas_data_with_ratios fn:
```
if resample:
        data = data.sample(frac=1, random_state=seed)
```
frac=1 so the intention is to shuffle, not re-sample. Also do we want to reset_index there or keep the original index?

reco_utils/dataset/python_splitters.py
```diff
@@ -162,17 +183,26 @@ def python_stratified_split(
...
+        concat_group_splits = pd.concat(group_splits)
+
+        splits.append(concat_group_splits)
+
+    # Concatenate splits for all the groups together.
+    splits_all = pd.concat(splits)
+
+    # Take split by split_index
+    splits_list = [splits_all[splits_all['split_index'] == x].drop('split_index', axis=1) for x in range(len(ratio))]
+
```
loomlike: Our stratified and chrono splitters share many similar codes. Does it make sense to define a private function, e.g. `_prepare_stratification()` and share them from both splitters? Maybe a bit tricky in the main splitting algorithm part but at least the column validity check and data preparation part look the same.

tests/smoke/test_deeprec_model.py
```diff
@@ -21,13 +24,17 @@ def resource_path():
...
+    data_file = os.path.join(data_path, "sample_FFM_data.txt")
+    output_file = os.path.join(data_path, "output.txt")
 
     if not os.path.exists(yaml_file):
-        download_deeprec_resources(r'https://recodatasets.blob.core.windows.net/deeprec/', data_path, 'xdeepfmresources.zip')
+        download_deeprec_resources(
+            "https://recodatasets.blob.core.windows.net/deeprec/",
+            data_path,
+            "xdeepfmresources.zip",
+        )
```
loomlike: We don't have docstrings for this function. Also, the functions under *reco_utils/recommender/deeprec* either don't have docstring nor don't follow our format.

reco_utils/dataset/python_splitters.py
anargyri: Good point, they look very similar. The main difference between the two methods is `sort_values(col_timestamp)`
Probably the chrono split could be reduced to sorting the timestamps and then calling the stratified splitter. We could open a new issue for this.

README.md
miguelgfierro: oh, it is true, it is a little bit messy. I'll reorder a little. PR here: https://github.com/Microsoft/Recommenders/pull/536

README.md
miguelgfierro: updated

reco_utils/dataset/python_splitters.py
miguelgfierro: I created an issue and added you https://github.com/Microsoft/Recommenders/issues/537

tests/smoke/test_deeprec_model.py
miguelgfierro: Yeah you are right, there is a lot to refactor in deeprec. I'll start working on that. on this branch: `miguel/fixes_prerelease`

As for the tests, there is no need to add docstrings there. Some examples [1](https://github.com/Microsoft/LightGBM/blob/master/tests/python_package_test/test_consistency.py), [2](https://github.com/Microsoft/LightGBM/blob/master/tests/python_package_test/test_basic.py), [3](https://github.com/pytorch/pytorch/blob/master/test/test_torch.py)

reco_utils/dataset/python_splitters.py
miguelgfierro: addressed in the next commit

PR #533, opened 2019-02-18
added check to pandas splitter

miguelgfierro:
### Description

Add checker to make sure that the ratios sum to 1

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/dataset/split_utils.py
+++ b/reco_utils/dataset/split_utils.py
@@ -152,6 +152,9 @@ def split_pandas_data_with_ratios(data, ratios, seed=1234, resample=False):
     Returns:
         list: List of pd.DataFrame splitted by the given specifications.
     """
+    if sum(ratios) != 1.0:
+        raise ValueError("The ratios have to sum to 1")
+        
     split_index = np.cumsum(ratios).tolist()[:-1]
 
     if resample:
--- a/reco_utils/dataset/split_utils.py
+++ b/reco_utils/dataset/split_utils.py
@@ -145,7 +145,7 @@ def split_pandas_data_with_ratios(data, ratios, seed=1234, resample=False):
 
     Args:
         data (pd.DataFrame): Pandas data frame to be split.
-        ratios (list of floats): list of ratios for split.
+        ratios (list of floats): list of ratios for split. The ratios have to sum to 1.
         seed (int): random seed.
         resample (bool): whether data will be resampled when being split.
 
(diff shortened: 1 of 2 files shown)


yueguoguo:
Note the ratio does not have to be summed to 1. If they are not, the codes will do the normalization. This is done by the function `process_split_ratio` in [here](https://github.com/Microsoft/Recommenders/blob/master/reco_utils/dataset/split_utils.py).

reco_utils/dataset/split_utils.py
```diff
@@ -152,6 +152,9 @@ def split_pandas_data_with_ratios(data, ratios, seed=1234, resample=False):
     Returns:
         list: List of pd.DataFrame splitted by the given specifications.
     """
+    if sum(ratios) != 1.0:
+        raise ValueError("The ratios have to sum to 1")
```
anargyri: Could you also add the requirement "The ratios have to sum to 1" in the docstring.

tests/unit/test_python_splitter.py
```diff
@@ -79,6 +79,30 @@ def random_date_generator(start_date, range_in_days):
     return rating
 
 
+@pytest.fixture(scope="module")
+def python_int_dataset(test_specs):
+    # fix the the random seed
```
anargyri: remove extra "the"

tests/unit/test_python_splitter.py
```diff
@@ -79,6 +79,30 @@ def random_date_generator(start_date, range_in_days):
...
+    return np.random.randint(
+        low=0,
+        high=6,
+        size=(test_specs["number_of_users"], test_specs["number_of_items"]),
+    )
+
+
+@pytest.fixture(scope="module")
+def python_float_dataset(test_specs):
+    # fix the the random seed
```
anargyri: here too

tests/unit/test_python_splitter.py
```diff
@@ -79,6 +79,30 @@ def random_date_generator(start_date, range_in_days):
...
+    np.random.seed(test_specs["seed"])
+
+    # generates the user/item affinity matrix. Ratings are from 1 to 5, with 0s denoting unrated items
```
anargyri: maybe better say that ratings are in the interval [0, 5)

reco_utils/dataset/split_utils.py
miguelgfierro: good suggestion

Issue #513, opened 2019-02-12
VW wrapper needed


gramhagen:
### Description
Current vw testing doesn't match other algorithms execution pattern so can't easily be run for benchmarking

### In which platform does it happen?
Notebooks for vw testing / benchmarking

### Expected behavior (i.e. solution)
should be able to load model with parameters then fit, then predict:
```
vw = VW(**params)
vw.fit(train_dataframe)
prediction_dataframe = vw.predict(test_dataframe)
```

loomlike:
#514 Check this issue about Recommender interface class and let's have discussion.

PR #534, opened 2019-02-19
Gramhagen/vw wrapper

gramhagen:
### Description
new wrapper for vowpal wabbit to support common interface for model instantiation, fit, and predict methods. this is a stop-gap solution until the vw python bindings (and sklearn interface) can be cleanly installed in a conda env for use in the notebooks here.

the unittests do not require vw to be installed, but for this class to work for real vw must be installed on the operating system it is running on, and on the path for the python process calling this (i.e. commands such as 'vw -f test.model -d train.dat' should work)

### Related Issues
#513 

### Checklist:

- [X] I have added tests.


reco_utils/recommender/vowpal_wabbit/vw.py
```diff
@@ -0,0 +1,242 @@
...
+            # add the correct hyphen to the parameter
+            cmd.append(f"-{k}" if len(k) == 1 else f"--{k}")
+            if v is not True:
+                # don't add an argument for parameters with value == True
+                cmd.append("{}".format(v))
+
+        return cmd
+
+    def parse_train_params(self, params):
+        """Parse input hyper-parameters to build vw train commands"""
```
yueguoguo: docstrings for input arguments

reco_utils/recommender/vowpal_wabbit/vw.py
```diff
@@ -0,0 +1,242 @@
...
+            {
+                "d": self.train_file,
+                "f": self.model_file,
+                "quiet": params.get("quiet", True),
+            }
+        )
+        return self.to_vw_cmd(params=train_params)
+
+    def parse_test_params(self, params):
+        """Parse input hyper-parameters to build vw test commands"""
```
yueguoguo: docstrings for input arguments

tests/unit/test_vowpal_wabbit.py
```diff
@@ -0,0 +1,125 @@
...
+from unittest import mock
+
+
+import pandas as pd
+
+from reco_utils.recommender.vowpal_wabbit.vw import VW
+
+
+@pytest.fixture(scope="module")
+def df():
```
yueguoguo: Do we want to use the existing fixture for dummy data generation?

tests/unit/test_vowpal_wabbit.py
```diff
@@ -0,0 +1,125 @@
...
+
+@pytest.fixture(scope="function")
+def model():
+    model = VW(col_user="user", col_item="item", col_prediction="prediction", q="ui")
+    yield model
+    del model
+
+
+def test_vw_init_del():
+    model = VW()
```
yueguoguo: This line can be saved by passing the fixture function?

reco_utils/recommender/vowpal_wabbit/vw.py
```diff
@@ -0,0 +1,242 @@
...
+        self,
+        col_user=DEFAULT_USER_COL,
+        col_item=DEFAULT_ITEM_COL,
+        col_rating=DEFAULT_RATING_COL,
+        col_timestamp=DEFAULT_TIMESTAMP_COL,
+        col_prediction=PREDICTION_COL,
+        **kwargs,
+    ):
+        """Initialize model parameters
+        Args:
```
miguelgfierro: There are some missing parameters

reco_utils/recommender/vowpal_wabbit/vw.py
```diff
@@ -0,0 +1,242 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+This file provides a wrapper to run Vowpal Wabbit from the command line through python.
+It is not recommended to use this approach in production, there are python bindings that can be installed from the
```
miguelgfierro: Why this is not recommended? it should work similarity to the command lines, doesn't it?
I remember you mentioned that the python bindings had some problems...

reco_utils/recommender/vowpal_wabbit/vw.py
```diff
@@ -0,0 +1,242 @@
...
+        self.col_item = col_item
+        self.col_rating = col_rating
+        self.col_timestamp = col_timestamp
+        self.col_prediction = col_prediction
+
+        self.logistic = "logistic" in kwargs.values()
+        self.train_cmd = self.parse_train_params(params=kwargs)
+        self.test_cmd = self.parse_test_params(params=kwargs)
+
+    @staticmethod
```
miguelgfierro: why static?

reco_utils/recommender/vowpal_wabbit/vw.py
gramhagen: gotit

reco_utils/recommender/vowpal_wabbit/vw.py
gramhagen: gotit

tests/unit/test_vowpal_wabbit.py
gramhagen: here I want to directly control creation and destruction instead of relying on the fixture.

tests/unit/test_vowpal_wabbit.py
gramhagen: we could but I still feel like this approach is a bandaid, i'd prefer to spend effort fixing the issue with vw python bindings themselves so we can remove this extra code.

reco_utils/recommender/vowpal_wabbit/vw.py
gramhagen: it does, but the extra step of having to write the dataframe to file is really inefficient. the desired approach would be to import the sklearn wrapper that comes with the python bindings. however the current code will not easily install under conda. I'm working on that, but it's not there yet.

once that is working i suggest we move over to using that code and pulling any useful dataframe formatting either into VW itself or into a utility function. Then we can reduce additional wrapper code here.

reco_utils/recommender/vowpal_wabbit/vw.py
gramhagen: just because it doesn't access any class members

reco_utils/recommender/vowpal_wabbit/vw.py
miguelgfierro: ok I see

gramhagen:
blarg, I don't know how those got there, I assumed it was from merging staging

gramhagen:
closing this in favor of clean pr

PR #535, opened 2019-02-19
Update issue templates

yueguoguo:


yueguoguo:
It makes things convenient - what do you guys think? 
**screen shot from azure-cli**
image

yueguoguo:
Benefits of this guy are
* We can pop up a page where issue creator can select from the pre-defined 3 types, i.e., bug, feature, general-ask. 
* The created issue by using the three types will be prefixed with [BUG], [FEATURE], and [GENERAL], respectively. 
These make it easy for us to template issues and track issues.

yueguoguo:
The new ones are for quick-starts of creating issues with the pre-defined types. If an issue does not fall into any of the three, the creator can go with a regular issue template, which is [this guy](1. https://github.com/Microsoft/Recommenders/blob/le/issue_template/.github/ISSUE_TEMPLATE.md).

Issue #539, opened 2019-02-19
Refactor fit method in deeprec base model


miguelgfierro:
### Description

Question to @Leavingseason, in this method:
```
def fit(self, train_file, valid_file, test_file=None):
        """Fit the model with train_file. Evaluate the model on valid_file per epoch to observe the training status.
        If test_file is not None, evaluate it too.
        
        Args:
            train_file (str): training data set.
            valid_file (str): validation set.
            test_file (str): test set.

        Returns:
            obj: An instance of self.
        """
        if self.hparams.write_tfevents:
            self.writer = tf.summary.FileWriter(
                self.hparams.SUMMARIES_DIR, self.sess.graph
            )

        train_sess = self.sess
        for epoch in range(1, self.hparams.epochs + 1):
            step = 0
            self.hparams.current_epoch = epoch

            epoch_loss = 0
            train_start = time.time()
            for batch_data_input in self.iterator.load_data_from_file(train_file):
                step_result = self.train(train_sess, batch_data_input)
                (_, step_loss, step_data_loss, summary) = step_result
                if self.hparams.write_tfevents:
                    self.writer.add_summary(summary, step)
                epoch_loss += step_loss
                step += 1
                if step % self.hparams.show_step == 0:
                    print(
                        "step {0:d} , total_loss: {1:.4f}, data_loss: {2:.4f}".format(
                            step, step_loss, step_data_loss
                        )
                    )

            train_end = time.time()
            train_time = train_end - train_start

            if self.hparams.save_model:
                if epoch % self.hparams.save_epoch == 0:
                    checkpoint_path = self.saver.save(
                        sess=train_sess,
                        save_path=self.hparams.MODEL_DIR + "epoch_" + str(epoch),
                    )

            eval_start = time.time()
            train_res = self.run_eval(train_file)
            eval_res = self.run_eval(valid_file)
            train_info = ", ".join(
                [
                    str(item[0]) + ":" + str(item[1])
                    for item in sorted(train_res.items(), key=lambda x: x[0])
                ]
            )
            eval_info = ", ".join(
                [
                    str(item[0]) + ":" + str(item[1])
                    for item in sorted(eval_res.items(), key=lambda x: x[0])
                ]
            )
            if test_file is not None:
                test_res = self.run_eval(test_file)
                test_info = ", ".join(
                    [
                        str(item[0]) + ":" + str(item[1])
                        for item in sorted(test_res.items(), key=lambda x: x[0])
                    ]
                )
[...]

PR #540, opened 2019-02-19
Miguel/fixes prerelease

miguelgfierro:
### Description

Fixes suggested by @loomlike  and documentation added to deeprec algos

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/reco_utils/recommender/deeprec/models/base_model.py
+++ b/reco_utils/recommender/deeprec/models/base_model.py
@@ -77,7 +77,9 @@ class BaseModel(object):
         Args:
             logit (obj): Base prediction value.
             task (str): A task (values: regression/classification)
-        :return: transformed score
+        
+        Returns:
+            obj: Transformed score
         """
         if task == "regression":
             pred = tf.identity(logit)
--- a/reco_utils/recommender/deeprec/models/base_model.py
+++ b/reco_utils/recommender/deeprec/models/base_model.py
@@ -1,9 +1,13 @@
-"""define base class model"""
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
 import abc
+import os
+import time
+import numpy as np
+import collections
 import tensorflow as tf
 from ..deeprec_utils import *
-import numpy as np
-import os, time, collections
 
 
 __all__ = ["BaseModel"]
@@ -46,7 +50,9 @@ class BaseModel(object):
 
         # set GPU use with demand growth
         gpu_options = tf.GPUOptions(allow_growth=True)
-        self.sess = tf.Session(graph=self.graph, config=tf.ConfigProto(gpu_options=gpu_options))
+        self.sess = tf.Session(
+            graph=self.graph, config=tf.ConfigProto(gpu_options=gpu_options)
+        )
         self.sess.run(self.init_op)
 
     def _get_loss(self):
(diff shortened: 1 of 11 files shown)


reco_utils/recommender/deeprec/models/base_model.py
```diff
@@ -46,24 +50,33 @@ def __init__(self, hparams, iterator_creator, graph=None):
...
-        """
-        Make final output as prediction score, according to different tasks.
-        :param logit: base prediction value
-        :param task:  regression or classification
+        """Make final output as prediction score, according to different tasks.
+        
+        Args:
+            logit (obj): Base prediction value.
+            task (str): A task (values: regression/classification)
         :return: transformed score
```
loomlike: Here, you missed one line of reformatting. `Returns:`

reco_utils/recommender/deeprec/models/base_model.py
```diff
@@ -385,16 +358,16 @@ def load_model(self, model_path=None):
             raise IOError("Failed to find any matching files for {0}".format(act_path))
 
     def fit(self, train_file, valid_file, test_file=None):
-        """
-        Fit the model with train_file. Evaluate the model on valid_file per epoch to observe the training status.
+        """Fit the model with train_file. Evaluate the model on valid_file per epoch to observe the training status.
         If test_file is not None, evaluate it too.
+
```
loomlike: This function accepting both validation and test data is strange to me... why does this function evaluate the model both on validation and test data?

Also, having validation data as an input to `fit` function doesn't sound like following single responsibility principle.
E.g. TF has a dedicated function for this ([train_and_evaluate](https://www.tensorflow.org/api_docs/python/tf/estimator/train_and_evaluate))

miguelgfierro: yeah, good point, I'm discussing this with Jianxun https://github.com/Microsoft/Recommenders/issues/539 He and I will work on this in the next release

reco_utils/recommender/deeprec/models/base_model.py
miguelgfierro: good catch

tests/unit/test_python_splitter.py
```diff
@@ -47,8 +47,7 @@ def python_dataset(test_specs):
     def random_date_generator(start_date, range_in_days):
         """Helper function to generate random timestamps.
```
loomlike: This file's `tech_specs["user_ids"]` is not used from anywhere.

tests/unit/test_python_splitter.py
```diff
@@ -94,28 +86,34 @@ def python_int_dataset(test_specs):
...
     assert len(splits[2]) == round(shape * 0.52)
 
     with pytest.raises(ValueError):
-        splits = split_pandas_data_with_ratios(pandas_dummy_timestamp, ratios=[0.6, 0.2, 0.4])
+        splits = split_pandas_data_with_ratios(
+            pandas_dummy_timestamp, ratios=[0.6, 0.2, 0.4]
+        )
 
 
 def test_min_rating_filter(python_dataset):
```
loomlike: This function doesn't test min_rating_filter thoroughly as it relies on random-generated data of 1000 records for 4 users, end up almost no chances to have the users whose `min_rating<5`. Should provide a dataset which actually includes those users.

miguelgfierro: @loomlike, do you have time today to take a look at this? if not, @yueguoguo, do you have time tomorrow during your day? if not, I'll take a look tomorrow in my morning :-)

loomlike: @miguelgfierro where is plan C? haha I can take a look at this today. Will push the changes to this branch.

miguelgfierro:
looks great!

PR #563, opened 2019-02-22
Addressed :bug: #550

### LINKED ISSUE #550: [BUG] conda update conda error - PackageNotInstalledError: Package is not installed in prefix
### Description
I'm getting an error when running the first setup instruction on my Linux (Ubuntu) DSVM.

### In which platform does it happen?
Azure Data Science Virtual Machine (Ubuntu)

I created a Linux (Ubuntu) DSVM, set up port forwarding, and opened a browser to log into to Jupyter using my VM username/password. I then ran the first step from the command line to setup the dependencies, from the instructions here: https://github.com/Microsoft/Recommenders/blob/master/SETUP.md#dependencies-setup:

conda update conda

This produced the following console output error:

PackageNotInstalledError: Package is not installed in prefix.
  prefix: /data/anaconda/envs/py35
  package name: conda

Should be no error

### Other Comments
I could run the next line, 'conda update anaconda', so it looks like conda is indeed there.

miguelgfierro:
### Description

Fixes #550

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -44,7 +44,7 @@ Currently, this repository supports the following environments:
 We install the dependencies with Conda. As a pre-requisite, we want to make sure that Anaconda and the package manager Conda are both up to date:
 
 ```{shell}
-conda update conda
+conda update conda -n root
 conda update anaconda
 ```
 


miguelgfierro:
hey @roalexan, can you please review?

The process we follow is that a person (in this case you) review the change. If the change is ok, then you approve. If there is something missing or that needs to be changed, you either commit to the same branch `miguelgfierro-patch-1` or you tell me, and I'll do it

PR #568, opened 2019-02-24
Prototype of Running Notebooks in AzureML

danielsc:
### Description
**This PR is just for discussion purposes, not meant to be merged in.**
The PR is showing a prototype of how a user can run a notebook without any changes on AzureML using papermill. The main action is happening [in this notebook](https://github.com/Microsoft/Recommenders/blob/danielsc/azureml/notebooks/02_train_in_the_cloud/train_fastai_on_aml.ipynb) and [this wrapper script](https://github.com/Microsoft/Recommenders/blob/danielsc/azureml/scripts/notebook_launcher.py). 
Things to note:
1. Due to [this bug](https://msdata.visualstudio.com/Vienna/_workitems/edit/377078), the compute target needs to be DSVM right now. After this has been fixed, the same should work on AML Compute
2. Due to [this bug](https://msdata.visualstudio.com/Vienna/_workitems/edit/377077), I had to change the order of the dependencies in the yaml file generation -- don't ask... it is already fixed, but will take time to be rolled out.
2. In this prototype, the parameter passing goes over the command line, which restricts us to strings. Setting TOP_K to 10 would fail. In the product implementation we should preserve the types.
3. The output parameters of the notebook are nicely written to run history and the output notebook is uploaded since it is written to ./outputs. However, it is only uploaded after the job completes. For the product implementation, we should think about uploading it while the job is still running and about visualizing it in the widget or workspace UX.
4. I am not happy with the amount of code that is required to set up a custom conda environment. We should look at making that easier/more compact in AML.

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/.amlignore
+++ b/.amlignore
@@ -1,0 +1,7 @@
+.ipynb_checkpoints
+azureml-logs
+.azureml
+.git
+outputs
+azureml-setup
+docs
--- a/scripts/notebook_launcher.py
+++ b/scripts/notebook_launcher.py
@@ -1,0 +1,35 @@
+import sys
+import os
+import papermill as pm
+import argparse
+from azureml.core import Run
+
+os.system(sys.executable + ' -m ipykernel install --user --name ipyazureml')
+
+def execute_notebook(notebook_path, parameters):
+    pm.execute_notebook(
+        notebook_path,
+        'outputs/output.ipynb',
+        kernel_name='ipyazureml',
+        parameters=parameters,
+    )
+    results = pm.read_notebook('outputs/output.ipynb').dataframe.set_index("name")["value"]
+
+    run = Run.get_context()
+    for key, value in results.items():
+        run.log(key, value)
+            
+    return results
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--notebook_path')
+    FLAGS, unparsed = parser.parse_known_args()
+
+    parameters= {unparsed[i]: unparsed[i+1] for i in range(0, len(unparsed), 2)}
+    print(FLAGS.notebook_path, parameters)
+    execute_notebook(notebook_path = FLAGS.notebook_path, 
+                       parameters=parameters)
+
+
--- a/notebooks/02_train_in_the_cloud/train_fastai_on_aml.ipynb
+++ b/notebooks/02_train_in_the_cloud/train_fastai_on_aml.ipynb
@@ -1,0 +1,285 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Run a notebook in Azure using AzureML\n",
+    "\n",
+    "This requires an AzureML workspace to be set up. The workspace will be loaded from the configuration using `Workspace.from_config()`. \n",
+    "\n",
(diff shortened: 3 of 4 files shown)


jreynolds01:
A couple of talking points to highlight that @loomlike and I have noticed when working on this as well:

- Due to bullet 4 above - features that depend on incremental logging (e.g. early stopping for hyperdrive) aren't fully leveraged with notebooks
- A similar approach to doing this for pipelines requires a good understanding of the `hash_paths` to parameter of `PythonScriptStep` and `regenerate_outputs` parameter of submit, because it is a really good example of when those things are important. I could not find any good examples of those parameters in use, and the documentation is also not easy to find. You can see an example [here ](https://github.com/jreynolds01/papermill_execution_azureml/blob/master/simple-pm-run-as-pipeline.ipynb) that walks through examples of that.

ateste:
This is very cool! Just for my information: If, for any reason, the execution of all this fails, will there be a way for the user to know which of the notebooks failed to complete? In other words, will the logs be explicit enough for the user to know which of the "master" or "training" notebook to fix?

scripts/notebook_launcher.py
```diff
@@ -0,0 +1,35 @@
...
+import argparse
+from azureml.core import Run
+
+os.system(sys.executable + ' -m ipykernel install --user --name ipyazureml')
+
+def execute_notebook(notebook_path, parameters):
+    pm.execute_notebook(
+        notebook_path,
+        'outputs/output.ipynb',
+        kernel_name='ipyazureml',
```
miguelgfierro: question here, is it mandatory to install the kernel pyazureml? will this run with the default kernel: https://github.com/Microsoft/Recommenders/blob/master/tests/notebooks_common.py#L8?

.amlignore
```diff
@@ -0,0 +1,7 @@
+.ipynb_checkpoints
```
miguelgfierro: is this file mandatory as well, can we not ignore these files with gitignore?

loomlike:
Adding to @jreynolds01 point regarding 4. When writing the same metrics via Papermill, it will overwrite. Example is recording error rate while training for every iterations (to track error rate while training).

One way go-around is to store all metrics into a temporal list (e.g. get metric every iteration and append the result to the list), at the end of the training, record the **list** to papermill, and from the outside script, get the list and record back to azureml log, which is not a very neat solution..

scripts/notebook_launcher.py
jreynolds01: based on tests @loomlike and I worked on - it will work on the 'python3' kernel as long as ipykernel is there to support the default kernel.

yueguoguo:
> * A similar approach to doing this for pipelines requires a good understanding of the `hash_paths` to parameter of `PythonScriptStep` and `regenerate_outputs` parameter of submit, because it is a really good example of when those things are important. I could not find any good examples of those parameters in use, and the documentation is also not easy to find. You can see an example [here ](https://github.com/jreynolds01/papermill_execution_azureml/blob/master/simple-pm-run-as-pipeline.ipynb) that walks through examples of that.

Thinking of the same thing. Alternatively can we wrap a utility that converts a notebook into a script, and go with a `PythonScriptStep`? In that case, we may bypass the limitation of `papermill` for recording the metrics. We can then write temporary results to `hash_paths` to detect changes.

nikhilrj:
Closing - FYI @jreynolds01 @bethz who are incorporating this into next steps

Issue #570, opened 2019-02-26
[FEATURE] update databricks admin scripts to leverage python to be more cross-platform


jreynolds01:
### Description
Similar to updating the conda generation script, this will move us towards cross-platform capability

### Expected behavior with the suggested feature
same behavior, just more platforms supported (e.g. #564 should be taken care of).

### Other Comments

Issue #567, opened 2019-02-22
[FEATURE] Support for Windows


miguelgfierro:
### Description

Test pipeline for a Windows machine, which will include unit, smoke and integration tests for CPU, GPU and PySpark envs.

### Expected behavior with the suggested feature

Similar to what we have in Linux, but with Windows

gramhagen:
after reinstalling visual studio and cuda toolkit on windows the gpu environment works for me, pyspark and base too, so this should be doable without too much hassle

loomlike:
Is visual studio necessary?

gramhagen:
pytorch needed the vs14 runtime libraries

gramhagen:
hmm, after looking more closely I don't think it's actually using the gpu device, tf.test.is_gpu_available() still indicates False =(

Issue #564, opened 2019-02-22
[FEATURE] SETUP for Azure Databricks uses 'zip' command not available on Windows


roalexan:
### Description
The setup instructions here, https://github.com/Microsoft/Recommenders/blob/master/SETUP.md#repository-installation, say to run 'zip' on your local computer. If you're running Windows, this won't work.

### Expected behavior with the suggested feature
For those users running Windows, describe the steps they should use to create the archived file. My workaround will probably be to use a Linux DSVM that I have already created (and downloaded the recommendations repo) and run the 'zip' command there.

yueguoguo:
Still working to support other OS, but yea this is something important. 

At least we need to make it clear somewhere that all the system commands used/mentioned in the repo are for Linux.

roalexan:
I agree that it might be sufficient to state somewhere that the commands are intended to be run on a Linux environment. Since these commands are designed to be run on the user's local computer, I do wonder how many people have a local Linux computer.

bethz:
I think we need both Windows and Linux support. In a couple of the prior companies I worked for, Linux was the most prevalent work environment with Windows used only for email/meeting setup/office. But, Windows dev is a huge environment that we have a unique advantage with and should not ignore. So, that means both.

loomlike:
You can use Windows with some extra tools, e.g. [GOW](https://github.com/bmatzelle/gow/wiki) (Gnu on Windows). I'm using Windows and all the scripts and setups in our repo are working on my env.

roalexan:
Maybe. I like the way some other Microsoft documentation handles setup instructions, example:

# Bash
zip -r myAppFiles.zip .

# PowerShell
Compress-Archive -Path * -DestinationPath myAppFiles.zip

from: https://docs.microsoft.com/en-us/azure/app-service/app-service-web-get-started-nodejs#create-a-project-zip-file

Issue #529, opened 2019-02-16
run notebook on Azure Databricks get error


tengfei86:
### Description
**als_movie_o16n.ipynb on databricks  I get error when execute codes below:**
`client = get_client_from_cli_profile(azure.mgmt.cosmosdb.CosmosDB,
                                    subscription_id=subscription_id)`

> Error Stack:
> CLIError                                  Traceback (most recent call last)
> <command-3006291621032396> in <module>()
> ----> 1 client = get_client_from_cli_profile(azure.mgmt.cosmosdb.CosmosDB)
>       2 
>       3 async_cosmosdb_create = client.database_accounts.create_or_update(
>       4     resource_group,
>       5     account_name,
> 
> /databricks/python/lib/python3.5/site-packages/azure/common/client_factory.py in get_client_from_cli_profile(client_class, **kwargs)
>      77         resource, _ = _client_resource(client_class, cloud)
>      78         credentials, subscription_id, tenant_id = get_azure_cli_credentials(resource=resource,
> ---> 79                                                                             with_tenant=True)
>      80         parameters.update({
>      81             'credentials': kwargs.get('credentials', credentials),
> 
> /databricks/python/lib/python3.5/site-packages/azure/common/credentials.py in get_azure_cli_credentials(resource, with_tenant)
>      46     """
>      47     profile = get_cli_profile()
> ---> 48     cred, subscription_id, tenant_id = profile.get_login_credentials(resource=resource)
>      49     if with_tenant:
>      50         return cred, subscription_id, tenant_id
> 
> /databricks/python/lib/python3.5/site-packages/azure/cli/core/_profile.py in get_login_credentials(self, resource, subscription_id, aux_subscriptions)
>     488 
>     489     def get_login_credentials(self, resource=None, subscription_id=None, aux_subscriptions=None):
> --> 490         account = self.get_subscription(subscription_id)
>     491         user_type = account[_USER_ENTITY][_USER_TYPE]
>     492         username_or_sp_id = account[_USER_ENTITY][_USER_NAME]
> 
> /databricks/python/lib/python3.5/site-packages/azure/cli/core/_profile.py in get_subscription(self, subscription)
>     450         subscriptions = self.load_cached_subscriptions()
>     451         if not subscriptions:
> --> 452             raise CLIError("Please run 'az login' to setup account.")
>     453 
>     454         result = [x for x in subscriptions if (
> 
> CLIError: Please run 'az login' to setup account.

### In which platform does it happen?

[...]

PR #579, opened 2019-02-28
Jeremr/python databricks setup

jreynolds01:
### Description
Change the databricks installation and preparation scripts to leverage python to be more cross-platform.

This will help get to the point of also making these operations a bit more modular as well.

### Related Issues

#570 
#564 
#567 

### Checklist:

- [ ] I have added tests. (no tests for databricks/azure resources yet.)
- [x] I have updated the documentation accordingly. (will update README with reviews in shortly; want to have people look at it if possible.)


jreynolds01:
updated setup.md, so docs are updated.

SETUP.md
```diff
@@ -169,19 +169,23 @@ This option utilizes an installation script to do the setup, and it requires add
...
 ```{shell}
 cd Recommenders
-./scripts/databricks_install.sh <CLUSTER_ID>
+./scripts/databricks_install.py <CLUSTER_ID>
 ```
 
+**Note** If you are planning on running through the sample code for operationalization [here](notebooks//05_operationalize/als_movie_o16n.ipynb), you need to prepare the cluster for operationalization. You can do so by adding an additional option to the script run:
+
+```{shell}
+./scripts/databricks_install.py --prepare-o16n <CLUSTER_ID>
```
yueguoguo: Would be good to explain where to get the <CLUSTER_ID>.

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env python
+
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# This script installs Recommenders/reco_utils as a library into Databricks
```
yueguoguo: Very minor point, to be more precise, it would be good to add `egg` library into Databricks.

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+COSMOSDB_JAR_FILE_OPTIONS = {
+  '3': 'https://search.maven.org/remotecontent?filepath=com/microsoft/azure/azure-cosmosdb-spark_2.2.0_2.11/1.1.1/azure-cosmosdb-spark_2.2.0_2.11-1.1.1-uber.jar',
+  '4': 'https://search.maven.org/remotecontent?filepath=com/microsoft/azure/azure-cosmosdb-spark_2.3.0_2.11/1.2.2/azure-cosmosdb-spark_2.3.0_2.11-1.2.2-uber.jar',
+  '5': 'https://search.maven.org/remotecontent?filepath=com/microsoft/azure/azure-cosmosdb-spark_2.4.0_2.11/1.3.5/azure-cosmosdb-spark_2.4.0_2.11-1.3.5-uber.jar'
+}
+
+PYPI_O16N_LIBS=["azure-cli==2.0.56", "azureml-sdk[databricks]==1.0.8", "pydocumentdb==2.3.3"]
+
+
+def create_egg(path_to_recommenders_repo_root=os.getcwd(),local_eggname='Recommenders.egg', overwrite=False):
```
yueguoguo: We use [`black`](https://github.com/ambv/black) to format Python script to follow PEP8 standard. It would be great to do so.

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+
+  databricks --profile """+args.profile+""" clusters list
+  """
+  )
+    
+    
+sys.exit()
+
+#######################################################
+## STOP HERE
```
yueguoguo: Wondering if the existence of the cluster can be detected programatically so that we can save these comments...

SETUP.md
```diff
@@ -169,19 +169,23 @@ This option utilizes an installation script to do the setup, and it requires add
...
 Once you have confirmed the databricks cluster is *RUNNING*, install the modules within this repository with the following commands:
 
 ```{shell}
 cd Recommenders
-./scripts/databricks_install.sh <CLUSTER_ID>
+./scripts/databricks_install.py <CLUSTER_ID>
 ```
 
+**Note** If you are planning on running through the sample code for operationalization [here](notebooks//05_operationalize/als_movie_o16n.ipynb), you need to prepare the cluster for operationalization. You can do so by adding an additional option to the script run:
+
```
miguelgfierro: typo in `notebooks//05_operationalize/als_movie_o16n.ipynb` -> `notebooks/05_operationalize/als_movie_o16n.ipynb`

SETUP.md
miguelgfierro: +1

SETUP.md
```diff
@@ -220,24 +224,16 @@ import reco_utils
...
+This repository includes an end-to-end example notebook that uses Azure Datbaricks to estimate a recommendation model using Alternating Least Squares, writes pre-computed recommendations to Azure Cosmos DB, and then creates a real-time scoring service that retrieves the recommendations from Cosmos DB. In order to execute that [notebook](notebooks//05_operationalize/als_movie_o16n.ipynb), you must install the Recommenders repository as a library (as described above), **AND** you must also install some additional dependencies. With the *Quick install* method, you just need to pass an additional option to the `scripts/databricks_install.py` script.
```
miguelgfierro: another minor comment, maybe add a link in databricks_install.py

SETUP.md
```diff
@@ -249,9 +245,9 @@ This script does all of the steps described in the *Manual setup* section below.
 
 You must install three packages as libraries from PyPI:
 
-* `azure-cli`
-* `azureml-sdk[databricks]`
-* `pydocumentdb`
+* `azure-cli==2.0.56`
```
miguelgfierro: here there are some versions that are not the same as in the conda installer. I agree that is safer to fix the versions. I think we are going to forget that there are different versions. Can't think of a good solution for this :-(

scripts/databricks_install.py
miguelgfierro: are these comments needed?

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+CLUSTER_NOT_RUNNING_MSG = """
+    Cluster {0} found, but it is not running. Status={1}
+    You can start the cluster with 'databricks clusters start --cluster-id {0}'.
+    Then, check the cluster status by using 'databricks clusters list' and
+    re-try installation once the status becomes 'RUNNING'.
+    """
+""
+
+## Variables for operationalization:
+COSMOSDB_JAR_FILE_OPTIONS = {
```
miguelgfierro: in the near future we would need MMLSpark, we can add it here, or if you prefer to wait until the Ligthgbm work is done, we have to remember to add it later

scripts/generate_conda_file.py
```diff
@@ -70,6 +70,7 @@
     "nvidia-ml-py3": "nvidia-ml-py3>=7.352.0",
     "papermill": "papermill>=0.15.0",
     "pydocumentdb": "pydocumentdb>=2.3.3",
+    "databricks-cli": "databricks-cli>=0.8.4"
```
miguelgfierro: how confident are we that databricks-cli is a stable library with few breaking changes? in fastai, we had to change from  `>=` to `==` because of breaking changes

scripts/databricks_install.py
jreynolds01: yeah, that's why i left them there. I was going to add that as a feature, but wanted to recreate the current functionality and have people take a look before I did too much addition. I can add that in the next couple of days.

scripts/databricks_install.py
jreynolds01: will be happy to go ahead and add it now since I'm doing a lot with it.

SETUP.md
jreynolds01: added a note. It's described in an earlier section, but I added a note here as well.

scripts/databricks_install.py
jreynolds01: updated

scripts/databricks_install.py
jreynolds01: Thanks for the reminder - I will update with that.

SETUP.md
jreynolds01: fixed

scripts/databricks_install.py
jreynolds01: no, i was just keeping them for adding a feature in the future to actually create a cluster. I will add that and clean this up. Just wanted feedback first

scripts/generate_conda_file.py
jreynolds01: no idea, so pinned.

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+  if os.path.exists(local_eggname) and overwrite:
+    os.unlink(local_eggname)
+  os.rename(myzipfile,local_eggname)
+  return local_eggname
+
+def prepare_for_operationalization(cluster_id, api_client, status=None, dbfs_path="dbfs:/FileStore/jars", overwrite=False):
+  """
+  Installs appropriate versions of several libraries to support operationalization.
+  """
+  print("Preparing for operationliazation...")
```
gramhagen: typo in 'operationalization' (if it were me I would have spelled it differently every time i used it!)

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+  if overwrite or not os.path.exists(local_jarname):
+    print('Downloading {}...'.format(cosmosdb_jar_url))
+    local_jarname, _ = urlretrieve(cosmosdb_jar_url,local_jarname)
+  else:
+    print('File {} already downloaded.'.format(local_jarname))
+
+  ## upload to dbfs:
+  uploadpath='/'.join([dbfs_path, local_jarname])
+  print("Uploading CosmosDB driver to databricks at {}".format(uploadpath))
+  DbfsApi(api_client).cp(recursive=False, src=local_jarname, dst=uploadpath, overwrite=overwrite)
```
gramhagen: is there any way to know if the file existed and was not overwritten after this call? might be useful to inform the user if they reran it without the overwrite flag

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+    print('appending .egg to end of name.')
+    logal_eggname = local_eggname+'.egg'
+  
+  ## overwrite egg if it previously existed
+  if os.path.exists(local_eggname) and overwrite:
+    os.unlink(local_eggname)
+  os.rename(myzipfile,local_eggname)
+  return local_eggname
+
+def prepare_for_operationalization(cluster_id, api_client, status=None, dbfs_path="dbfs:/FileStore/jars", overwrite=False):
```
gramhagen: do you need default values here for dbfs_path and overwrite? it's one more thing to keep consistent with argparse defaults.

scripts/databricks_install.py
gramhagen: do you need default values here for dbfs_path and overwrite? it's one more thing to keep consistent with argparse defaults.

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+  """
+  ## create the zip archive:
+  myzipfile = shutil.make_archive('reco_utils',
+                'zip', 
+                root_dir = path_to_recommenders_repo_root, 
+                base_dir = 'reco_utils'
+              )
+
+  ## make sure extension is .egg
+  if local_eggname[-4:] != ".egg":
```
gramhagen: it might be better to check / correct the eggname value right after the args are parsed and pass that into this function (so it will be consistent across other references), plus this could generate an index error if the input is too short
```
if not args.eggname.endswith('.egg'):
    args.eggname += '.egg'
```

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+  local_jarname = os.path.basename(cosmosdb_jar_url)
+  ## only download if you need it:
+  if overwrite or not os.path.exists(local_jarname):
+    print('Downloading {}...'.format(cosmosdb_jar_url))
+    local_jarname, _ = urlretrieve(cosmosdb_jar_url,local_jarname)
+  else:
+    print('File {} already downloaded.'.format(local_jarname))
+
+  ## upload to dbfs:
+  uploadpath='/'.join([dbfs_path, local_jarname])
```
gramhagen: same comment about using pathlib

scripts/databricks_install.py
gramhagen: also, do you need status? basically we just want the spark_version? I would suggest just putting the spark_version as an input and extracting that from the status external to this function

SETUP.md
jreynolds01: I think trying to keep pydocumentdb and azure-cli on same version as rest of repo is good, as long as we can test it :).

The bigger issue is azureml-sdk[datbricks]; it's not exactly the same as azureml-sdk, so I'm ok with having it as a different version, but at the same time, we need to decide on the resolution of the authentication issues that are popping up in places like #529 so that we can eventually get to a place of being current with it

miguelgfierro: makes sense

scripts/databricks_install.py
```diff
@@ -0,0 +1,255 @@
...
+
+  ###############################
+  ## Interact with Databricks:
+  ###############################
+
+  ## first make sure you are using the correct profile and connecting to the intended workspace
+  my_api_client = _get_api_client(ProfileConfigProvider(args.profile).get_config())
+
+  ## upload the egg:
+  uploadpath='/'.join([args.dbfs_path, os.path.basename(myegg)])
```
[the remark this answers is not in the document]
miguelgfierro: I'm curious, what is the difference between Path vs os.path?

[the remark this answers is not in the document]
gramhagen: pathlib seems to be the new standard recommendation for dealing with paths in python3, it is a bit more concise and efficient, but the main advantage here is you can generate proper paths for OSs that are not what you are currently working on. So in this case it makes it easy to generate a posix path even though the user is on a Windows system.

scripts/databricks_install.py
loomlike: I mean, not in the future but now ;-)

scripts/databricks_install.py
```diff
@@ -0,0 +1,272 @@
...
+    if os.path.exists(local_eggname) and overwrite:
+        os.unlink(local_eggname)
+    os.rename(myzipfile, local_eggname)
+    return local_eggname
+
+
+def prepare_for_operationalization(
+    cluster_id,
+    api_client,
+    status=None,
```
loomlike: For `status`, would be good to have either docstring or better name like `cluster_status`. btw, should we pass this here? can we just check the status from this function instead? Also, previously checked status could be changed at this moment.

scripts/databricks_install.py
```diff
@@ -0,0 +1,272 @@
...
+        print(textwrap.dedent(CLUSTER_NOT_FOUND_MSG.format(args.cluster_id)))
+        raise
+
+    if status["state"] != "RUNNING":
+        print(
+            textwrap.dedent(
+                CLUSTER_NOT_RUNNING_MSG.format(args.cluster_id, status["state"])
+            )
+        )
+        sys.exit()
```
loomlike: Can we force-start cluster here, wait until done, and then install egg instead of exit? At this moment, a user's intention is to install the repo anyway.

scripts/databricks_install.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+
+    ## restart the cluster for new installation(s) to take effect.
+    print("Restarting databricks cluster {}".format(args.cluster_id))
+    ClusterApi(my_api_client).restart_cluster(args.cluster_id)
+
+    ## wrap up and send out a final message:
+    print(
+        """
+  Requests submitted. You can check on status of your cluster with:
```
loomlike: Shorter indentation here too.

gramhagen:
another general comment, I think it would be handy to create the Recommenders egg with branch / version info if possible: e.g.: Recommenders-v2019.01-staging

thoughts? otherwise it can be very confusing to understand what version of the egg is being used on databricks

yueguoguo:
> another general comment, I think it would be handy to create the Recommenders egg with branch / version info if possible: e.g.: Recommenders-v2019.01-staging

thoughts? otherwise it can be very confusing to understand what version of the egg is being used on databricks

It would be great that we could fix the set-up here for just using the release which is the stable version of the repo.

miguelgfierro:
> another general comment, I think it would be handy to create the Recommenders egg with branch / version info if possible: e.g.: Recommenders-v2019.01-staging

this can be done programatically using the version here: 
https://github.com/Microsoft/Recommenders/blob/master/reco_utils/__init__.py#L8

scripts/databricks_install.py
jreynolds01: thanks for catching that!

scripts/databricks_install.py
jreynolds01: it fails if the file exists and overwrite=False. Added a helper function to alert the user to when it is being overwritten., we could also change behavior if we want to allow it to continue if the file exists.

scripts/databricks_install.py
jreynolds01: done for an initial attempt. Right now, it's not configurable, and it creates a cluster with the same default values they would get in the portal. We can add customization later. The biggest issue with configuring customizable clusters is that there are so many variables that it becomes tedious to parse them all. I think we'd want people to pass a `--cluster-config xx.json` argument that provides a path to a json file that can be parsed. Does that seem reasonable @loomlike?

scripts/databricks_install.py
jreynolds01: adjusted to just take spark_version per Scott's comments.

scripts/databricks_install.py
jreynolds01: i did a version of this with pending status, we could generalize and just make it the default behavior... The issue is just the waiting, but I suppose they'd be waiting no matter whether we do it for them or not.

jreynolds01:
> another general comment, I think it would be handy to create the Recommenders egg with branch / version info if possible: e.g.: Recommenders-v2019.01-staging
> 
> thoughts? otherwise it can be very confusing to understand what version of the egg is being used on databricks

Definitely considered this, and agree that this would be good. The issue that I worry about is two-fold.

1. Is the un-installation of other prior versions. that requires a bit more overhead if they're not the same name. If the file gets overwritten, then it has to install from that same location. If we start versioning the files (which I agree is a good idea), then we need to do more to check whether a library is already installed and manually uninstall. We could just regex for Recommenders-XXX.egg in the results of a databricks libraries list
2. I view the ability to install from any branch as a feature, because it's super helpful to be able to test things rapidly. It would definitely have more value if we could also version the egg file as well.

gramhagen:
Those are good points, i think being able to overwrite easily is better. One option might be creating a temporary file prior to compression which would allow you to inspect the branch / commit that was used. We could modify reco_utils.__init__.py to look for this file and if it exists read in the git details. Anyway, I don't want to hold this PR up. We can add this as a future feature.

miguelgfierro:
hey @jreynolds01 is there more work here to be done?

gramhagen:
one minute, I'm updating some of the docs

gramhagen:
Ok, should be all set once tests pass, I testing recreating the reco_spark.yaml, updating my conda env, uploading the recommenders egg (failed with existing file, works with overwrite flag) and install mmlspark.

nice work @jreynolds01

jreynolds01:
Thanks Scott and Miguel for updates.  I think it's ready.

Issue #445, opened 2019-01-24
Add the metric AUC 


miguelgfierro:
This can be used by any algo that implemented regression metrics

yueguoguo:
To clarify, AUC is meant for binary classification, right? Probably this would be useful to the content-based algorithms for predicting hit/not-hit?

miguelgfierro:
yes, that's exactly the reason. It was used in xDeepFM

miguelgfierro:
Metrics to implement:
- limited AUC
- average AUC
- direct AUC from sklearn

PR #580, opened 2019-02-28
AUC and Logloss

yueguoguo:
### Description
Discussed with @Leavingseason and added
* AUC metric
* Logloss metric

### Related Issues
#445 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/evaluation/python_evaluation.py
+++ b/reco_utils/evaluation/python_evaluation.py
@@ -9,6 +9,8 @@ from sklearn.metrics import (
     mean_absolute_error,
     r2_score,
     explained_variance_score,
+    roc_auc_score,
+    log_loss
 )
 
 from reco_utils.common.constants import (
@@ -621,6 +623,18 @@ def auc(
     Return:
         float: auc (min=0, max=1).
     """
+    rating_true_pred = merge_rating_true_pred(
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+    auc_score = roc_auc_score(
+        np.asarray(rating_true_pred[DEFAULT_RATING_COL]),
+        np.asarray(rating_true_pred[PREDICTION_COL])
+    )
+
+    # Round the AUC score to four digits.
+    auc_score = round(auc_score, 4)
+
+    return auc_score
 
 
 # This is a placeholder.
@@ -634,7 +648,11 @@ def logloss(
 ):
     """
     Calculate the logloss metric for implicit feedback typed
-    recommender.
+    recommender, where rating is binary and prediction is float number ranging
+    from 0 to 1.
+
+    https://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_the_curve
+
 
     TODO: need a reference.
 
@@ -652,6 +670,25 @@ def logloss(
     Return:
         float: auc (min=0, max=1).
     """
+    rating_true_pred = merge_rating_true_pred(
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+
+    # To make sure that there is no N/A in the logloss metric calculation.
+    rating_true_pred[PREDICTION_COL] = (
+        rating_true_pred[PREDICTION_COL].apply(lambda x: max(min(x, 1.0 - 10e-12), 10e-12))
+    )
+
+    log_loss_score = log_loss(
+        np.asarray(rating_true_pred[DEFAULT_RATING_COL]),
+        np.asarray(rating_true_pred[PREDICTION_COL])
+    )
+
+    # Round the AUC score to four digits.
+    log_loss_score = round(log_loss_score, 4)
+
+    return log_loss_score
+
 
 
 def get_top_k_items(
--- a/reco_utils/evaluation/python_evaluation.py
+++ b/reco_utils/evaluation/python_evaluation.py
@@ -593,6 +593,7 @@ def map_at_k(
     return np.float64(df_sum_all.agg({"map": "sum"})) / n_users
 
 
+@check_column_dtypes
 # This is a placeholder.
 def auc(
     rating_true,
@@ -609,6 +610,12 @@ def auc(
 
     https://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_the_curve
 
+    Note:
+        The evaluation does not require a leave-one-out scenario.
+        This metric does not calculate group-based AUC which considers the AUC scores
+        averaged across users. It is also not limited to k. Instead, it calculates the
+        scores on the entire prediction results regardless the users.
+
     Args:
         rating_true (pd.DataFrame): True data.
         rating_pred (pd.DataFrame): Predicted data.
@@ -616,12 +623,9 @@ def auc(
         col_item (str): column name for item.
         col_rating (str): column name for rating.
         col_prediction (str): column name for prediction.
-        relevancy_method (str): method for getting the most relevant items.
-        k (int): number of top k items per user.
-        threshold (float): threshold of top items per user (optional).
 
     Return:
-        float: auc (min=0, max=1).
+        float: auc_score (min=0, max=1).
     """
     rating_true_pred = merge_rating_true_pred(
         rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
@@ -637,7 +641,7 @@ def auc(
     return auc_score
 
 
-# This is a placeholder.
+@check_column_dtypes
 def logloss(
         rating_true,
         rating_pred,
@@ -651,10 +655,7 @@ def logloss(
     recommender, where rating is binary and prediction is float number ranging
     from 0 to 1.
 
-    https://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_the_curve
-
-
-    TODO: need a reference.
+    https://en.wikipedia.org/wiki/Loss_functions_for_classification#Cross_entropy_loss_(Log_Loss)
 
     Args:
         rating_true (pd.DataFrame): True data.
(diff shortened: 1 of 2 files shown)


yueguoguo:
Updates in the notebooks will be given in another PR for the convenience of review.

reco_utils/evaluation/python_evaluation.py
```diff
@@ -591,6 +593,101 @@ def map_at_k(
...
+        col_prediction (str): column name for prediction.
+
+    Return:
+        float: auc_score (min=0, max=1).
+    """
+    rating_true_pred = merge_rating_true_pred(
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+    auc_score = roc_auc_score(
+        np.asarray(rating_true_pred[DEFAULT_RATING_COL]),
```
miguelgfierro: not 100%, but I would say that `rating_true_pred[DEFAULT_RATING_COL].values` is faster

reco_utils/evaluation/python_evaluation.py
```diff
@@ -591,6 +593,101 @@ def map_at_k(
...
+    rating_true_pred = merge_rating_true_pred(
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+
+    # To make sure that there is no N/A in the logloss metric calculation.
+    rating_true_pred[PREDICTION_COL] = (
+        rating_true_pred[PREDICTION_COL].apply(lambda x: max(min(x, 1.0 - 10e-12), 10e-12))
+    )
+
+    log_loss_score = log_loss(
```
miguelgfierro: same comment as before

tests/unit/test_python_evaluation.py
```diff
@@ -38,75 +41,89 @@ def target_metrics(scope="module"):
...
-                9,
-                8,
-                7,
-                6,
-                5,
-            ],
-        }
-    )
-    return rating_true, rating_pred, rating_nohit
+    def _generate_python_data(binary_rating=False):
```
miguelgfierro: don't understand why this is needed, instead of directly returning the ratings

tests/unit/test_python_evaluation.py
```diff
@@ -38,75 +41,89 @@ def target_metrics(scope="module"):
...
+            # Normalize the prediction.
+            rating_pred[PREDICTION_COL] = minmax_scale(rating_pred[PREDICTION_COL].astype(float))
+
+        return rating_true, rating_pred, rating_nohit
+    return _generate_python_data
 
 
 def test_column_dtypes_match(python_data):
-    rating_true, rating_pred, _ = python_data
+    rating_true, rating_pred, _ = python_data()
```
miguelgfierro: here we are not using python_data as fixture but as function. The correct way to add it would be just `python_data`.  Here more about fixtures https://docs.pytest.org/en/latest/fixture.html

tests/unit/test_python_evaluation.py
yueguoguo: Because a function with argument `binary_rating` is returned from the fixture to allow testing different types of metrics, rating metric and binary prediction metric, respectively.

tests/unit/test_python_evaluation.py
yueguoguo: To pass a function, then I can use the parameter. See comment above.

tests/unit/test_python_evaluation.py
miguelgfierro: yeah, but then we are  using it as fixtures. Let me ping you offline

yueguoguo: See https://alysivji.github.io/pytest-fixures-with-function-arguments.html

miguelgfierro: oh I see what you are doing, sorry I didn't understand it initially. Yeah, you are right, this is correct. I think it is because I didn't see the argument. 
Would it make sense to set explicitly `binary_rating` in all the instances to make it clearer?

yueguoguo: @miguelgfierro great point - will do that.

reco_utils/evaluation/python_evaluation.py
```diff
@@ -591,6 +593,100 @@ def map_at_k(
...
+    rating_true_pred = merge_rating_true_pred(
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+    auc_score = roc_auc_score(
+        np.asarray(rating_true_pred[DEFAULT_RATING_COL].values),
+        np.asarray(rating_true_pred[PREDICTION_COL].values)
+    )
+
+    # Round the AUC score to four digits.
+    auc_score = round(auc_score, 4)
```
anargyri: Sorry to be a bit late to the PR! 
This is a magic number. Why do we need to round?

reco_utils/evaluation/python_evaluation.py
```diff
@@ -591,6 +593,100 @@ def map_at_k(
...
+        rating_true_pred[PREDICTION_COL].apply(lambda x: max(min(x, 1.0 - 10e-12), 10e-12))
+    )
+
+    log_loss_score = log_loss(
+        np.asarray(rating_true_pred[DEFAULT_RATING_COL].values),
+        np.asarray(rating_true_pred[PREDICTION_COL].values)
+    )
+
+    # Round the AUC score to four digits.
+    log_loss_score = round(log_loss_score, 4)
```
anargyri: Same here, this is a magic number.

reco_utils/evaluation/python_evaluation.py
```diff
@@ -591,6 +593,100 @@ def map_at_k(
...
+    Return:
+        float: log_loss_score (min=-\inf, max=\inf).
+    """
+    rating_true_pred = merge_rating_true_pred(
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+
+    # To make sure that there is no N/A in the logloss metric calculation.
+    rating_true_pred[PREDICTION_COL] = (
+        rating_true_pred[PREDICTION_COL].apply(lambda x: max(min(x, 1.0 - 10e-12), 10e-12))
```
anargyri: Again this is a magic number. Btw 10e-12 is larger than 1e-12.
You could also use `numpy.clip`
https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.clip.html

reco_utils/evaluation/python_evaluation.py
```diff
@@ -591,6 +593,100 @@ def map_at_k(
...
+        col_prediction (str): column name for prediction.
+
+    Return:
+        float: auc_score (min=0, max=1).
+    """
+    rating_true_pred = merge_rating_true_pred(
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+    auc_score = roc_auc_score(
+        np.asarray(rating_true_pred[DEFAULT_RATING_COL].values),
```
anargyri: `.values` returns already a numpy array.

reco_utils/evaluation/python_evaluation.py
```diff
@@ -591,6 +593,100 @@ def map_at_k(
...
+        rating_true, rating_pred, col_user, col_item, col_rating, col_prediction
+    )
+
+    # To make sure that there is no N/A in the logloss metric calculation.
+    rating_true_pred[PREDICTION_COL] = (
+        rating_true_pred[PREDICTION_COL].apply(lambda x: max(min(x, 1.0 - 10e-12), 10e-12))
+    )
+
+    log_loss_score = log_loss(
+        np.asarray(rating_true_pred[DEFAULT_RATING_COL].values),
```
anargyri: Same about `.values`

reco_utils/evaluation/python_evaluation.py
anargyri: There are redundant parentheses.

tests/unit/test_python_evaluation.py
```diff
@@ -38,75 +41,89 @@ def target_metrics(scope="module"):
...
+                    7,
+                    6,
+                    5,
+                ],
+            }
+        )
+
+        if binary_rating:
+            # Convert to binary case.
+            rating_true[DEFAULT_RATING_COL] = (
```
anargyri: No need to have parentheses.

tests/unit/test_python_evaluation.py
```diff
@@ -38,75 +41,89 @@ def target_metrics(scope="module"):
...
+                rating_true[DEFAULT_RATING_COL].apply(lambda x: 1.0 if x >= 3 else 0.0)
```
anargyri: Can use `numpy.clip`.

tests/unit/test_python_evaluation.py
```diff
@@ -307,8 +324,42 @@ def test_python_recall(python_data, target_metrics):
...
 
 
+def test_python_auc(python_data, target_metrics):
+    rating_true, rating_pred, _ = python_data(binary_rating=True)
+
+    assert auc(
+        rating_true=rating_true,
+        rating_pred=rating_true,
+        col_prediction=DEFAULT_RATING_COL
+    ) == pytest.approx(1.0, 0.1)
```
anargyri: The tolerance for this should be low. Also define a constant.

tests/unit/test_python_evaluation.py
```diff
@@ -307,8 +324,42 @@ def test_python_recall(python_data, target_metrics):
...
+
+
+def test_python_logloss(python_data, target_metrics):
+    rating_true, rating_pred, _ = python_data(binary_rating=True)
+
+    assert logloss(
+        rating_true=rating_true,
+        rating_pred=rating_true,
+        col_prediction=DEFAULT_RATING_COL
+    ) == pytest.approx(0, 0.1)
```
anargyri: Again magic number.

reco_utils/evaluation/python_evaluation.py
yueguoguo: Yea it wrapped the codes in deeprec implementation. Removed rounding for consistency.

reco_utils/evaluation/python_evaluation.py
yueguoguo: Ah the parentheses is generated from `black` but yea we can remove it.

yueguoguo: Even clip is not needed - this is handled with the `log_loss` function in `sklearn` directly. The default value is 1e-15. https://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html

tests/unit/test_python_evaluation.py
yueguoguo: This was because of `black` but I removed them anyway.

Issue #542, opened 2019-02-20
Test spark hypertune notebook


anargyri:
### Description

There are no tests for the notebook for Spark tuning  https://github.com/Microsoft/Recommenders/blob/master/notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb 

### Expected behavior with the suggested feature

Should we add unit, smoke and integration tests, or just unit tests? What are the scalability limitations?

yueguoguo:
Great suggestion. Is it reasonable to have these after #432 and #450 are resolved, or the tests can be run on the existing one-node DevOps machine?

loomlike:
This will applicable to aml_* notebooks too (currently, one in #448 also doesn't have tests). The thing is, it is very tricky to test. Should start a run and wait until the run finishes and check its status with [get_status()](https://docs.microsoft.com/en-us/python/api/azureml-core/azureml.core.run.run?view=azure-ml-py#get-status--) to see if it is "Completed" or "Failed".

anargyri:
I opened this issue only for this notebook. This notebook is more straightforward to test because it does not use AML. For the other notebooks I opened a separate discussion here #547

yueguoguo:
Fixed in #583

PR #583, opened 2019-02-28
Unit test for spark hypertune notebook

yueguoguo:
### Description
Unit test for the Spark hypertune notebook.

### Related Issues
#542 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
+++ b/notebooks/04_model_select_and_optimize/hypertune_spark_deep_dive.ipynb
@@ -95,7 +95,11 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
+    "tags": [
+     "parameters"
+    ]
+   },
    "outputs": [],
    "source": [
     "NUMBER_CORES = 1\n",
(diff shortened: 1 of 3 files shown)


yueguoguo:
@miguelgfierro @anargyri this test takes some time and it resulted in time-out failure.

miguelgfierro:
I see, what we have done in the past is either to move this to smoke/integration test or to add parameters in the unit tests so it makes as less loops as possible

yueguoguo:
> I see, what we have done in the past is either to move this to smoke/integration test or to add parameters in the unit tests so it makes as less loops as possible

But this is not a smoke/integration test - it is just to run the notebook and check if there is any failure, like other unit tests for notebooks. The only thing is that the notebook itself takes some time to finish...

miguelgfierro:
yeah, that's right, then maybe we can parametrize it and reduce the number of runs. Do you know how long does it take to run this notebook with the current parameters?

yueguoguo:
> yeah, that's right, then maybe we can parametrize it and reduce the number of runs. Do you know how long does it take to run this notebook with the current parameters?

~30 - 40 min

miguelgfierro:
oh that's a lot for a unit test

yueguoguo:
@miguelgfierro parameterize the notebook and it now takes 3~4 mins 🚀

Issue #584, opened 2019-02-28
[ASK] MAP_AT_K cross check


seancaodo:
### Description

Hi,

I was doing some calculations for MAP_AT_K off SAR example: https://github.com/Microsoft/Recommenders/blob/master/notebooks/00_quick_start/sar_movielens.ipynb
and noticed the difference between MAP and Precision@K: 0.105815 and 0.326617.

The Precision@K is re-creatable at 0.326617
######### Precision@K
new_df = pd.merge(top_k, test, how='left', on=['UserId','MovieId'])
from dplython import (DplyFrame, X, select, sift, head, arrange, mutate, group_by, summarize, DelayFunction)
new_df = DplyFrame(new_df)
where = DelayFunction(np.where)
new_df = new_df >> mutate(Relevant = where(X.Rating > 0, 1, 0))
dd2 = new_df >> group_by(X.UserId) >> summarize(mm = X.Relevant.sum()/10)
dd2['mm'].mean()
######### End of Precision@K

But MAP might open for a discussion:
######### MAP 1
new_df = pd.merge(top_k, test, how='left', on=['UserId','MovieId'])
from dplython import (DplyFrame, X, select, sift, head, arrange, mutate, group_by, summarize, DelayFunction)
new_df = DplyFrame(new_df)
where = DelayFunction(np.where)
new_df = new_df >> mutate(Relevant = where(X.Rating > 0, 1, 0))
new_df = new_df >> arrange(X.UserId, -X.prediction)
new_df['ForCount'] = 1
new_df['cumsum'] = new_df.groupby('UserId')['Relevant'].transform(pd.Series.cumsum)
new_df['cumsum2'] = new_df.groupby('UserId')['ForCount'].transform(pd.Series.cumsum)
new_df['prec'] = new_df['cumsum'] / new_df['cumsum2']
new_df['prec'] = new_df['prec'] * new_df['Relevant']
dd2 = new_df >> group_by(X.UserId) >> summarize(ss = X.prec.sum())
dd2['prec2'] = dd2['ss'] / 10
dd2['prec2'].mean()
######### End of MAP 1

The above lines had MAP calculated as 0.2271

Instead of dividing the precision rates by 10, if we divide them by total Movies rated by user in test set like below:

############## MAP 2
new_df = pd.merge(top_k, test, how='left', on=['UserId','MovieId'])
from dplython import (DplyFrame, X, select, sift, head, arrange, mutate, group_by, summarize, DelayFunction)
[...]

PR #588, opened 2019-02-28
EVALUATION: fix link to reference and upate docstring

yueguoguo:
### Description
Fixed the broken link in the MAP evaluator.

### Related Issues
#584

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/reco_utils/evaluation/python_evaluation.py
+++ b/reco_utils/evaluation/python_evaluation.py
@@ -531,11 +531,17 @@ def map_at_k(
     threshold=DEFAULT_THRESHOLD,
 ):
     """
-    Get mean average precision at k. A good reference can be found at
-    https://people.cs.umass.edu/~jpjiang/cs646/03_eval_basics.pdf
+    The implementation of the MAP is referenced from Spark MLlib evaluation metrics.
+    https://spark.apache.org/docs/2.3.0/mllib-evaluation-metrics.html#ranking-systems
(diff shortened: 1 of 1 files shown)


reco_utils/evaluation/python_evaluation.py
```diff
@@ -531,11 +531,17 @@ def map_at_k(
...
     Get mean average precision at k. A good reference can be found at
-    https://people.cs.umass.edu/~jpjiang/cs646/03_eval_basics.pdf
+    http://web.stanford.edu/class/cs276/handouts/EvaluationNew-handout-6-per.pdf
 
-    NOTE: The MAP is at k because the evaluation class takes top k items for
-    the prediction items.
+    Note:
+        1. The MAP is at k because the evaluation class takes top k items for
+        the prediction items.
+        2. The MAP is to calculate the AP in the relevant items, so it is normalized
```
gramhagen: I would say: The MAP is meant to calculate Avg. Precision for the relevant items, so it is normalized by the number of relevant items in the ground truth data, instead of k.

yueguoguo: Thanks @gramhagen this is more precise! The original note #1 is to explain the name of the function as in some context (eg Spark) it is called just “MAP” instead of “MAP@k”. I think it also makes sense as it indeed is not normalised by k.

PR #589, opened 2019-02-28
Benchmark discussion

miguelgfierro:
### Description

This is more for discussion that for merging

Several ideas were suggested:
1) have a benchmark folder in the repo root or inside notebooks
2) use scripts instead of notebooks, or use the notebook to show the results
3) Save the results to a json file which can be read by other notebooks or scripts
4) Split this notebook into 3 (cpu/gpu/spark) or even have one script for each algo to be computed in parallel in AzureML
4) Instead of using something like this code, modify the notebooks to have the same data and make a benchmark that uses papermill to run the other notebooks.
5) we need to make sure that the results are comparable

The objective here is to let users compare different algos. A path that users might have is to look at the summary table at the end to understand which algos are performing well, then they might try to add their own data and generate the results. Then maybe they will go to the deep dive notebooks to understand better the algo that has the better performance.

Please note that I haven't added /modified comments in this notebook

### Checklist:

- [ ] I have added tests.


yueguoguo:
Great proposal! The following are just my very personal opinions:

> have a benchmark folder in the repo root or inside notebooks
Currently, we put the things into `model_select_and_optimize`, we can change if it is not a desirable place. It might be good to put the notebook into a dedicated folder but show the results in the front page (i.e., README).

> * use scripts instead of notebooks, or use the notebook to show the results
Either is fine, because a notebook can be converted to a script anyway 😄 . But as discussed in the meeting, personally I think it might be neater to standardize the "quickstart" notebooks and run the comparison with `papermill`. 

> * Save the results to a json file which can be read by other notebooks or scripts
Good idea.

> * Split this notebook into 3 (CPU/GPU/Spark) or even have one script for each algo to be computed in parallel in AzureML
I think this can be done more easily if we just run each notebook onto a different remote instance (i.e., a DSVM or a Databricks cluster for Spark, a GPU DSVM or AML Compute for GPU algorithms, and a CPU DSVM or AML Compute for classic algorithms, respectively). This will also be in line with the on-going work with AML integration with papermill proposed by @danielsc 

> * Instead of using something like this code, modify the notebooks to have the same data and make a benchmark that uses papermill to run the other notebooks.
Personally, I think this is a desirable approach.

> * we need to make sure that the results are comparable
Very important point. The quantitative comparison should be apple-to-apple, this should be carefully done in terms of evaluation protocol (i.e., data split, evaluation metric, etc.), hyperparameter tuning, etc. Maybe we can agree on the execution pipeline, and then design the protocol - I guess the latter requires some discussion.

gramhagen:
I think this scales better if the individual algorithms are separate scripts or notebooks (notebooks retain consistency so that's how I would lean). If there are common pieces of code we should consider refactoring them into the utils. Also if they are separated it becomes easier to tune them.

I like the idea of setting up the data and kicking off all the notebooks via papermill, it basically becomes a big integration test suite. Even better to parallelize this on AzureML.

miguelgfierro:
I'm happy to revert the code and follow Le's approach of running the notebooks using papermill and fixing the split of the data. 

I think as a good point, it will cleaner, since we will add less code. 

On the other hand, it will be less flexible since in the notebooks we will have some extra code that we might not need in the benchmark. (This could be addressed, as @yueguoguo suggested by standardizing the quickstart notebooks)

Just recap of the discussion so far:

1. people are ok to have a benchmark folder
2. not strong feeling about using scripts or notebooks for the benchmark
3. people are ok with saving the results to a json
4. We should split this so they can run independently
5. people tend more to revert the code, instead of using factories to define the parts of each algo, use the quickstart notebooks

some more views will be great @jreynolds01 @loomlike @anargyri @wutaomsft @nikhilrj @msalvaris @heatherbshapiro @eisber

heatherbshapiro:
A few comments:
1) We make a lot of reference to "collaborative filtering" algorithms in notebooks, but it would be great if we could describe it in the readme or in this notebook as well. It is in the 
2) Why are there so many blank cells?
3) I think it will be good to have benchmarks folder so that users can easily find it
4) Add more text /comments around the cells to explain what is happening in each cell

bethz:
I've added a benchmarks folder to the recipe template so that we can consistently have it as a folder across projects.

anargyri:
Thanks Miguel, this is really great! Do you plan to add all the algos (Surprise SVD, ALS, VW, NCF) too?
Regarding the discussion, do we also want to save models too? (I don't see any reason, but maybe there is). I also agree with Le that the data should be the same (ideally the splits would be the same, or we could average over several trials).

What you are doing currently is a benchmark for collaborative filtering on a specific data set (Movielens). However, the scope of the repo is broader than this. There will be other use cases soon (XDeepFM, Criteo etc.) and Wide & Deep already falls outside the scope of this comparison. 
What is the broader plan beyond this notebook? It is not fair to "fit" those other use cases into this benchmark. Should we create additional benchmarking notebooks for other use cases e.g. one for feature based Movielens, one for click prediction etc.?

notebooks/03_evaluate/comparison_movielens.ipynb
```diff
@@ -0,0 +1,982 @@
...
+    "Ranking predictions"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def recommend_k_als(model, test, train):\n",
```
anargyri: This is not needed because ALS has its own recommend_k function:
https://spark.apache.org/docs/2.2.0/api/python/pyspark.ml.html#pyspark.ml.recommendation.ALSModel.recommendForAllUsers

notebooks/03_evaluate/comparison_movielens.ipynb
```diff
@@ -0,0 +1,982 @@
...
+    "                                       relevancy_method=\"top_k\")\n",
+    "    return {\n",
+    "        \"MAP\": rank_eval.map_at_k(),\n",
+    "        \"nDCG@k\": rank_eval.ndcg_at_k(),\n",
+    "        \"Precision@k\": rank_eval.precision_at_k(),\n",
+    "        \"Recall@k\": rank_eval.recall_at_k()\n",
+    "    }\n",
+    "    \n",
+    "    \n",
+    "def rating_metrics_python(test, predictions):\n",
```
anargyri: You can avoid repetitions here by using `eval()` since the syntax is the same for all rating metrics functions.

notebooks/03_evaluate/comparison_movielens.ipynb
```diff
@@ -0,0 +1,982 @@
...
+    "        \"RMSE\": rmse(test, predictions, **cols),\n",
+    "        \"MAE\": mae(test, predictions, **cols),\n",
+    "        \"R2\": rsquared(test, predictions, **cols),\n",
+    "        \"Explained Variance\": exp_var(test, predictions, **cols)\n",
+    "    }\n",
+    "    \n",
+    "    \n",
+    "def ranking_metrics_python(test, predictions, k=10):\n",
+    "    cols = {\n",
+    "        \"col_user\": USER_COL, \n",
```
anargyri: Similarly here, you could use `eval(metric)` where metric is the name of the metric.

notebooks/03_evaluate/comparison_movielens.ipynb
```diff
@@ -0,0 +1,982 @@
...
+    "    \n",
+    "rating_evaluator = {\n",
+    "    \"als\": lambda test, predictions: rating_metrics_pyspark(test, predictions),\n",
+    "    \"svd\": lambda test, predictions: rating_metrics_python(test, predictions),\n",
+    "    \"fastai\": lambda test, predictions: rating_metrics_python(test, predictions)\n",
+    "}\n",
+    "    \n",
+    "    \n",
+    "ranking_evaluator = {\n",
+    "    \"als\": lambda test, predictions, k: ranking_metrics_pyspark(test, predictions, k),\n",
```
anargyri: Also here you could reduce code with a couple of for loops.

notebooks/03_evaluate/comparison_movielens.ipynb
```diff
@@ -0,0 +1,982 @@
...
+   "source": [
+    "def rating_metrics_pyspark(test, predictions):\n",
+    "    rating_eval = SparkRatingEvaluation(test, \n",
+    "                                        predictions, \n",
+    "                                        col_user=USER_COL, \n",
+    "                                        col_item=ITEM_COL, \n",
+    "                                        col_rating=RATING_COL, \n",
+    "                                        col_prediction=PREDICTION_COL)\n",
+    "    return {\n",
+    "        \"RMSE\": rating_eval.rmse(),\n",
```
anargyri: I think you could also rewrite this as a for loop, like for the python metrics below. Here you need to use `getattr()`. It is a bit less explicit though.

notebooks/03_evaluate/comparison_movielens.ipynb
```diff
@@ -0,0 +1,982 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def generate_summary(data, algo, k, train_time, time_rating=np.nan, rating_metrics=None, time_ranking=np.nan, ranking_metrics=None):\n",
+    "    summary = {\"Data\": data, \"Algo\": algo, \"K\": k, \"Train time\": train_time, \"Rating time\": time_rating, \"Ranking time\": time_ranking}\n",
+    "    if rating_metrics is None:\n",
```
anargyri: Again here you repeat the names of the metrics. What about defining two lists of rating and ranking metrics in the beginning of this notebook and using dict comprehensions etc. in all other places? If you want to add another metric in the future, you will need to scan through the notebook and find all occurrences of the metrics.

notebooks/03_evaluate/comparison_movielens.ipynb
```diff
@@ -0,0 +1,982 @@
...
+      "CPU times: user 32.2 s, sys: 4.82 s, total: 37 s\n",
+      "Wall time: 40.3 s\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%time\n",
+    "\n",
+    "# For each data size and each algorithm, a recommender is evaluated. \n",
+    "cols = [\"Data\", \"Algo\", \"K\", \"Train time\", \"Rating time\", \"RMSE\", \"MAE\", \"R2\", \"Explained Variance\", \"Ranking time\", \"MAP\", \"nDCG@k\", \"Precision@k\", \"Recall@k\", ]\n",
```
anargyri: Here the cols should be the same as the dict in `generate_summary()`, so better to use a common variable.

reco_utils/recommender/surprise/surprise_utils.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import pandas as pd
+from reco_utils.common.general_utils import invert_dictionary
+
+
+def surprise_trainset_to_df(trainset, col_user="uid", col_item="iid", col_rating="rating"):
```
anargyri: Is this function used anywhere?

miguelgfierro:
@anargyri sorry for the confusion. I did this last commit because @gramhagen and I were trying yesterday to find out what is the problem with this issue https://github.com/Microsoft/Recommenders/issues/631. The current master or staging are failing, but this notebook is working. So today we are going to debug this to fix it.

In relation with this PR, I think the idea supported by more people is to revert this code and use the quickstart notebooks as benchmark. We will need to make sure that the dataset splits that we have are exactly the same.

anargyri:
> @anargyri sorry for the confusion. I did this last commit because @gramhagen and I were trying yesterday to find out what is the problem with this issue #631. The current master or staging are failing, but this notebook is working. So today we are going to debug this to fix it.
> 
> In relation with this PR, I think the idea supported by more people is to revert this code and use the quickstart notebooks as benchmark. We will need to make sure that the dataset splits that we have are exactly the same.

Ok, that makes sense. My comments applied to the earlier commits btw; but since you are going to change the code, they are obsolete.

loomlike:
Agreed to all the points you guys already made. On top of that, just want to **+1 (Strong plus one)** to parallel compute w/ AzureML!

miguelgfierro:
hey, I'm going to close this and will start a different PR with what w agreed. Please don't erase the branch just yet, because I have parts here that will be reused

PR #590, opened 2019-02-28
Small refactoring of pandas_df_util and baseline notebook

loomlike:
### Description
This PR includes changes of baseline notebook to utilize a util function when filter out seen items.
Also other minor changes are:
* Refactor pandas_df_util
* Add concluding remark section to the baseline notebooks

### Checklist:

- [x] I have added tests.


notebooks/02_model/baseline_deep_dive.ipynb
```diff
@@ -196,7 +195,7 @@
     "    header=['UserId', 'MovieId', 'Rating', 'Timestamp']\n",
     ")\n",
     "\n",
-    "data.head()"
+    "display(data.head())"
```
miguelgfierro: about `display`, removing it would give the exact same result as `data.head()`, is that right? if so, I think we should remove it for consistency with the rest of the notebooks

loomlike: I personally like to call 'display' explicitly for two reasons,
1. It shows my intention explicitly, like "I'm gonna display this data here"
2. If the cell printout some lines after data.head(), dataframe will be not showing. If use display(data.head()) and printout texts after that, both are shown up correctly.

But also agree would be good to have consistency. What do you think? I will follow Rome's rule ;-)

miguelgfierro: I'm ok with whichever you prefer, but I would say we should have consistency, so all with display or all without display. Up to you :-)

loomlike: yes sir! I'm changing the notebook. will ping you once done!

notebooks/02_model/baseline_deep_dive.ipynb
```diff
@@ -195,7 +195,7 @@
     "    header=['UserId', 'MovieId', 'Rating', 'Timestamp']\n",
     ")\n",
     "\n",
-    "display(data.head())"
+    "data.head()"
```
gramhagen: this is at least cross-platform, but it will render as just the first row in databricks.
you might want to do data.head(5) or something, unless we want to build a utility to check the platform and call display or not

loomlike: @gramhagen Does Databricks draw pd.dataframe with data.head() or should we call display(...) explicitly?

gramhagen: no, you just see text details of the row, so actually display would look better (though you still want to define how much of the head to show)
```display(df.head(10))```

loomlike: Got it. Currently, other notebooks are not using display() so I will just follow them for now, but this is good to know.

PR #592, opened 2019-03-01
adding vowpal wabbit wrapper

gramhagen:
### Description
new wrapper for vowpal wabbit to support common interface for model instantiation, fit, and predict methods. this is a stop-gap solution until the vw python bindings (and sklearn interface) can be cleanly installed in a conda env for use in the notebooks here.
the unittests do not require vw to be installed, but for this class to work for real vw must be installed on the operating system it is running on, and on the path for the python process calling this (i.e. commands such as 'vw -f test.model -d train.dat' should work)

### Related Issues
#513
#534 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/recommender/vowpal_wabbit/__init__.py
+++ b/reco_utils/recommender/vowpal_wabbit/__init__.py
@@ -1,0 +1,8 @@
+__title__ = "Microsoft Recommenders"
+__version__ = "2019.02"
+__author__ = "RecoDev Team at Microsoft"
+__license__ = "MIT"
+__copyright__ = "Copyright 2018-present Microsoft Corporation"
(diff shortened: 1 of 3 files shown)


reco_utils/recommender/vowpal_wabbit/__init__.py
```diff
@@ -0,0 +1,8 @@
+__title__ = "Microsoft Recommenders"
```
miguelgfierro: we don't need this info here, I guess you copied this from the root folder of recommenders

>

gramhagen: oops yep

PR #596, opened 2019-03-01
Svd dd wutao

wutaomsft:
### Description

I tried to improve readability of this notebook by revising the intro.  
I re-arranged the sections so they are more consistent with other notebooks (e.g., remove a separate section on removing duplicates).  
Also, the academic bibliography style of making references is not very effective in notebooks, so I tried to make the links directly available where they are referenced.  
I added a section for additional reading if people are interested in more materials.
I am also proposing bibliography in other notebooks to follow the same style.  Will create an issue for this.

### Checklist:

- [x] I have added tests.

### DIFF
--- a/notebooks/02_model/surprise_svd_deep_dive.ipynb
+++ b/notebooks/02_model/surprise_svd_deep_dive.ipynb
@@ -734,7 +734,7 @@
    "source": [
     "## Additional Reading\n",
     "\n",
-    "[Koren, Bell and Volinksy's paper in 2009](https://ieeexplore.ieee.org/document/5197422) provides a good overview of matrix factorization methods in recommendations.  Nicolas Hug, developer of the Surprise library and the author of the first versions of this notebook, has a nice 4-part [blog post](http://nicolas-hug.com/blog/matrix_facto_1) about matrix factorization and specifically SVD."
+    "[Koren, Bell and Volinksy's paper in 2009](https://ieeexplore.ieee.org/document/5197422) provides a good overview of matrix factorization methods in recommendations.  Nicolas Hug, developer of the Surprise library and author of the first versions of this notebook, has a nice 4-part [blog post](http://nicolas-hug.com/blog/matrix_facto_1) about matrix factorization and specifically SVD."
    ]
   },
   {
--- a/notebooks/02_model/surprise_svd_deep_dive.ipynb
+++ b/notebooks/02_model/surprise_svd_deep_dive.ipynb
@@ -51,13 +51,10 @@
     "### 1.2 Stochastic Gradient Descent\n",
     "\n",
     "Stochastic Gradient Descent (SGD) is a very common algorithm for optimization where the parameters (here the biases and the factor vectors) are iteratively incremented with the negative gradients w.r.t the optimization function. The algorithm essentially performs the following steps for a given number of iterations:\n",
-    "\n",
-    "\n",
     "$$b_u \\leftarrow b_u + \\gamma (e_{ui} - \\lambda b_u)$$\n",
     "$$b_i \\leftarrow b_i + \\gamma (e_{ui} - \\lambda b_i)$$\n",
     "$$p_u \\leftarrow p_u + \\gamma (e_{ui} \\cdot q_i - \\lambda p_u)$$\n",
     "$$q_i \\leftarrow q_i + \\gamma (e_{ui} \\cdot p_u - \\lambda q_i)$$\n",
-    "\n",
     "where $\\gamma$ is the learning rate and $e_{ui} =  r_{ui} - \\hat r_{u,i} = r_{u,i} - (\\mu + b_u + b_i + q_{i}^{T}p_{u})$ is the error made by the model for the pair $(u, i)$."
    ]
   },
--- a/notebooks/02_model/surprise_svd_deep_dive.ipynb
+++ b/notebooks/02_model/surprise_svd_deep_dive.ipynb
@@ -15,20 +15,21 @@
    "source": [
     "# Surprise Singular Value Decomposition (SVD)\n",
     "\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


notebooks/02_model/surprise_svd_deep_dive.ipynb
```diff
@@ -15,20 +15,21 @@
...
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## 1 Matrix factorization algorithm\n",
+    "## 1 Singular Value Decomposition\n",
+    "\n",
+    "SVD is a *matrix factorziation* technique, which we briefly review in the [ALS deep dive notebook](als_deep_dive.ipynb).  SVD is similar to ALS (Alternating Least Squares), with the following differences:\n",
     "\n",
-    "The SVD model algorithm is very similar to the ALS algorithm presented in the ALS deep dive notebook. The two differences between the two approaches are:\n",
+    "[comment]: <> (Should we add references below?)\n",
```
miguelgfierro: I've been looking and in these notebooks: [NCF](https://github.com/Microsoft/Recommenders/blob/master/notebooks/02_model/ncf_deep_dive.ipynb), [VW](https://github.com/Microsoft/Recommenders/blob/master/notebooks/02_model/vowpal_wabbit_deep_dive.ipynb), [DKN](https://github.com/Microsoft/Recommenders/blob/master/notebooks/00_quick_start/dkn_synthetic.ipynb) and [xDeepFM](https://github.com/Microsoft/Recommenders/blob/master/notebooks/00_quick_start/xdeepfm_synthetic.ipynb) we have references at the bottom. I personally find them useful

notebooks/02_model/surprise_svd_deep_dive.ipynb
```diff
@@ -748,20 +732,25 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## References\n",
+    "## 4. To Read More...\n",
     "\n",
-    "1. Ruslan Salakhutdinov and Andriy Mnih. Probabilistic matrix factorization. 2008. URL: http://papers.nips.cc/paper/3208-probabilistic-matrix-factorization.pdf\n",
-    "2. Yehuda Koren, Robert Bell, and Chris Volinsky. Matrix factorization techniques for recommender systems. 2009.\n",
-    "3. Francesco Ricci, Lior Rokach, Bracha Shapira, and Paul B. Kantor. Recommender Systems Handbook. 1st edition, 2010."
+    "[Koren, Bell and Volinksy's paper in 2009](https://ieeexplore.ieee.org/document/5197422) provides a good overview of matrix factorization methods in recommendations.  Nicolas Hug, developer of the Surprise library and the author of the first versions of this notebook, has a nice 4-part [blog post](http://nicolas-hug.com/blog/matrix_facto_1) about matrix factorization and specifically SVD."
```
yueguoguo: In the notebooks [template](https://github.com/Microsoft/Recommenders/blob/master/notebooks/template.ipynb) we have created earlier, it is encouraged to add a reference section where a list of use learning materials mentioned in the notebook are provided to the reader. If needed, we can modify this section to be the one as shown in this notebook - in addition to the reference itself, some details about "why" it is cited is also provided. What do you (@wutaomsft @miguelgfierro) think?

wutaomsft: Yes, I think Reference section can be quite valuable.  The only thoughts I have are that 
1. In blogs, it is more popular to link the references directly  rather than point to the bibliography at the end of the blog.  In Wikipedia it is the other way around. In notebooks, the two approaches have their own pros and cons.  Direct link reduce context switching and I find it is instantly satisfying; On the other hand, if a particular paper is to be referenced multiple times, it is quite easy to use bibliography.  In this PR I used the first approach but used a very descriptive section about related papers in the hope that people who are interested in knowing more can get a little bit of context/history.
2. I think any papers appearing in References should be referenced (or discussed) at least once in the notebook; otherwise it is not clear to the reader why it is relevant.  I removed a couple of papers based on this criteria but think they could be added back if we can introduce them properly.  
Any thoughts/comments on the above?

yueguoguo: Ah I see. Agree. This makes a lot of sense!

miguelgfierro: agree

NicolasHug:
Sorry I wasn't able to provide feedback on this (holidays!), but this looks great! Thanks!

notebooks/02_model/surprise_svd_deep_dive.ipynb
```diff
@@ -748,20 +732,25 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## References\n",
+    "## Additional Reading\n",
     "\n",
-    "1. Ruslan Salakhutdinov and Andriy Mnih. Probabilistic matrix factorization. 2008. URL: http://papers.nips.cc/paper/3208-probabilistic-matrix-factorization.pdf\n",
```
anargyri: We may want to keep these since they are the papers cited in the Surprise docs. In fact (1) is closest to the SVD formulation.

wutaomsft: Yes thanks for the reference.  Will add it back once I have a little more time.

wutaomsft:
So I came back to add the reference that @anargyri suggested.  However, my changes were not there anymore.  The history showed that @anargyri made some commits right around the same time.  Was there a conflict?  My commit hash 3b46aaf2217e46257781903c2f9c2aa76738107a  @anargyri  @miguelgfierro 

Also, it would be a good discussion point what is preferred way to make references in notebooks.  I prefer not to have a"reference" section where references are separate from where they are referred.  Instead, link the reference "in place".  And then add a paragraph for "additional reading", which is more descriptive than a list of papers.  Thoughts? @yueguoguo @anargyri @miguelgfierro

miguelgfierro:
mmm there could have been a conflict, sometimes when there are several commits at the same time, information could be lost. 

I think what you said makes sense. I created an issue for tracking this https://github.com/Microsoft/Recommenders/issues/694

PR #601, opened 2019-03-04
Some tools for several algos and notebooks

miguelgfierro:
### Description

Utilities taken from PR #589 to divide the PR

Test will come in a later PR

### Checklist:

- [ ] I have added tests.


notebooks/00_quick_start/fastai_movielens.ipynb
```diff
@@ -561,7 +519,7 @@
    },
    "outputs": [],
    "source": [
-    "training_removed = pd.concat([users_items, train_valid_df[[USER,ITEM]], train_valid_df[[USER,ITEM]]]).drop_duplicates(keep=False)"
+    "training_removed = pd.concat([users_items, train_valid_df[[USER,ITEM]]]).drop_duplicates(keep=False)"
```
yueguoguo: Minor point - do we want to use the constants for column names consistently? That is, `reco_utils.common.constants.DEFAULT_USER_COL` for user column, etc.

reco_utils/common/general_utils.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import os
+
+
+def invert_dictionary(dictionary):
```
yueguoguo: I ❤️ this

reco_utils/common/gpu_utils.py
```diff
@@ -15,14 +19,76 @@ def get_number_gpus():
...
+def get_cuda_version():
+    """Get CUDA version
+    
+    Returns:
+        str: Version of the library.
+    """
+    if sys.platform == "win32":
+        raise NotImplementedError("Implement this!")
+    elif sys.platform == "linux" or sys.platform == "darwin":
+        path = "/usr/local/cuda/version.txt"
```
yueguoguo: Do we want expose this to be a function parameter in case it is not installed in the default place?

reco_utils/recommender/surprise/surprise_utils.py
```diff
@@ -0,0 +1,36 @@
...
+        trainset._inner2raw_id_users
+        if trainset._inner2raw_id_users is not None
+        else invert_dictionary(trainset._raw2inner_id_users)
+    )
+    map_item = (
+        trainset._inner2raw_id_items
+        if trainset._inner2raw_id_items is not None
+        else invert_dictionary(trainset._raw2inner_id_items)
+    )
+    df[col_user] = df[col_user].map(map_user)
```
yueguoguo: If possible, can we add some codes to make sure the returned dataframe column types are fixed, i.e., either `int` or `str` as long as the columns are prediction dataframe are matched with the ground truth one. This is for evaluation module working properly.

reco_utils/common/gpu_utils.py
```diff
@@ -15,14 +19,76 @@ def get_number_gpus():
...
+                        version += "." + line.split()[-1]
+            if version:
+                return version
+            else:
+                return "Cannot find CUDNN version"
+        else:
+            return "No CUDNN in this machine"
+
+    if sys.platform == "win32":
+        candidates = [r"C:\NVIDIA\cuda\include\cudnn.h"]
```
gramhagen: I think the default for cuda toolkit (where [cuDNN instructions](https://docs.nvidia.com/deeplearning/sdk/cudnn-install/index.html) tell you to put the header) on windows is usually:
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\vX.X\include
where X.X is the cuda toolkit version (e.g. 9.2)

reco_utils/common/timer.py
```diff
@@ -0,0 +1,51 @@
...
+        return str(timedelta(seconds=self.interval))
+
+    def start(self):
+        """Start the timer."""
+        self.init = self._timer()
+
+    def stop(self):
+        """Stop the timer. Calculate the interval in seconds."""
+        self.end = self._timer()
+        self.interval = self.end - self.init
```
gramhagen: this is nice, I wish the builtin timer api was cleaner to use like this...
might want to clear self.init here and raise an exception if stop is called before start

reco_utils/recommender/fastai/fastai_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+import numpy as np
+import pandas as pd
+import fastai
+from fastprogress import force_console_behavior
+import fastprogress
+
+
+def cartesian_product(*arrays):
```
gramhagen: why not put this in a more general location?

reco_utils/azureml/wide_deep.py
```diff
@@ -34,7 +34,7 @@
 
 def _log(metric, value):
     if run is not None:
-        if type(value) == list and len(value) > 0 and (type(value[0]) == int or type(value[0]) == float):
+        if isintance(value, list) and len(value) > 0 and isinstance(value[0], (int, float))
             run.log_list(metric, value)
         else:
             run.log(metric, str(value))
```
gramhagen: is it necessary to convert to string? the docs seem to imply that it's not (https://docs.microsoft.com/en-us/azure/machine-learning/service/how-to-track-experiments)

reco_utils/azureml/wide_deep.py
```diff
@@ -34,7 +34,7 @@
 
 def _log(metric, value):
     if run is not None:
-        if type(value) == list and len(value) > 0 and (type(value[0]) == int or type(value[0]) == float):
+        if isintance(value, list) and len(value) > 0 and isinstance(value[0], (int, float))
```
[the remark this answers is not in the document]
gramhagen: what happens if the list contains mixed types?

reco_utils/recommender/fastai/fastai_utils.py
miguelgfierro: afaik, we are not using this in any other notebook or algo, but tbh, I don't have a strong opinion on this

reco_utils/recommender/surprise/surprise_utils.py
miguelgfierro: that's a good point, I'll try to reuse your nice context function for types :-)

reco_utils/common/timer.py
miguelgfierro: good suggestion, coming in the next commit, together with tests

reco_utils/common/gpu_utils.py
miguelgfierro: yeah that makes sense

notebooks/00_quick_start/fastai_movielens.ipynb
miguelgfierro: not sure, we are not doing it in the other notebooks. I'm not sure if it is more clear to the user to explicitly provide these values or to leave them with the default values. Not strong preference here tbh

reco_utils/common/gpu_utils.py
miguelgfierro: added in the next commit

reco_utils/azureml/wide_deep.py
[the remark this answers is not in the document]
miguelgfierro: good catch
I hope we don't get mixed types right?

reco_utils/azureml/wide_deep.py
miguelgfierro: fixed

reco_utils/common/general_utils.py
miguelgfierro: coming in the next commit

reco_utils/azureml/wide_deep.py
```diff
@@ -49,6 +49,7 @@ def _log(metric, value):
 parser.add_argument('--datastore', type=str, dest='DATA_DIR', help="Datastore path")
 parser.add_argument('--train-datapath', type=str, dest='TRAIN_PICKLE_PATH')
 parser.add_argument('--test-datapath', type=str, dest='TEST_PICKLE_PATH')
+parser.add_argument('--model-dir', type=str, dest='MODEL_DIR', default='model_checkpoints')
```
anargyri: Have you checked this works when called from the AML Hyperdrive notebook?

reco_utils/recommender/fastai/fastai_utils.py
miguelgfierro: added docstrings and explanation here

reco_utils/recommender/surprise/surprise_utils.py
miguelgfierro: Added a note about the type this method returns. The [inner transformation of surprise](https://github.com/NicolasHug/Surprise/blob/master/surprise/trainset.py#L155) maps strings, so the in the output, the user and item columns are strings and the rating columns are floats.

reco_utils/azureml/wide_deep.py
miguelgfierro: I haven't but I hope it should work, why shouldn't?

loomlike: This is for the checkpoint dir clean up (not sure if aml needs this or not though).
If you want to see tensorboard, you should store the model_checkpoints under the output folder so that you can download later and open it from tensorboard.

reco_utils/common/timer.py
```diff
@@ -0,0 +1,56 @@
...
+    
+    Examples:
+        >>> big_num = 1000
+        >>> t = Timer()
+        >>> t.start()
+        >>> r = 0
+        >>> a = [r+i for i in range(big_num)]
+        >>> t.stop()
+        >>> t.interval < 1
+        True
```
loomlike: quick question. if a user calls `t.interval` without calling `t.stop()`, it will give wrong value. will that okay? few suggestions:
* reset the interval whenever `start()` get called
* make the interval as a private var `_interval` and `stop()` returns it. e.g. `interval = my_timer.stop()` this will resolve the issue of the user-mistake I mentioned.
* maybe add `lap()`? which return the current interval and reset & restart timer internally...do we need this? not sure. if we measure something in loops.. maybe.

reco_utils/recommender/fastai/fastai_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    scores = scores.sort_values([user_col, prediction_col], ascending=[True, False])
+    if top_k > 0:
+        top_scores = scores.groupby(user_col).head(top_k).reset_index(drop=True)
+    else:
+        top_scores = scores
+    return top_scores
+
+
+def hide_fastai_progress_bar():
+    """Hide fastai progress bar"""
```
loomlike: can we toggle it?

tests/unit/test_timer.py
```diff
@@ -0,0 +1,40 @@
...
+    big_num = 1000
+    t.start()
+    r = 0
+    a = [r + i for i in range(big_num)]
+    t.stop()
+    assert t.interval < 1
+    r = 0
+    with Timer() as t2:
+        a = [r + i for i in range(big_num)]
+    assert t2.interval < 1
```
loomlike: is this correct way to test? what about use time.time() or timeit's and compare the values between manual time measurement and our module with a small tolerance value?

notebooks/00_quick_start/fastai_movielens.ipynb
anargyri: My opinion is that we should use these constants everywhere in the "internal" code; but we don't want to use them in notebooks, since the outside users do not know or care about DEFAULT_USER_COL etc. And they will often want to edit the notebooks and replace these names with the ones they have. Otherwise we would parameterize everything in the notebooks and make them look like callable python functions.

loomlike:
Nice! I left few questions and suggestions.

reco_utils/recommender/fastai/fastai_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        item_col (str): Item column name.
+        prediction_col (str): Prediction column name.
+        top_k (int): Number of top items to recommend.
+
+    Returns:
+        pd.DataFrame: Result of recommendation 
+    """
+    # replace values not known to the model with #na#
+    total_users, total_items = learner.data.train_ds.x.classes.values()
+    test_df.loc[~test_df[user_col].isin(total_users), user_col] = total_users[0]
```
anargyri: Why use `total_users[0]` and not another element of the collection?

reco_utils/recommender/fastai/fastai_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    return arr.reshape(-1, la)
+
+
+def score(
+    learner,
+    test_df,
+    user_col=cc.DEFAULT_USER_COL,
+    item_col=cc.DEFAULT_ITEM_COL,
+    prediction_col=cc.DEFAULT_PREDICTION_COL,
+    top_k=0,
```
anargyri: Maybe None would be more meaningful than 0 as default value.

reco_utils/azureml/wide_deep.py
miguelgfierro: this should work, the default behavior is exactly what we had before, but now we can add a different name

tests/unit/test_timer.py
miguelgfierro: that's a good idea, I'm adding a solution with time.sleep(1).

reco_utils/recommender/fastai/fastai_utils.py
miguelgfierro: as discussed, if we need to turn it on again, we can try to find the way to do it

reco_utils/common/timer.py
miguelgfierro: good catch. I'll think about how to fix it, not sure how seting interval = my_timer.stop() will work with the contextual format, maybe it is easier to have a property and then raise an exception if the timer is not stopped. 

for your other questions, when start is called, interval is resetted. And lap is the same as interval

notebooks/00_quick_start/fastai_movielens.ipynb
miguelgfierro: yeah, that makes sense

reco_utils/recommender/fastai/fastai_utils.py
miguelgfierro: fixed

reco_utils/recommender/fastai/fastai_utils.py
miguelgfierro: changes in the next commit

reco_utils/common/timer.py
miguelgfierro: I changed this behavior, now if you call interval before stopping you get an exception.

reco_utils/common/gpu_utils.py
```diff
@@ -15,14 +22,79 @@ def get_number_gpus():
...
+            if version:
+                return version
+            else:
+                return "Cannot find CUDNN version"
+        else:
+            return "No CUDNN in this machine"
+
+    if sys.platform == "win32":
+        candidates = ["C:\\NVIDIA\\cuda\\include\\cudnn.h",
+                     "C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v[0-99]\\include\\cudnn.h"]
```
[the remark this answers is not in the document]
miguelgfierro: good catch, I'll take a look at this on Monday

Issue #604, opened 2019-03-05
[FEATURE] Pandas DataFrame to LibFFM converter


yueguoguo:
### Description
DeepRec API expects libffm formatted data which needs conversion from pandas dataframe. A utility function will make life easier.

### Expected behavior with the suggested feature
Python utility functions that help developers to convert a Pandas dataframe to libffm format data/file, that can be then processed by the algorithms in DeepRec.

### Other Comments
@yexing99 is working on utility functions of converter for movielens data.

yueguoguo:
Fixed in #605

PR #605, opened 2019-03-05
DF2LibFFM

yueguoguo:
### Description
A generic function that converts Pandas DataFrame to LibFFM format.

### Related Issues
#604

### Checklist:

- [x] I have added tests.


yueguoguo:
@yexing99 this is complementary to your ongoing work for the libffm converter. As discussed offline, personally I think it might make sense to decouple a generic converter with the data-specific wrapper (like the work you have done) for Single Responsibility reason.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
             filter_by_df.set_index(filter_by_cols).index
         )
     ]
+
+
+def df_to_libffm(df, col_rating=DEFAULT_RATING_COL, filepath=None):
+    """Converts an input Dataframe (df) to an text file in libffm format.
```
anargyri: "a text file"

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+        The above data will be converted to the libffm format by following the convention as explained in
+        https://www.csie.ntu.edu.tw/~r01922136/slides/ffm.pdf
+
+    Args:
+        df (pd.DataFrame): input Pandas dataframe.
+        col_rating (str): rating of the data.
+        filepath (str): path to save the converted data.
+
+    Return:
+        df_libffm (pd.DataFrame): a Pandas DataFrame in libffm format.
```
miguelgfierro: minor comment, the format would be:
`pd.DataFrame: a Pandas DataFrame in libffm format.`

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
         )
     ]
+
+
+def df_to_libffm(df, col_rating=DEFAULT_RATING_COL, filepath=None):
+    """Converts an input Dataframe (df) to an text file in libffm format.
+
+    Note:
+        The input dataframe is expected to represent the feature data in the following schema
+        |field-1|field-2|...|field-n|rating|
```
anargyri: I think this is too restrictive. You require that all fields have k features. Most of the time the number of features differs across fields.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+
+    # Move rating column to the first.
+    field_names.append(col_rating)
+    field_names.insert(0, field_names.pop(field_names.index(col_rating)))
+    df_new = df_new[field_names]
+
+    if filepath is not None:
+        np.savetxt(filepath, df_new.values, delimiter=' ', fmt='%s')
+
+    return df_new
```
anargyri: LIBFFM is a sparse format that is difficult to represent as a Pandas dataframe. You represent only the nonzero values, which can save a lot of space, and the representation of each data point varies in length. The closest data representation is a scipy sparse matrix, but even that would be unnecessarily cumbersome. I think we should keep it simple and write as csv or tsv file.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Proofreading should be practiced every time submitting a PR. Should we put this into the coding principles 😄

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Yea this is a good point - I am thinking whether we want to put it into some auxiliary functions that perform the data preparation if the original data is not in the expected format. What do you think?

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Hi @anargyri I am not sure if we need to represent the zero ones in the converted data. In the [paper](https://www.csie.ntu.edu.tw/~cjlin/papers/ffm.pdf), FFM has variable number of **nfk** where **n**, **f**, **k** are the number of non-zero elements, fields, latent factors, respectively.  

Also, in Section 3.3 of the paper, the first sub-section, it says
> Note that according to the number of possible values in a categorical feature, the same number of binary features are generated and every time only one of them has the value 1. In the LIBSVM format, features with zero values are not stored. We apply the same setting to all models, so in this paper, every categorical feature is transformed to several binary ones. 

The writing to file is to benefit parsing by the deeprec algorithms such as xDeepFM - it reads text file directly.

Would love to have your advice!

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+        if df_new[col].dtype == int:
+            df_new[col] = df_new[col].astype(str)
+        if df_new[col].dtype == object:
+            field_feature_concat = ['{}-{}'.format(col, x) for x in df_new[col].values]
+            field_feature_list.extend(field_feature_concat)
+    field_feature_dict = {k: v+1 for v, k in enumerate(field_feature_list)}
+
+    def _convert(field, feature, field_index_dict, field_feature_index_dict):
+        field_index = field_index_dict[field]
+        field_feature_index = (
```
gramhagen: it would be a little cleaner to handle the str type branching just once
```
if isinstance(feature, str):
    feature_idx = field_feature_map[(field, feature)]
    feature = 1
else:
    feature_idx = field_idx

return "{field_idx}:{feature_idx}:{feature}".format(field_idx=field_idx, feature_idx=feature_idx, feature=feature)
```

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+            else None
+        )
+
+        return (
+            "{}:{}:1".format(field_index, field_feature_index)
+            if isinstance(feature, str)
+            else "{}:{}:{}".format(field_index, field_index, feature)
+        )
+
+    for col in field_names:
```
gramhagen: you don't really need to create field_dict prior to this loop
```
for field_idx, col in enumerate(field_names, 1):
    ….
    df_new[col] = df_new[col].apply(lambda x: _convert(col, x, field_idx, field_feature_map))
```

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+
+    # Encode field (rating column is not included)
+    field_names = list(df_new.drop(col_rating, axis=1).columns)
+    field_dict = {k: v+1 for v, k in enumerate(field_names)}
+
+    # Encode field-feature
+    field_feature_list = []
+    for col in field_names:
+        if df_new[col].dtype == int:
+            df_new[col] = df_new[col].astype(str)
```
gramhagen: why is this converted to string? don't you want to retain handling this value numerically in _convert?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+        if df_new[col].dtype == object:
+            field_feature_concat = ['{}-{}'.format(col, x) for x in df_new[col].values]
+            field_feature_list.extend(field_feature_concat)
+    field_feature_dict = {k: v+1 for v, k in enumerate(field_feature_list)}
```
gramhagen: possible to clean this up a bit
```
idx = 1
field_feature_map = {}
for field in field_names:
    if df_new[field].dtype == object:
        for feature in df_new[field].values:
            field_feature_map[(field, feature)] = idx
            idx += 1
````

File: reco_utils/dataset/pandas_df_utils.py [hunk 9]
```diff
@@ -74,3 +75,75 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+            else "{}:{}:{}".format(field_index, field_index, feature)
+        )
+
+    for col in field_names:
+        if col is not col_rating:
+            df_new[col] = df_new[col].apply(lambda x: _convert(col, x, field_dict, field_feature_dict))
+
+    # Move rating column to the first.
+    field_names.append(col_rating)
+    field_names.insert(0, field_names.pop(field_names.index(col_rating)))
```
gramhagen: why append then pop-index-insert? can you just do this?
```
field_names.insert(0, col_rating)
```

tests/unit/test_pandas_df_utils.py
```diff
@@ -94,3 +96,40 @@ def test_filter_by():
...
+    filename = 'test'
+    filepath = os.path.join(filedir, filename)
+
+    df_feature_libffm = df_to_libffm(df_feature, col_rating='rating', filepath=filepath)
+
+    # Check if the dim is the same.
+    assert df_feature_libffm.shape == df_feature.shape
+
+    # Check if the columns are converted successfully.
+    assert df_feature_libffm.iloc[0, :].values.tolist() == [1, '1:1:1', '2:6:1', '3:3:1.0', '4:11:1']
```
gramhagen: is this the entire dataframe output? I was expecting 5 values?

reco_utils/dataset/pandas_df_utils.py
anargyri: They have more examples in the slides with varying sizes. I guess I do not understand the purpose of this function. You don't really need fields in this case, just features. To have both fields and features in the representation, you would need the user to specify the mapping of features to fields, or to identify the fields with the categorical features (how you obtain these from a pandas DF is yet another discussion).

reco_utils/dataset/pandas_df_utils.py
anargyri: What I was saying is that you should not represent the zeros; but a pandas DF will inevitably have zeros in it.
Better sync up with @Leavingseason and also take a look at the example in https://www.csie.ntu.edu.tw/~r01922136/slides/ffm.pdf
In general, LIBFFM represents cases like that in a sparse format.

reco_utils/dataset/pandas_df_utils.py
anargyri: +1
Shouldn't floating types and integers treated in the same way?

reco_utils/dataset/pandas_df_utils.py
loomlike: Even don't need to say 'a pandas dataframe' again. Maybe simply 'data in libffm format'

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Hmmm I am not sure if I understand... 'zeros' here refer to those in the 'field-feature' tuples? Are they needed explicitly in the libffm format? I can avoid using pandas DF and directly write a text file. The purpose of having a pandas DF is for the convenience of splitting with our repo utils. 

In the slides example, it shows a case that is not captured here, which is more-than-one features in an entry of a particular field. That is, in the movielens dataset, the genre of a movie is 'Comedy|Action|...'. In this case, I am not sure whether it is good to leave the multi-feature column encoding to another preparation function before the data is converted, or wrap the thing into this function... For example, for this movielens dataset, we may apply an encoding operation on the genres to encode them in a consistent way (example as you can see in the w&d notebook), and then it can be regarded as a new field with unique features in each column entry, which is the standard format expected by the function. 

What do you (@anargyri @loomlike @Leavingseason @yexing99) think?

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Good points. Fixed

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Yea this is why I think the input of the function is standardized so that the field and feature information can be derived from the Dataframe directly. I think the purpose of the function is to help map the field and feature from the information in the Dataframe to another which is represented in `field:feature:1` or `field:field:value` format.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Some times the column values can be represented in `int` type while they are actually ordinal variables or something similar.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: The reason was initially the `field_names` contain `rating` column name (now it does not), so that the pop-insert approach can move the last one to the front. Anyway, this was not an ideal one as it is also not robust (it assumes rating column is placed in the last).

Fixed!

tests/unit/test_pandas_df_utils.py
yueguoguo: Ah I just took the first row - yea update to use the entire converted data to test.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Good point - fixed it based on @gramhagen's codes. 
The older approach will lead to overwriting and thus make the index values not in a sequence.

reco_utils/dataset/pandas_df_utils.py
anargyri: I think it is impossible, in general, to write a "universal" data preprocessor or converter. There is always an ad hoc element in these things. 
So, I think it is better to narrow down what is the context in which you need / will use this util:
- what parts of the code use it?
- document exactly what it does i.e. how does it extract features and fields from the DF, how it treats different data types etc. so that the user does not need to look into the code to figure what is going on.
- rename it to something less generic; in many cases the util will not do what the data scientist wants (and that is ok because there is no easy way to write a universal util).

reco_utils/dataset/pandas_df_utils.py
anargyri: See my comments below.

reco_utils/dataset/pandas_df_utils.py
anargyri: See my comments below.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -100,45 +103,43 @@ def df_to_libffm(df, col_rating=DEFAULT_RATING_COL, filepath=None):
...
-    def _convert(field, feature, field_index_dict, field_feature_index_dict):
-        field_index = field_index_dict[field]
-        field_feature_index = (
-            field_feature_index_dict['{}-{}'.format(field, feature)]
-            if isinstance(feature, str)
-            else None
-        )
+    # Check column types.
+    types = df_new.dtypes
+    if not all([x == object or x == float or x == int for x in types]):
```
loomlike: Tested this and seems does not work with np.int64

reco_utils/dataset/pandas_df_utils.py
yexing99: xDeepFM uses the FFM format as data input: <label> <field_id>:<feature_id>:<feature_value> I guess it is the same for other deeprec algorithms.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: > So, I think it is better to narrow down what is the context in which you need / will use this util

Completely agree. This is what I amended in the most recent update - to allow only `str` and `numeric` types whilst the latter is regarded as numeric variable regardless it is int, float, or anything else. 

> what parts of the code use it? e.g. can it be used for DeepRec?

I have tested with the xDeepFM and it works well. The use case is to convert a data that is usually represented in dataframe with feature columns to be a format that is required by the APIs in DeepRec algorithms (or any package that reads a libffm format file).

> document exactly what it does i.e. how does it extract features and fields from the DF, how it treats different data types etc. so that the user does not need to look into the code to figure what is going on.

This was texted in the docstring. However, I am not sure if the conversion logic is needed as the reference is given where it is explained very clearly.

> rename it to something less generic; in many cases the util will not do what the data scientist wants (and that is ok because there is no easy way to write a universal util).

Make sense. Any suggestion here?

reco_utils/dataset/pandas_df_utils.py
yueguoguo: > xDeepFM uses the FFM format as data input: <field_id>:<feature_id>:<feature_value> I guess it is the same for other deeprec algorithms.

If **feature_value** here in the <field_id>:<feature_id>:<feature_value> I is **1** or **the actual feature value** (if feature here is the one used in the original paper), yea, this is what the function does.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Did you test it on Windows? It works on my Linux machine. But good to know as something related to dyptes also failed in the notebook, and it must have something to do with the OS. #616

loomlike: Seems in Windows, C long is not 64bit even you use 64bit os, and python int is C long, so this bug happens...  but actually this behavior gives us good insight about potential bug. 

Let's say the dataset is too large so that users want to reduce the size to save some memory, maybe by calling `pd.to_numeric(s, downcast='integer')` and the result is saved as `np.int16` type. Then we'll have bugs wherever we used a simple type comparison. same to float.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,76 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+        filepath (str): path to save the converted data.
+
+    Return:
+        pd.DataFrame: data in libffm format.
+    """
+    df_new = df.copy()
+
+    # Check column types.
+    types = df_new.dtypes
+    if not all([x == object or x == np.integer or x == np.float for x in types]):
```
loomlike: per our discussion, let's use issubdtype for more generic type check. Other than this, everything looks good!

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Good observation and indeed this is something we need to particularly careful about. @gramhagen @roalexan @miguelgfierro FYI

reco_utils/dataset/pandas_df_utils.py
anargyri: Maybe the best way would be to demonstrate by example. I.e. show an input DF, the resulting output DF and what is written to file.

yueguoguo:
Talked with @yexing99 and base the PR on the `staging` branch.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Yea this is good to have. Added.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -74,3 +75,99 @@ def filter_by(df, filter_by_df, filter_by_cols):
...
+                'field1': ['xxx1', 'xxx2', 'xxx4', 'xxx4', 'xxx4'],
+                'field2': [3, 4, 5, 6, 7],
+                'field3': [1.0, 2.0, 3.0, 4.0, 5.0],
+                'field4': ['1', '2', '3', '4', '5']
+            })
+
+        >>> df_out = libffm_converter(df_feature, col_rating='rating')
+        >>> df_out
+            rating field1 field2   field3 field4
+        0       1  1:1:1  2:2:3  3:3:1.0  4:4:1
```
anargyri: I think the feature indexes should be distinct. See slide 13 in https://www.csie.ntu.edu.tw/~r01922136/slides/ffm.pdf
You reuse however feature 2 and feature 3 for field1, field2, field3.

Issue #208, opened 2018-11-07
SAR to set remove_seen flag later


loomlike:
Would be nice if I can set remove_seen flag later too (or pass as an arg for predict / recommend).
remove_seen affects only at prediction / recommendation phase, which does filter out items when generating the results.

I wanted to test both including vs. excluding seen-items results, and had to re-train the model just for that because the current implementation we cannot change it after the model initialization.

miguelgfierro:
@loomlike is this still needed?

FYI @gramhagen

gramhagen:
I recommend keeping it in init, the latest update to sar single node removes seen items during fit (since the final score is already calculated there). and we stay consistent with als for better or worse.

anargyri:
> I recommend keeping it in init, the latest update to sar single node removes seen items during fit (since the final score is already calculated there). and we stay consistent with als for better or worse.

I wouldn't be worried so much about being consistent with ALS. It seems to go against the single responsibility principle if you specify this flag during training. I would think more about what are possible scenarios in practice. For example, maybe I want to test the same model both using recommendations that include seen items and recommendations that don't (I don't know a priori which of the two my customers will prefer).

loomlike:
+1 to @anargyri's point. Does remove_seen flag affect heavily on the training time?

Alternatively, a user can use predict() by feeding all user-item pairs as the test df, where the user can remove seen_items when preparing the df.

gramhagen:
I believe a more flexible implementation than what is done currently would only compute the user-item affinity and item-item similarity matrices as part of the fit method. The recommend method could be responsible for taking a subset of the user-item affinity for the test users and doing the matrix multiplication with item-item similarity and ranking for recommendation. At that point you could specify whether to remove seen items which could have been recorded during training or derived from the user-affinity matrix. Or as @loomlike suggests, provide the specific user-item pairs for calculation which would just be the dot product of one user vector and one item similarity column. It should also be easier to distribute the predict / recommendation calculation with this approach.

gramhagen:
this has been addressed by #617

PR #617, opened 2019-03-06
Gramhagen/sar move remove seen

gramhagen:
### Description
Improvements to SAR Single Node implementation, this does not compute the full matrix multiplication during fit, just the user affinity and item similarity matrices (as well as keeping track of items seen during training).
The computation of the scores happens in a new score method, which is leveraged by predict and recommend_k_items. In the latter case the remove_seen flag can be used to remove items seen during training (this flag is removed from the constructor)

This tries harder to ensure the datatypes of the matrices match what is used as input (for both simplicity in evaluating / joining results afterwards and maintaining numerical precision for reduced memory use)

Additionally a small simplification of the exponential decay function (and corresponding formula in the deep dive notebook) were made.

The two notebooks were also cleaned up a bit to remove unnecessary arguments in favor of defaults, and to add the title column to help demonstrate results.

### Related Issues
#208

### Checklist:

- [X] I have added tests.

### DIFF
--- a/notebooks/00_quick_start/sar_movielens.ipynb
+++ b/notebooks/00_quick_start/sar_movielens.ipynb
@@ -264,8 +264,11 @@
     "                    format='%(asctime)s %(levelname)-8s %(message)s')\n",
     "\n",
     "model = SARSingleNode(\n",
-    "    remove_seen=True, similarity_type=\"jaccard\", \n",
-    "    time_decay_coefficient=30, time_now=None, timedecay_formula=True, **header\n",
+    "    similarity_type=\"jaccard\", \n",
+    "    time_decay_coefficient=30, \n",
+    "    time_now=None, \n",
+    "    timedecay_formula=True, \n",
+    "    **header\n",
     ")"
    ]
   },
@@ -345,7 +348,7 @@
    "source": [
     "start_time = time.time()\n",
     "\n",
-    "top_k = model.recommend_k_items(test)\n",
+    "top_k = model.recommend_k_items(test, remove_seen=True)\n",
     "\n",
     "test_time = time.time() - start_time\n",
     "print(\"Took {} seconds for prediction.\".format(test_time))\n",
@@ -433,7 +436,7 @@
       ]
      },
      "metadata": {},
-     "output_type": "display_data"
+     "output_type": "execute_result"
     }
    ],
    "source": [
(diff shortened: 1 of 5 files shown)


notebooks/00_quick_start/sar_movielens.ipynb
```diff
@@ -59,9 +59,7 @@
     "import sys\n",
```
miguelgfierro: <p>oh I didn't understand how by removing the columns you would get `UserId MovieId Rating Timestamp` since they are not the default columns. I think you didn't rerun the notebook, right?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/617/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

miguelgfierro: <p>I guess removing the columns is cleaner, but I'm not sure if people will get confused when the input column names are not the default ones, maybe we should indicate it somehow, not sure if in the comments or put the columns names back. </p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/617/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

miguelgfierro: <p>yeah, here same comment as before.</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/617/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

notebooks/02_model/sar_deep_dive.ipynb
```diff
@@ -70,9 +70,9 @@
     "\n",
```
miguelgfierro: <p>this is nice</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/617/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

miguelgfierro: <p>mmm why are we getting this variability in the result? SAR is a deterministic algo, we should get the same results</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/617/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -208,97 +200,116 @@ def fit(self, df):
...
+        Args:
+            test (pd.DataFrame): user to test
+            top_k (int): number of top items to recommend
+            sort_top_k (bool): flag to sort top k results
+            remove_seen (bool): flag to remove items seen in training from recommendation
+        Returns:
+            pd.DataFrame: top k recommendation items for each user
+        """
+
+        k = min(top_k, self.n_items)
```
miguelgfierro: Maybe here we should have a warning in case num_items < k, so the user is aware that the k is modified

notebooks/00_quick_start/sar_movielens.ipynb
gramhagen: yes, my bad. fixing that

gramhagen: now the columns should be the default ones, (at least in terms of reco_utils.common.constants defaults). I felt that in the quickstart version, the less additional pieces that don't have material impact on the algorithm the better.

notebooks/02_model/sar_deep_dive.ipynb
gramhagen: this was driving me crazy, i checked the results against staging and master and they were both the same as what I'm getting now (but different than what the notebook had previously). After some digging I discovered two things
1. Including the title_col parameter perturbs the dataset due to a join of the title information
2. <i>someone</i> presumably after contemplating the Answer to the Ultimate Question of Life, the Universe, and Everything changed the default seed in python_random_split from 123 to 42.
Removing both the title and the using seed 123 I confirm I get the same results as before, however I'd prefer to just update them to the new values as they are here.

miguelgfierro: oh man sorry, it was me, yeah I changed the seed, makes sense to have the values with 42

notebooks/00_quick_start/sar_movielens.ipynb
miguelgfierro: yeah, that's reasonable.

PR #619, opened 2019-03-06
Jeremr criteo dataloader addition

miguelgfierro:
### Description

Support of Jeremy's work.

* Added general spark loader
* Added tests
* Added sample dataset

Please be aware that this PR is targeting Jeremy's branch, not staging

### Checklist:

- [ ] I have added tests.


reco_utils/dataset/criteo_dac.py
```diff
@@ -20,9 +14,18 @@
...
+
+CRITEO_URL_FULL = "https://s3-eu-west-1.amazonaws.com/kaggle-display-advertising-challenge-dataset/dac.tar.gz"
+CRITEO_URL_SAMPLE = "http://labs.criteo.com/wp-content/uploads/2015/04/dac_sample.tar.gz"
+
+
 def load_spark_df(
     spark,
-    type='train',
+    size="sample",
+    type="train",
```
miguelgfierro: hey @jreynolds01 a question here, do we really need to give the option for getting the test set if it doesn't have labels? If we can remove this, the code will be simpler

reco_utils/dataset/criteo_dac.py
```diff
@@ -110,34 +117,64 @@ def _get_schema(include_label=True):
...
-    _clean_up(local_cache_path)
+    #train_file = os.path.join(extracted_dir,'train.txt')
+    #test_file = os.path.join(extracted_dir,'test.txt')
+    
+    remove_filepath(local_cache_path)
 
     # Make sure a temporal data file get cleaned up when done
-    atexit.register(_clean_up, train_file)
-    atexit.register(_clean_up, test_file)
+    #atexit.register(remove_filepath, train_file)
```
miguelgfierro: do we need this?

reco_utils/dataset/criteo_dac.py
```diff
@@ -110,34 +117,64 @@ def _get_schema(include_label=True):
...
+
+
+def _load_datafile(size, local_cache_path="dac.tar.gz", dbutils=None):
     """ Download and extract file """
 
     path, filename = os.path.split(os.path.realpath(local_cache_path))
 
     # Make sure the temporal tar file gets cleaned up no matter what
-    atexit.register(_clean_up, local_cache_path)
+    atexit.register(remove_filepath, local_cache_path)
```
miguelgfierro: I'm thinking that we should rethink of having this optional, in both criteo and movielens. Right now we are using maybe_download, but we are always erasing the data

reco_utils/dataset/criteo_dac.py
```diff
@@ -110,34 +117,64 @@ def _get_schema(include_label=True):
...
+
+  """
+  url = {
+    "full": CRITEO_URL_FULL,
+    "sample": CRITEO_URL_SAMPLE
+  }
+  return url[size]
+
+
+def download_criteo(size="full", filename="dac.tar.gz", work_directory="."):
```
miguelgfierro: FYI @loomlike, please take a look at the way we are splitting the code to follow single responsibility. I think this can be also done in movielens.py

reco_utils/dataset/url_utils.py
```diff
@@ -53,3 +53,15 @@ def maybe_download(url, filename, work_directory=".", expected_bytes=None):
 
     return filepath
 
+
```
miguelgfierro: not sure if we should rename this file from url_utils to download_utils... any preference here?

reco_utils/dataset/criteo_dac.py
loomlike: good point. can we utilize tmp folder so that even the user forget to remove later, os automatically cleans up?

reco_utils/dataset/criteo_dac.py
loomlike: I think this could be a good discussion point... currently to use the identical train and test split to other notebooks (which is needed for benchmarking/comparison), we use the same seed for splitter. If we support that from dataloader level, users don't need to worry about checking seed from the other notebooks.

Some datasets may have pre-split train and test sets in their files, but not sure if we want to give that back though.

miguelgfierro: the issue with this test set is that we don't have the labels (correct me Jeremy if I'm wrong), so not sure in which situation we will use this in our repo

reco_utils/dataset/url_utils.py
loomlike: maybe better name to include 'upload', etc. network related utils together? or will that be too broad?

Btw, can we move tqdm to other utils? it is general function to show any progress bars, right? we can even use it for showing model training or evaluation status, not just for downloading status.

reco_utils/dataset/criteo_dac.py
```diff
@@ -110,34 +117,64 @@ def _get_schema(include_label=True):
   return schema
 
 
-def _load_datafile(local_cache_path="dac.tar.gz", dbutils=None):
+def criteo_urls(size):
```
gramhagen: i would drop this method in favor of storing the dict at the top of the file

reco_utils/dataset/criteo_dac.py
```diff
@@ -110,34 +117,64 @@ def _get_schema(include_label=True):
...
-    atexit.register(_clean_up, local_cache_path)
+    atexit.register(remove_filepath, local_cache_path)
 
     ## download if it doesn't already exist locally
-    maybe_download(
-          "https://s3-eu-west-1.amazonaws.com/kaggle-display-advertising-challenge-dataset/dac.tar.gz",
-          filename,
-          work_directory=path,
-    )
+    download_criteo(size, filename, path)
```
gramhagen: why is there an intermediate method here?

reco_utils/dataset/criteo_dac.py
miguelgfierro: ok, I'll assume for now that we are not using the test part for simplicity. In the future, we can change this if needed

reco_utils/dataset/criteo_dac.py
miguelgfierro: done in the next commit

loomlike:
@miguelgfierro I guess you merged staging under detached head status or something... since this PR now has 201 commits and 58 file changed

reco_utils/dataset/criteo_dac.py
miguelgfierro: in some cases, people would just want to download criteo, there was a person asking for this method in movielens as well

Issue #616, opened 2019-03-05
[BUG] Error running test notebook surprise_svd_deep_dive on Windows DSVM


roalexan:
### Description
Ran this test:
pytest tests/unit/test_notebooks_python.py -k test_surprise_deep_dive_runs

See this error output:

E           papermill.exceptions.PapermillExecutionError:
E           ---------------------------------------------------------------------------
E           Exception encountered at "In [10]":
E           ---------------------------------------------------------------------------
E           TypeError                                 Traceback (most recent call last)
E           <ipython-input-10-c83530820ecb> in <module>
E           ----> 1 eval_rmse = rmse(test, predictions)
E                 2 eval_mae = mae(test, predictions)
E                 3 eval_rsquared = rsquared(test, predictions)
E                 4 eval_exp_var = exp_var(test, predictions)
E                 5
E
E           ~\notebooks\Recommenders\reco_utils\evaluation\python_evaluation.py in check_column_dtypes_wrapper(rating_true, rating_pred, col_user, col_item, col_rating, col_prediction, *args, **kwargs)
E                57             raise TypeError(
E                58                 "data types of column {} are different in true and prediction".format(
E           ---> 59                     col_user
E                60                 )
E                61             )
E
E           TypeError: data types of column userID are different in true and prediction

### In which platform does it happen?
Tested on a Windows CPU DSVM

### How do we replicate the issue?
I created a Windows DSVM (Standard DS3 v2 - 4 vcpus, 14 GB memory). Don't know if this issue is specific to Windows (probably not).
Windows setup steps were (for now):

1. login to DSVM
2. cd notebooks
3. git clone https://github.com/Microsoft/Recommenders
4. git checkout staging
5. conda update conda
6. conda update --all
7. cd Recommenders
8. python scripts/generate_conda_file.py
9. conda env create -f reco_base.yaml
10. conda info --envs (list environments, verify "reco_base" is there)
11. conda activate reco_base
12. pytest tests/unit/test_notebooks_python.py -k test_surprise_deep_dive_runs

### Expected behavior (i.e. solution)
The test should run successfully

yueguoguo:
Hmmm it runs well on my environment. The issue was fixed in #553. 

Step 4 should happen after Step 7 as after clone the current dir is not the cloned repository. Not sure if this is the cause but have you tried `git branch` and `git status` to make sure it is indeed `staging` branch?

roalexan:
[...]

PR #620, opened 2019-03-06
updating notebook to explicitly use int64 datatype for prediction

gramhagen:
### Description
explicitly convert userID and itemID columns to int64 (vs int) in surprise svd notebook

### Related Issues
#616 

### Checklist:

- [X] I have added tests.

### DIFF
--- a/notebooks/02_model/surprise_svd_deep_dive.ipynb
+++ b/notebooks/02_model/surprise_svd_deep_dive.ipynb
@@ -431,7 +431,7 @@
     "                                                     'est': 'prediction'})\n",
     "predictions = predictions.drop(['details', 'r_ui'], axis='columns')\n",
     "for col in ['userID', 'itemID']:\n",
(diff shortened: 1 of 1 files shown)


yueguoguo:
This is a general question about how to well handle the dtype in Python/Pandas/Numpy. It looks like using `np.integer` is a good practice. See this https://stackoverflow.com/questions/22471644/how-do-i-check-if-a-numpy-dtype-is-integral/22476290.

gramhagen:
@yueguoguo np.integer is parent dtype to lots of different dtypes, in this case we need to be pretty specific about which dtype to use, though if you're saying later if we want to check type consistency we can use np.issubdtype and np.integer I agree.

here I used the string 'int64' vs. np.int64 to avoid adding an additional import of numpy, this is converted to np.int64 inside pandas [astype](https://github.com/pandas-dev/pandas/blob/e3b095074c5c3b29ec25a994cd204349b872d961/pandas/core/dtypes/common.py#L2037)

Issue #629, opened 2019-03-11
[FEATURE] Negative sampling utility


yueguoguo:
### Description
To generate negative label. Negative sampling is commonly used in literate to generate negative implicit feedback. There is an ad-hoc implementation in the NCF codes of the repo, and it would be good to have it a generic function.

### Expected behavior with the suggested feature
To generate negative feedback in the dataset to favor certain deep learning algorithms for implicit feedback scenario

yueguoguo:
Resolved in #637

PR #637, opened 2019-03-12
Negative sampling utility

yueguoguo:
### Description
Pandas DataFrame utility function that does negative sample on a given input dataframe of user-item interactions.

### Related Issues
#629 

### Checklist:

- [x] I have added tests.


reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,93 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+    https://www.comp.nus.edu.sg/~xiangnan/papers/ncf.pdf
+    
+    Examples:
+        >>> import pandas as pd
+        >>> df = pd.DataFrame({
+            'userID': [1, 2, 3],
+            'itemID': [1, 2, 3],
+            'rating': [5, 5, 5]
+        })
+        >>> df_neg_sampled = negative_sampler_from_user_item_tuple(
```
anargyri: "negative_sampler_from_user_item_tuple" is obsolete?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,93 @@ def _convert(field, feature, field_index, field_feature_index_dict):
     return df_new
 
 
+def negative_feedback_sampler(
```
anargyri: Maybe it also makes sense to have a threshold option. I.e. if threshold == 2 any ratings below 2 will be considered negative; if you have 100 ratings >= 2, 70 ratings < 2 then you can fill in the remaining 30 negatives with sampling.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,93 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+        col_user (str): user id column name.
+        col_item (str): item id column name.
+        number_neg_per_pos (int): number of negative feedback sampled for each positive feedback on the user-item interactive data. 
+        seed (int): seed for the random state of the sampling function.
+
+    Returns:
+        pandas.DataFrame: data with negative feedback 
+    """
+    columns = df.columns
+    if col_user not in columns:
```
anargyri: How about using decorators as in https://github.com/Microsoft/Recommenders/blob/master/reco_utils/evaluation/python_evaluation.py

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,93 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+    users, items = df[col_user].unique(), df[col_item].unique()
+
+    # Create a dataframe for all user-item pairs
+    user_item_tuples = [(row[col_user], row[col_item]) for _, row in df.iterrows()]
+
+    # Generate feedback for both positive and negative
+    user_item_feedback = [
+        (user, item, 1 if (user, item) in user_item_tuples else 0) for user in users for item in items
+    ]
+    df_all = pd.DataFrame(user_item_feedback, columns=[col_user, col_item, 'rating'])
```
anargyri: Could you replace 'rating' with `col_rating` as an argument?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,93 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+    """
+    columns = df.columns
+    if col_user not in columns:
+        raise ValueError("Column {} is not in the dataframe".format(col_user))
+
+    if col_item not in columns:
+        raise ValueError("Column {} is not in the dataframe".format(col_item))
+
+    # Get all of the users and items.
+    users, items = df[col_user].unique(), df[col_item].unique()
```
anargyri: Better rewrite this in two lines.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,93 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+
+    # Create a dataframe for all user-item pairs
+    user_item_tuples = [(row[col_user], row[col_item]) for _, row in df.iterrows()]
+
+    # Generate feedback for both positive and negative
+    user_item_feedback = [
+        (user, item, 1 if (user, item) in user_item_tuples else 0) for user in users for item in items
```
anargyri: Why do you need the Cartesian product of users and items? You will only sample `number_neg_per_pos * sample_size` negative samples.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,93 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+
+    # Sample # negative feedback for each user
+    # If the total negative feedback for a user is less than the sampling size, all of the 
+    # negative feedback will be generated.
+    df_neg = df_all[df_all['rating'] == 0]
+    df_neg_sample = (
+        df_neg
+        .groupby(col_user)
+        .apply(lambda x: x.sample(
+            n=min(number_neg_per_pos, len(x)),
```
anargyri: This will not sample `number_neg_per_pos * sample_size` negatives (according to the docstring). It will sample based on the number of samples per user.

miguelgfierro:
@yexing99 not sure if you have already connected with Le, but this PR is super similar to what you and I were talking earlier

yueguoguo:
> @yexing99 not sure if you have already connected with Le, but this PR is super similar to what you and I were talking earlier

Yea I noticed that @yexing99 created a feature request issue in #644 @yexing99 is the code in this PR something you are looking for? I would love to understand better about your ask here.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Yea plan to add these (maybe in another PR). We have these in PySpark implementations.

yueguoguo: Good point. Do we want to have another function for this? Another thing is that the negative sampler in this PR does not assume an existence of rating column, meaning that, the negative samples (to a user) are defined as those items that have not been interacted with the user before.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Good idea!

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Good catch.

yueguoguo: Ah no, `col_rating` is not a mandatory column in the input data.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Right. Need to update the docstring. The `min` operator is to make sure there are sufficient number of rows to sample for a user.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: This is to create a dataframe for all the possible interactions between the users and items in the input dataset. Then the existing interactions between users and items in the input data wil be labelled with 1 (positive feedback), whilst the rest will be labelled with 0 (negative feedback). Then for each user, all of the positive feedback and a number of negative feedback (determined by the input `number_neg_per_user`. **NOTE** this is updated, it used to be `number_neg_per_pos` which is not correct).

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Right. This was updated.

reco_utils/dataset/pandas_df_utils.py
anargyri: Right, better leave this for another function. I guess there is some common code to be reused.

reco_utils/dataset/pandas_df_utils.py
anargyri: But it is in the output DataFrame.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +166,94 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+        2   2   1
+        2   1   0
+        3   3   1
+        3   1   0
+
+    Args:
+        df (pandas.DataFrame): input data that contains user-item tuples.
+        col_user (str): user id column name.
+        col_item (str): item id column name.
+        number_neg_per_user (int): number of negative feedback sampled for each positive feedback on the user-item interactive data.
```
anargyri: This description is not  accurate; it should be "for each user".

tests/unit/test_pandas_df_utils.py
```diff
@@ -140,5 +141,50 @@ def test_csv_to_libffm():
         assert line == '1 1:1:1 2:2:3 3:3:1.0 4:4:1\n'
 
 
+def test_negative_feedback_sampler():
+    df = pd.DataFrame({
+        'userID': [1, 2, 3, 4, 4, 5, 5, 5],
+        'itemID': [1, 2, 3, 1, 2, 1, 2, 3],
+        'rating': [5, 5, 5, 5, 5, 5, 5, 5]
```
anargyri: It may be a good idea to add a test without the 'rating' column, to ensure the functionality in both cases.

reco_utils/dataset/pandas_df_utils.py
anargyri: I suspect you could do the sampling in another way so as not to expand the full Cartesian product.
Instead of line 243 below, where you sample from the negative rows of the Cartesian product, you could sample from the items (per user) that do not appear in `df`.
The problem now is that you need O(n^2) memory and CPU. But you could do it in  O(`df.shape[0] + number_neg_per_user * n_users`). We can think about what is the best method available to do this. One way, for example, could be, for each user, to sample `number_neg_per_user` + the number of items in `df`, and then take the setdiff from the items already in `df`.

yexing99:
@yueguoguo @miguelgfierro  yes, it is related. For implicit feedback, having a negative sampler is needed if we want to formulate it into a two class classification problem. It would cause issues if # of negative sample  is far bigger than positive sample. 
To solve real problem nicely, we'd change to RBM format and loss function.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: > One way, for example, could be, for each user, to sample number_neg_per_user + the number of items in df, and then take the setdiff from the items already in df.

However, in this case, we don't know which ones are the negative samples, right? And if setdiff does not give us the exact number of negative samples which is pre-determined by `number_neg_per_user`, the operation needs to be done for another time. 

Alternative may be when randomly sample the negative samples, it can be checked against the positive items, and then negative items. This can reduce the complexity to O(df.shape[0] + number_neg_per_user * n_users)

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Updated.

tests/unit/test_pandas_df_utils.py
yueguoguo: Done

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Added.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +167,97 @@ def _convert(field, feature, field_index, field_feature_index_dict):
     return df_new
 
 
+def negative_feedback_sampler(
+    df, 
+    col_user=DEFAULT_USER_COL,
+    col_item=DEFAULT_ITEM_COL,
+    col_label=DEFAULT_LABEL_COL,
+    number_neg_per_user=1,
```
loomlike: Are there any cases we want to have imbalanced labels? If not, why don't we just remove `number_neg_per_user` argument and make this function to (try to) sample the same number of negative examples as positive's?

If we still want to have this option, wouldn't be better to pass the ratio to positive samples? Similar to `pd.DataFrame.sample`'s fraction argument.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +167,97 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+    Returns:
+        pandas.DataFrame: data with negative feedback 
+    """
+    columns = df.columns
+    if col_user not in columns:
+        raise ValueError("Column {} is not in the dataframe".format(col_user))
+
+    if col_item not in columns:
+        raise ValueError("Column {} is not in the dataframe".format(col_item))
+
```
loomlike: These two errors will be raised by the below codes anyway.
```
    users = df[col_user].unique()
    items = df[col_item].unique()
```

reco_utils/dataset/pandas_df_utils.py
loomlike: For this, can you reuse `user_item_pairs()` in the same module?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +167,97 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+    ]
+    df_all = pd.DataFrame(user_item_feedback, columns=[col_user, col_item, col_label])
+
+    # Take all positive feedback
+    df_pos = df_all[df_all[col_label] == 1]
+
+    # Sample # negative feedback for each user
+    # If the total negative feedback for a user is less than the sampling size, all of the 
+    # negative feedback will be generated.
+    df_neg = df_all[df_all[col_label] == 0]
```
loomlike: Is this different from "remove-seen" or can you regard this as that? If this is something similar, you may also reuse `filter_by()` or use `user_item_pairs(user_item_filter_df=your_positive_samples)` which will return all the user-item pairs except your positive samples.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -166,3 +167,97 @@ def _convert(field, feature, field_index, field_feature_index_dict):
...
+        .groupby(col_user)
+        .apply(lambda x: x.sample(
+            n=min(number_neg_per_user, len(x)),
+            random_state=seed, 
+            replace=False))
+        .reset_index(drop=True)
+    )
+
+    # Concatenate the positive and negative feedback dataframes
+    df_sampled = (
```
loomlike: Correct me if I'm thinking wrong...but I have feeling that we maybe able to reuse stratified sampling function or chrono-splitter for this task... e.g. if you set 1 for positive and 0 as negative and regards them as 'timestamp' and run w/ chrono-splitter.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Good point. `ratio_neg_per_user` sounds a good idea.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: 👍

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Updated.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Hmmm this is really an interesting and great idea. I am not sure if it can be nicely coped with the existing splitter method - as you said random sampling may be an issue and sample the negative feedback items w.r.t the positive samples can also be something tricky. In that case, perhaps reusing the splitter may introduce even more overhead to the code 😅

yueguoguo: @loomlike I will merge now and then if there is anything to refactor I'll revisit the code.

PR #639, opened 2019-03-12
Setup.py

dciborow:
### Checklist:

- [ ] I have added tests.


setup.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+
+from distutils.core import setup
+
+setup(name='RecoUtils',
+      version='2019.02',
```
miguelgfierro: version, author and maybe description could be taken programmatically from here: https://github.com/Microsoft/Recommenders/blob/master/reco_utils/__init__.py#L8

I think it would be something like:
```
from reco_utils import VERSION

setup(name='RecoUtils',
      version=VERSION,
     ...
)
same for the others

```

setup.py
```diff
@@ -0,0 +1,10 @@
...
+      description='Reco Utils',
+      author='Microsoft',
+      packages=['reco_utils','reco_utils.common','reco_utils.dataset','reco_utils.evaluation','reco_utils.recommender'],
```
gramhagen: not including azureml?

gramhagen:
can you comment on what functionality this supports? would be good to have a record in case others search for something that this solves

yueguoguo:
@gjwgit @simonzhaoms FYI the mlhub model loading feature may be related to the work by @dciborow in this PR.

gramhagen:
pulling this in is pending a decision to allow this repo to be more package-like...

nikhilrj:
@gramhagen sounds like we have the go ahead to do this now - @wutaomsft can you confirm? :)

setup.py
dciborow: Done!

setup.py
dciborow: this should only be the list of packages within this project, and not dependencies.

yueguoguo: This is just to create a package (distribution) from github repo, right? The ask is to make sure our purpose is just to distribute a package of the repo onto PyPi for easy downloading, or to ease the installation as well (with the help of `distutils`. If there is just a zipped package distribution for downloading, after that, a user is expected to run `python setup install` for a particular environment probably an option parameter should be specified.

We may want to have advice from @gjwgit and @simonzhaoms as they have a nice setup script in mlhub which may be helpful in our case. See [here](https://github.com/mlhubber/mlhub/blob/master/setup.py).

gramhagen:
My comment was that we now have additional folders under reco_utils that need to be included.

The setup tools example is nice too. I do think we need to constrain this to be run in an existing environment, otherwise the dependencies get tricky due to different options for gpu, spark, and base.

miguelgfierro:
a nice template I found for setup.py with setuptools: https://github.com/pypa/sampleproject/blob/master/setup.py

gramhagen:
moving work to #851

Issue #631, opened 2019-03-12
[BUG] FastAI quickstart notebook failure


yueguoguo:
### Description
It looks like the failure of fastai is owing to the train-valid split in model building. 

Cell 17 in the [notebook](https://github.com/Microsoft/Recommenders/blob/master/notebooks/00_quick_start/fastai_movielens.ipynb) creates a new test dataset that removed the seem items. 
```
training_removed = pd.concat([users_items, train_valid_df[[USER,ITEM]]]).drop_duplicates(keep=False)
```
However, `train_valid_df` also includes the user-item pairs in the validation set (not sure if there is a way to find out the user-item pairs used in the validation dataset in the model building). These items are not seen in the training set for building the model.

### In which platform does it happen?
Azure Data Science Virtual Machine

### How do we replicate the issue?
Clone the notebooks and run the fast ai [notebook](https://github.com/Microsoft/Recommenders/blob/master/notebooks/00_quick_start/fastai_movielens.ipynb)

### Expected behavior (i.e. solution)
It should run successfully

miguelgfierro:
@gramhagen and I have been looking into this today. We found that in this old commit, things were working https://github.com/Microsoft/Recommenders/blob/benchmark_feb/notebooks/03_evaluate/comparison_movielens.ipynb

gramhagen:
addressed in #646

PR #646, opened 2019-03-14
Gramhagen/fix fastai

gramhagen:
### Description
Updates to FastAI notebook to avoid internal train/validation split causing items to not appear in training set. Also using stratified split for the same reason. Lastly, adjusting the top_k scoring to properly remove training data and generate scores.

### Related Issues
#631

### Checklist:

- [X] I have added tests.

### DIFF
--- a/notebooks/00_quick_start/fastai_movielens.ipynb
+++ b/notebooks/00_quick_start/fastai_movielens.ipynb
@@ -20,7 +20,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


gramhagen:
If I remove it now, won't it break the test? Or I have to skip the test which also seems bad.

I'd prefer to leave it there so when we create the PR to address that issue we can search and replace all papermill record instances and change them to use scrapbook glue.

Issue #645, opened 2019-03-12
Movielens loader to load movie release year


loomlike:
### Description
Load movie release year so that can be used as an item feature

### In which platform does it happen?
Both pandas and spark data loader

### Expected behavior (i.e. solution)
```
item_df = movielens.load_item_df(
    ...
    movie_col='MovieId',
    title_col='Title',
    genres_col='Genres',
    year_col='Year'
)
```
will return pd.DataFrame of

| MovieId | Title | Genres | Year |
| --- | --- | --- | --- |
| 1 | Toy Story (1995) | Animation\|Children's\|Comedy | 1995 |
| ... | ... | ... | ... |

### Other Comments
Will create PR along with the changes for #622

loomlike:
Merged into Staging

Issue #622, opened 2019-03-07
refactor movielens dataloader for single responsibility


loomlike:
### Description
movielens dataloader needs refactor for single responsibility and better code-readability

### Expected behavior (i.e. solution)
change `_load_datafile` to `maybe_download_and_extract`
which will have `_maybe_download` and `_extract_rating_and_item_data`

loomlike:
Merged into Staging

PR #647, opened 2019-03-14
Movielens load movie's release year

loomlike:
### Description
Movie's release year can be used as an item feature along with rating timestamp.

Usage:
```
df = movielens.load_pandas_df(
    size="1m",
    header=("UserId", "MovieId", "Rating", "Timestamp"),
    title_col="Title",
    genres_col="Genres",
    year_col="Year"
)
```

Other changes include:
 - Refactor submodules, especially local cache path and cleanup
   (now download-and-extract only when ratings and movie data files
   do not exist in the local cache path)
 - Add download_movielens and load_item_df functions
 - Trim down redundant smoke and integration tests

### Related Issues
#645 #622 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/dataset/movielens.py
+++ b/reco_utils/dataset/movielens.py
@@ -1,9 +1,11 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+from contextlib import contextmanager
 import os
 import re
 import shutil
+from tempfile import TemporaryDirectory
 import warnings
 import pandas as pd
 from zipfile import ZipFile
@@ -111,7 +113,7 @@ DATA_FORMAT = {
     ),
 }
 
-# 100K data genres index to string
+# 100K data genres index to string mapper. For 1m, 10m, and 20m, the genres labels are already in the dataset.
 GENRES = (
     "unknown", "Action", "Adventure", "Animation",
     "Children's", "Comedy", "Crime", "Documentary", "Drama", "Fantasy",
@@ -133,8 +135,6 @@ WARNING_MOVIE_LENS_HEADER = """MovieLens rating dataset has four columns
 WARNING_HAVE_SCHEMA_AND_HEADER = """Both schema and header are provided.
     The header argument will be ignored."""
 ERROR_MOVIE_LENS_SIZE = "Invalid data size. Should be one of {100k, 1m, 10m, or 20m}"
-ERROR_LOCAL_CACHE_PATH = """Local cache path only accepts a zip file path:
-    use/something/like_this.zip"""
 ERROR_NO_HEADER = "No header (schema) information"
 
 
@@ -153,8 +153,8 @@ def load_pandas_df(
     Args:
         size (str): Size of the data to load. One of ("100k", "1m", "10m", "20m").
         header (list or tuple or None): Rating dataset header.
-        local_cache_path (str): Path where to cache the zip file locally.
-            If None, use default, which is "data/ml-{size}/ml.zip"
+        local_cache_path (str): Path (directory or a zip file) to cache the downloaded zip file.
+            If None, all the intermediate files will be stored in a temporary directory and removed after use.
         title_col (str): Movie title column name. If None, the column will not be loaded.
         genres_col (str): Genres column name. Genres are '|' separated string.
             If None, the column will not be loaded.
@@ -163,27 +163,25 @@ def load_pandas_df(
     Returns:
         pd.DataFrame: Movie rating dataset.
     """
-    movie_col = DEFAULT_ITEM_COL if header is None or len(header) < 2 else header[1]
-
-    # fix capitalization
     size = size.lower()
-
-    datapath, _ = _maybe_download_and_extract(size, local_cache_path)
-
-    # Load movie features such as title, genres, and release year
-    if title_col is None and genres_col is None and year_col is None:
-        item_df = None
-    else:
-        item_df = load_item_df(size, local_cache_path, movie_col, title_col, genres_col, year_col)
-
-    # Load rating data
+    if size not in DATA_FORMAT:
+        raise ValueError(ERROR_MOVIE_LENS_SIZE)
     if header is None or len(header) == 0:
         raise ValueError(ERROR_NO_HEADER)
-    else:
-        if len(header) > 4:
-            warnings.warn(WARNING_MOVIE_LENS_HEADER)
-            header = header[:4]
-        elif len(header) == 1 and item_df is not None:
+
+    if len(header) > 4:
+        warnings.warn(WARNING_MOVIE_LENS_HEADER)
+        header = header[:4]
+    movie_col = DEFAULT_ITEM_COL if len(header) < 2 else header[1]
+
+    with _real_path(local_cache_path, "ml-{}.zip".format(size)) as path:
+        datapath, item_datapath = _maybe_download_and_extract(size, path)
+
+        # Load movie features such as title, genres, and release year
+        item_df = _load_item_df(size, item_datapath, movie_col, title_col, genres_col, year_col)
+
+        # Load rating data
+        if len(header) == 1 and item_df is not None:
             # MovieID should be loaded to merge rating df w/ item_df
             header = [header[0], movie_col]
 
@@ -204,7 +202,7 @@ def load_pandas_df(
         if item_df is not None:
             df = df.merge(item_df, on=header[1])
 
-        return df
+    return df
 
 
 def load_item_df(
@@ -215,12 +213,12 @@ def load_item_df(
     genres_col=None,
     year_col=None
 ):
-    """Loads Movie data.
+    """Loads Movie info.
 
     Args:
         size (str): Size of the data to load. One of ("100k", "1m", "10m", "20m").
-        local_cache_path (str): Path where to cache the zip file locally.
-            If None, use default, which is "data/ml-{size}/ml.zip"
+        local_cache_path (str): Path (directory or a zip file) to cache the downloaded zip file.
+            If None, all the intermediate files will be stored in a temporary directory and removed after use.
         movie_col (str): Movie id column name.
         title_col (str): Movie title column name. If None, the column will not be loaded.
         genres_col (str): Genres column name. Genres are '|' separated string.
@@ -228,18 +226,31 @@ def load_item_df(
         year_col (str): Movie release year column name. If None, the column will not be loaded.
 
     Returns:
-        pd.DataFrame: Movie information data, such as title, genres, and release year
+        pd.DataFrame: Movie information data, such as title, genres, and release year.
     """
-    if movie_col is None:
-        return None
-
-    # fix capitalization
     size = size.lower()
+    if size not in DATA_FORMAT:
+        raise ValueError(ERROR_MOVIE_LENS_SIZE)
 
-    datapath, item_datapath = _maybe_download_and_extract(size, local_cache_path)
+    with _real_path(local_cache_path, "ml-{}.zip".format(size)) as path:
+        _, item_datapath = _maybe_download_and_extract(size, path)
+        item_df = _load_item_df(size, item_datapath, movie_col, title_col, genres_col, year_col)
 
-    item_header = [movie_col, "title_year"]
-    usecols = [0, 1]  # id and title columns
+    return item_df
+
+
+def _load_item_df(size, item_datapath, movie_col, title_col, genres_col, year_col):
+    """Loads Movie info"""
+    item_header = []
+    usecols = []
+    if movie_col is not None:
+        item_header.append(movie_col)
+        usecols.append(0)
+
+    # Year is parsed from title
+    if title_col is not None or year_col is not None:
+        item_header.append("title_year")
+        usecols.append(1)
 
     genres_header_100k = None
     if genres_col is not None:
@@ -252,6 +263,9 @@ def load_item_df(
         else:
             item_header.append(genres_col)
             usecols.append(2)  # genres column
+
+    if len(item_header) == 0:
+        return None
 
     item_df = pd.read_csv(
         item_datapath,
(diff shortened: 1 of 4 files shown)


reco_utils/dataset/movielens.py
```diff
@@ -128,50 +135,50 @@ def item_has_header(self):
...
-    movie_col = DEFAULT_ITEM_COL if header is None or len(header) < 2 else header[1]
-    item_df = _load_item_df(size, movie_col, title_col, genres_col, item_datapath)
+    # Load movie features such as title, genres, and release year
+    if title_col is None and genres_col is None and year_col is None:
+        item_df = None
+    else:
+        item_df = load_item_df(size, local_cache_path, movie_col, title_col, genres_col, year_col)
 
     # Load rating data
     if header is None or len(header) == 0:
```
gramhagen: can we move this check to the beginning, no point in downloading anything if the header is invalid

reco_utils/dataset/movielens.py
```diff
@@ -128,50 +135,50 @@ def item_has_header(self):
...
-            If None, the genres column is not loaded.
+            If None, the column will not be loaded.
+        year_col (str): Movie release year column name. If None, the column will not be loaded.
 
     Returns:
         pd.DataFrame: Movie rating dataset.
-            If header is None but either title_col or genres_col is not None,
-            returns movie titles and/or genres.
     """
+    movie_col = DEFAULT_ITEM_COL if header is None or len(header) < 2 else header[1]
```
gramhagen: header should never be None?

reco_utils/dataset/movielens.py
```diff
@@ -377,48 +424,60 @@ def _get_schema(header, schema):
...
+    maybe_download(url, dest_filename, work_directory=dest_path)
 
 
-def _clean_up(filepath):
-    """ Remove cached file. Be careful not to erase anything else. """
-    try:
-        os.remove(filepath)
-    except OSError:
-        pass
+def _extract(size, local_cache_path, datapath, item_datapath):
```
gramhagen: this is pretty specialized for movielens right? I don't think we need to separate it out as its own function

reco_utils/dataset/movielens.py
```diff
@@ -377,48 +424,60 @@ def _get_schema(header, schema):
...
 
-    _clean_up(local_cache_path)
+    return datapath, item_datapath
 
-    # Make sure a temporal data file get cleaned up when done
-    atexit.register(_clean_up, datapath)
-    atexit.register(_clean_up, item_datapath)
 
-    return datapath, item_datapath
+def download_movielens(size, dest_filename, dest_path="."):
```
gramhagen: i don't think we gain anything by encapsulating this as a wrapper function

reco_utils/dataset/movielens.py
```diff
@@ -377,48 +424,60 @@ def _get_schema(header, schema):
...
     if size not in DATA_FORMAT:
         raise ValueError(ERROR_MOVIE_LENS_SIZE)
-    if not local_cache_path.endswith(".zip"):
+    if local_cache_path is None:
+        default_dir = os.path.join("data", "ml-{}".format(size))
+        local_cache_path = os.path.join(default_dir, "ml.zip")
+    elif not local_cache_path.endswith(".zip"):
         raise ValueError(ERROR_LOCAL_CACHE_PATH)
 
     path, filename = os.path.split(os.path.realpath(local_cache_path))
```
gramhagen: I would suggest simplifying this section by leveraging tempfile.TemporaryDirectory and registering it's cleanup at process close. Also, I think it will be easier to cache the zip file only, decompression if fast, and that way you only have to keep track of a single file using local_cache_path.

```
from tempfile import TemporaryDirectory

tmp_dir = TemporaryDirectory()
tmp_path = tmp_dir.name

if local_cache_path is None:
  path = tmp_dir.name
  filename = "data.zip"
  atexit.register(tmp_dir.cleanup)
  ... download file to tmp dir...

... extract file to tmp dir ...
... load files from tmp dir ...
```

tests/integration/test_movielens.py
```diff
@@ -20,113 +22,147 @@
...
+    assert year[0][0] == year[1][0]
+    assert year[0][0] == year_example
 
-    # Test if use schema when both schema and header are provided
-    header = ["1", "2"]
-    schema = StructType([StructField("u", IntegerType())])
-    with pytest.warns(Warning):
-        df = movielens.load_spark_df(spark, header=header, schema=schema)
-        assert len(df.columns) == len(schema)
+    shutil.rmtree(local_cache_dir, ignore_errors=True)
```
gramhagen: this is a little scary, and one reason working with TemporaryDirectories is safer, if someone accidentally changes local_cache_dir we could cause some damage here.

tests/smoke/test_movielens.py
```diff
@@ -22,109 +24,151 @@
...
     assert genres[0] == genres[1]
     assert genres[0] == genres_example
 
-    # Test movie data load (not rating data)
-    df = movielens.load_pandas_df(size=size, header=None, title_col="Title", genres_col="Genres")
+    year = head["Year"].values
+    assert year[0] == year[1]
+    assert year[0] == year_example
+
+    shutil.rmtree(local_cache_dir, ignore_errors=True)
```
gramhagen: same comment as above

tests/smoke/test_movielens.py
```diff
@@ -22,109 +24,151 @@
...
-    with pytest.raises(ValueError):
-        movielens.load_spark_df(spark, local_cache_path='.')
-
-    # Test if use schema when both schema and header are provided
-    header = ["1", "2"]
-    schema = StructType([StructField("u", IntegerType())])
-    with pytest.warns(Warning):
-        df = movielens.load_spark_df(spark, header=header, schema=schema)
-        assert len(df.columns) == len(schema)
+    shutil.rmtree(local_cache_dir, ignore_errors=True)
```
gramhagen: same

reco_utils/dataset/movielens.py
loomlike: Love this idea! cool! and yeah, totally agree that shutil.rmtree is very scary

loomlike: Checked my codes and recalled that we want to keep the downloaded data file (ratings and item data files) for later use. We don't need to keep the original zip file but `url_utils.maybe_download` uses urllib's `urlretrieve` which writes the downloaded data directly into the given file path. Can use w/ tempfile like [this](https://docs.python.org/3/howto/urllib2.html) which requires changes in `maybe_download` to use `shutil.copyfileobj()`

So I would like to keep the current mechanism for now. Instead, I'm going to refactor smoke and integration tests' shutil.rmtree to use tempdir you suggested, where we don't want to keep any downloaded data. What do you think @gramhagen?

reco_utils/dataset/movielens.py
loomlike: Good catch. Should've moved up after the refactor.

reco_utils/dataset/movielens.py
loomlike: Same as the previous comment. Should've removed after the refactor.

reco_utils/dataset/movielens.py
gramhagen: it's fine, i was just proposing to keep the zip file instead of the extracted files for simplicity, but it's one less thing to do if we keep the data files.

if we want to continue using urlretrieve (which is legacy, but I'm fine with retaining), then you can just give it a path in the temp dir

```
with TemporaryDirectory() as tmp_dir:
  filepath = maybe_download(url=URL, filename='data.zip', work_directory=tmp_dir.name)
  ... extract data from filepath to cache location...
```

reco_utils/dataset/movielens.py
miguelgfierro: I think we need this function, same in criteo. There are two reasons, one is single responsibility and the second is because this is a function that can be used independently for just download the dataset

reco_utils/dataset/movielens.py
miguelgfierro: Same reason here, I think we need something like this. I would even do it public, this is something that we could use independently

loomlike: Right. This was used from multiple functions before the refactoring, which is replaced by _maybe_download_and_extract.

reco_utils/dataset/movielens.py
loomlike: @miguelgfierro Remember you mentioned someone asked to have this standalone download function. What's the use-case for this compare to him directly downloads the data from MovieLens' url?

gramhagen: but we already have a function that provides this responsibility, wrapping it with another function that literally reformats a variable is redundant.

what advantage is this function versus calling maybe_download directly?

reco_utils/dataset/movielens.py
loomlike: Ok, github is not loading the latest thread automatically, so didn't see @miguelgfierro's comment when I replied, and this makes the discussion really complicated :-(

I prefer to merge this `_extract` function into `_maybe_download_and_extract`, because as Scott pointed out, extract will work only after the zip file is downloaded. Regarding the single responsibility, I guess the function is small enough and its name describes the task enough. A similar example is from tensorflow's tutorial's [cifar10 dataset downloader](https://github.com/tensorflow/models/blob/master/tutorials/image/cifar10/cifar10.py) which has maybe_download_and_extract function.

Regarding `download_movielens`, I think we may want to keep it separately. As Miguel pointed out, someone may want to just download it so that he can use other files like tag or user info (100k data includes user info). Currently, `_maybe_download_and_extract` only extracts rating and item data. One good reason for having this download wrapper is, to make sure the users download the same data as the repo's data-loader.

reco_utils/dataset/movielens.py
loomlike: @gramhagen One good reason of having the download wrapper is, to make sure the users download the same data as the repo's data-loader. 

Please find more context for this from my reply to the earlier comment.

reco_utils/dataset/movielens.py
loomlike: this is good. I like the solution.

loomlike:
@miguelgfierro @gramhagen I've been thinking about `local_cache_path` and using `tempfile`. I would like to use `local_cache_path` as users' intention of keeping the data files and wanting to reuse the file sometime later. What do you guys think? Is that too implicit or make sense?

So my expected behavior is:
* when local_cache_path is not provided (==None), store the zip and data files under `tempfile.TemporaryDirectory` so that can be cleaned up safely once the data loading is done.
* if local_cache_path is explicitly provided, that means user will reuse the files later so that want to keep it in the local cache directory. In this case, download and extract files to the folder and don't remove the files.

reco_utils/dataset/movielens.py
miguelgfierro: > but we already have a function that provides this responsibility, wrapping it with another function that literally reformats a variable is redundant. what advantage is this function versus calling maybe_download directly?

not exactly the same responsibility we are making super easy to download a dataset. I can do `download_movielens("100k")` I don't need to know which url I have to go. 

Also, there was a person asking for this. 

Then finally, before, we had a function called load_dataset, which downloads, extracts and loads the data. That's 3 responsibilities in one function.

gramhagen:
I like that approach @loomlike

reco_utils/dataset/movielens.py
miguelgfierro: I think maybe_download_and_extract is ok, however, I ended up changing that. Please take a look at this function: https://github.com/Microsoft/Recommenders/blob/a66ec60b0fdd8be9e593dc22d94a8de98ee85df1/reco_utils/dataset/criteo_dac.py#L48

Basically there is something called load_criteo() which will do the whole thing, which is download, extract and load criteo. I created 3 steps if you look at the function. Initially, I thought of using download_and_extract, but I don't see the value of having 2 of the 3 steps together and then the load in a different steps. I think, at least in this case, it's clearer to have 3 single steps: download, extract and load. 

But again, I don't think it is super key as long as we follow single responsibility

reco_utils/dataset/movielens.py
gramhagen: heh, ok. i'm fine either way honestly. it makes sense to create a seam between pulling data and loading it so we can easily change how we are pulling/extracting/caching and how we load the data for different formats. But in my mind pulling/extracting/caching is a single responsibility for getting the data until it is generalized beyond one dataset. I'm hoping we can offload much of this kind of work to something like https://github.com/intake/intake in the future though, since it's only a necessity to do the more interesting recommendation work

tests/integration/test_movielens.py
loomlike: Will change to use tempfile as you suggested!

reco_utils/dataset/movielens.py
jingyanwangms: Just to chime in on one use case, we need a separate download_movielens function for running aml example where the dataset is downloaded to local cache then uploaded to Azure datastore.

reco_utils/dataset/movielens.py
```diff
@@ -128,50 +135,50 @@ def item_has_header(self):
...
-    header=(
-        DEFAULT_USER_COL,
-        DEFAULT_ITEM_COL,
-        DEFAULT_RATING_COL,
-        DEFAULT_TIMESTAMP_COL
-    ),
-    local_cache_path="ml.zip",
+    header=DEFAULT_HEADER,
+    local_cache_path=None,
     title_col=None,
```
yueguoguo: The column names are `col_user` for example in some other utilities. Do we want to make them consistent?

miguelgfierro:
About the solution @loomlike proposed, I think it's a great idea

reco_utils/dataset/movielens.py
loomlike: Good suggestion! I personally prefer to name a variable as we speak. e.g. "column name" -> `col_name` and "user column" -> `user_col`. What do you guys think?

reco_utils/dataset/movielens.py
loomlike: @jingyanwangms Usually a user will need to do feature extraction after dataset download (especially when using item or user features), and if he/she runs hyperdrive 100 jobs, the same feature extraction will be done 100 times... Is there any reason for uploading a raw datafile instead of uploading the preprocessed data? If so, isn't it better to run the downloader script from aml compute?

fyi, in wide & deep aml-hyperdrive example, we upload the preprocessed data (pandas dataframe) to the datastore.

reco_utils/dataset/movielens.py
yueguoguo: I see. No strong personal preference here lol. Maybe later we can change these all together.

PR #650, opened 2019-03-14
Set up CI with Azure Pipelines

bethz:


bethz:
I created a public Devops pipeline to test and it automatically created this PR (which is NOT desirable).

PR #661, opened 2019-03-19
Refactored libffm converter

yueguoguo:
### Description
* Refactor libffm converter to follow scikit learn style.
* Generate output like feature count, field count, etc., which are useful for using DeepRec algorithms. 

### Related Issues
#605 

### Checklist:

- [x] I have added tests.


tests/unit/test_pandas_df_utils.py
```diff
@@ -110,19 +109,20 @@ def test_csv_to_libffm():
     import tempfile
     import os
 
-    filedir = tempfile.tempdir
+    filedir = tempfile.gettempdir()
```
miguelgfierro: What is the difference? I'm using tempdir in other parts of the code

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -117,55 +117,120 @@ def libffm_converter(df, col_rating=DEFAULT_RATING_COL, filepath=None):
...
+    def fit(self, df, col_rating=DEFAULT_RATING_COL):
+        """Fit the dataframe for libffm format. In there method does nothing but check the validity of 
+        the input columns
+
+        Args:
+            df (pd.DataFrame): input Pandas dataframe.
+            col_rating (str): rating of the data.
+
+        Return:
+            The instance of the conveter
```
miguelgfierro: typo in converter
Not sure what is the correct naming when we are returning self, could it be:
```
Returns:
   obj: The instance of the converter
```

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -117,55 +117,120 @@ def libffm_converter(df, col_rating=DEFAULT_RATING_COL, filepath=None):
...
+            df (pd.DataFrame): input Pandas dataframe.
+            col_rating (str): rating of the data.
+
+        Return:
+            pd.DataFrame: output libffm format dataframe.
+        """
+        return self.fit(df, col_rating=col_rating).transform(df)
+
+    def get_params(self):
+        """Get parameters (attributes) of the libffm converter
```
miguelgfierro: The return argument in the doctstrings is missing

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -117,55 +117,120 @@ def libffm_converter(df, col_rating=DEFAULT_RATING_COL, filepath=None):
...
+                    # Check whether (field, feature) tuple exists in the dict or not.
+                    # If not, put them into the key-values of the dict and count the index.
+                    if (field, feature) not in field_feature_dict:
+                        field_feature_dict[(field, feature)] = idx
+                        idx += 1
+
+        self.field_count = len(self.field_names)
+        self.feature_count = idx - 1
+
+        def _convert(field, feature, field_index, field_feature_index_dict):
```
miguelgfierro: I would say it is confusing to have this private function in the middle of the code, probably taking it out the method transform would make the code more readable

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -73,7 +73,7 @@ def filter_by(df, filter_by_df, filter_by_cols):
     ]
 
 
-def libffm_converter(df, col_rating=DEFAULT_RATING_COL, filepath=None):
+class LibffmConverter():
```
miguelgfierro: are we sure we need a class here? Normally classes in python are useful when we are saving state, do we need to do it here?

miguelgfierro: if we are going to use a class (see comment above), it is a good practice to inhereit from object: `LibffmConverter(object)`

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -117,55 +117,120 @@ def libffm_converter(df, col_rating=DEFAULT_RATING_COL, filepath=None):
...
+        if not all([x in df.columns for x in self.field_names]):
+            raise ValueError("Not all columns in the input dataset appear in the fitting dataset")
+
+        df_new = df.copy()
+
+        # Encode field-feature.
+        idx = 1
+        field_feature_dict = {}
+        for field in self.field_names:
+            if df_new[field].dtype == object:
```
miguelgfierro: here maybe:
`isinstance(df_new[field], object)`
is this what you want?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -117,55 +117,120 @@ def libffm_converter(df, col_rating=DEFAULT_RATING_COL, filepath=None):
...
+    def __init__(self, filepath=None):
+        if filepath is not None:
+            # Check the existence of path
+            import os
+            if not os.path.exists(filepath):
+                raise ValueError("The specified file path {} does not exist".format(self.filepath))
+        
+        self.filepath = filepath
+    
+    def fit(self, df, col_rating=DEFAULT_RATING_COL):
```
miguelgfierro: This method is confusing. I would have thought that fit actually makes the transformation, but transform does. Also, the code `LibffmConveter().fit(df_feature, col_rating='rating')` looks a lot like Java. 
What if you have all the checkers in the init and then just user transform to compute the transformation? Would this be doable?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -117,55 +117,120 @@ def libffm_converter(df, col_rating=DEFAULT_RATING_COL, filepath=None):
...
+        Return:
+            pd.DataFrame: output libffm format dataframe.
+        """
+        if not self.col_rating in df.columns:
+            raise ValueError("Input dataset does not contain the label column {} in the fitting dataset".format(self.col_rating))
+
+        if not all([x in df.columns for x in self.field_names]):
+            raise ValueError("Not all columns in the input dataset appear in the fitting dataset")
+
+        df_new = df.copy()
```
miguelgfierro: is there a way to avoid this to reduce memory? In my mind we are taking data from the input df and convert it in another thing. Are there parts of df that we are reusing or the output is completely new?

reco_utils/dataset/pandas_df_utils.py
yueguoguo: The function is only used in the method here.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: This is to mimic the sklearn Binarizer and other similar stuff in the module of `preprocessing`, so mostly I followed their pattern. See [here](https://github.com/scikit-learn/scikit-learn/blob/7b136e9/sklearn/preprocessing/data.py#L1754).

yueguoguo: It should be equivalent in this case see [here](https://stackoverflow.com/questions/4015417/python-class-inherits-object).

reco_utils/dataset/pandas_df_utils.py
yueguoguo: `isinstance` does not work because here we are checking the data type of the column, instead of the type of the instance (i.e., the dataframe column) itself - the instance of the latter is a pandas Series.

tests/unit/test_pandas_df_utils.py
yueguoguo: This is an interesting thing - using `-s` in `pytest` is not compatible with `tempdir`. Did you notice anything similar?

reco_utils/dataset/pandas_df_utils.py
yueguoguo: This is to mimic the behaviour of `scikit-learn` preprocessing modules.

reco_utils/dataset/pandas_df_utils.py
miguelgfierro: cool

reco_utils/dataset/pandas_df_utils.py
miguelgfierro: ok

reco_utils/dataset/pandas_df_utils.py
miguelgfierro: I see, you are following [this example](https://github.com/scikit-learn/scikit-learn/blob/7b136e9/sklearn/preprocessing/data.py#L1812).

The thing is, sklearn are adding the fit method because they need it for pipelines. They say:
```
    def fit(self, X, y=None):
        """Do nothing and return the estimator unchanged
        This method is just there to implement the usual API and hence
        work in pipelines.
        Parameters
        ----------
        X : array-like
        """
        check_array(X, accept_sparse='csr')
        return self
```
I guess they are adding fit, just because many of their methods has a fit and because of that, they really need to have it so other elements like pipelines don't break. @msalvaris do you know if this is the case?

yueguoguo: Decided to remove `fit_transform` method as what you suggested. After a while (I mean after having a good sleep and breakfast 😄) I think this makes sense - indeed chaining the method is not very necessary in this class. I will still keep `fit` and `transform` methods for the reason that a fitted converter can be applied to a new dataframe with the same schema.

reco_utils/dataset/pandas_df_utils.py
yueguoguo: Removed it.

reco_utils/dataset/pandas_df_utils.py
miguelgfierro: makes sense

tests/unit/test_pandas_df_utils.py
miguelgfierro: as I said to @loomlike, here maybe we could leverage the fixtures for temporary paths of pytest: https://docs.pytest.org/en/latest/tmpdir.html

yueguoguo: This is a good idea, as it will be used across different test suites. Has this been implemented in the data loader utility PR? I did not find it...

miguelgfierro:
+ @loomlike for review, since he is also having fun with directories in pytest :-)

yueguoguo:
@miguelgfierro makes perfect sense! @loomlike feel free to commit if you would like to incorporate the changes in this PR.

loomlike:
I will create new issue for that and modify all related tests.
Please don't be blocked by that!

yueguoguo:
@miguelgfierro @loomlike and @anargyri I will merge this guy and update afterwards when test fixture of temp dir is put in.

Issue #665, opened 2019-03-19
[FEATURE] Use nb_conda for kernel management


gramhagen:
### Description
Including nb_conda in installed conda packages will simplify kernel mgmt in jupyter notebooks 

### Expected behavior with the suggested feature
- removes step of manually installing the kernel spec
- active env will default to Python 3
- other conda envs can be inspected and kernels will automatically be available

### Other Comments
this was a suggestion from @ilyaomsft

PR #666, opened 2019-03-20
adding nb_conda to conda script and updating setup docs

gramhagen:
### Description
Includes nb_conda in conda packages (which also installs nb_conda_kernels). Supports automatic installation of conda envs for jupyter and ability to inspect them through conda tab.

- note there will likely need to be some work done to the test setup to ensure this is supported there

### Related Issues
#665 

### Checklist:

- [ ] I have added tests.


gramhagen:
also I win the 😈 prize!

SETUP.md
```diff
@@ -119,12 +119,10 @@ To install all three environments:
...
 </details>
 
-### Register the conda environment as a kernel in Jupyter
+### Using the conda environment as a kernel in Jupyter
 
-We can register our created conda environment to appear as a kernel in the Jupyter notebooks.
-
-    conda activate my_env_name
-    python -m ipykernel install --user --name my_env_name --display-name "Python (my_env_name)"
+By using [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels) all conda environments (with ipykernel installed) will automatically be made available as kernels in the Jupyter notebooks.<br>
```
miguelgfierro: I wouldn't remove  the ipykernel instructions, not everyone will use nb_conda. What I was talking to Ilya in the hack is to have the nb_conda_kernels as an optional suggestion to developers.

SETUP.md
```diff
@@ -119,12 +119,10 @@ To install all three environments:
 
 </details>
 
-### Register the conda environment as a kernel in Jupyter
+### Using the conda environment as a kernel in Jupyter
```
miguelgfierro: Since you did a PR about this, there have been some users that were not aware that the DSVM has jupyterhub which is accessible from port 8000. Some people are starting jupyter kernels, which is confusing. I think we should mention jupyterhub, how to register the new kernel and how to connect to the notebooks

SETUP.md
gramhagen: why wouldn't we want to use nb_conda? i would think if it's easier for the user that would be the default and developers could specify to not install it via a flag in the generate_conda_file script

miguelgfierro: I'm not sure which one would be easier for the user. For me at least it is easier to use the command line, it is just one command vs having to go to the notebook, press several tabs, etc. 

What I'm suggesting is to have both options, in case more people want to use the command line as an alternative to nb_conda

yueguoguo: It might be reasonable to provide options for the SETUP.

gramhagen: yeah, we should definitely support both. but the idea would be with this you don't have to run an extra command, and it's the same experience on the jupyter side - you just select the kernel you want (assuming you even need to switch it).

should i keep this as the default behavior and provide details in the setup for manually disabling and /or installing the kernel via the command line

miguelgfierro: that sounds reasonable

PR #671, opened 2019-03-21
Updates to Spark configuration used in tests

gcampanella:
### Description
Updates the Spark configuration used for tests based on available cores and physical memory.

### Checklist:
- [ ] I have added tests.


tests/conftest.py
```diff
@@ -37,21 +35,27 @@ def spark(app_name="Sample", url="local[*]", memory="1G"):
...
+        .config("spark.driver.memory", "{:.2f}g".format(physical_mem * 0.2))
+        .config("spark.executor.cores", n_cores - 1)
+        .config("spark.executor.instances", 1)
+        .config("spark.executor.memory", "{:.2f}g".format(physical_mem * 0.6))
         .config("spark.local.dir", "/mnt")
-        .config("spark.worker.cleanup.enabled", "true")
-        .config("spark.worker.cleanup.appDataTtl", "3600")
-        .config("spark.worker.cleanup.interval", "300")
-        .config("spark.storage.cleanupFilesAfterExecutorExit", "true")
+        .config("spark.sql.shuffle.partitions", 1)
```
miguelgfierro: I think the error comes because  the value is a string:

("spark.sql.shuffle.partitions","n")

gcampanella: From the Azure DevOps logs, it looked like the JVM didn't like fractional memory amounts. I've fixed this now; let's see if it runs...

Issue #655, opened 2019-03-18
[BUG] broken links in Hyperdrive wide and deep notebook


anargyri:
### Description

https://github.com/Microsoft/Recommenders/blob/master/notebooks/04_model_select_and_optimize/azureml_hyperdrive_wide_and_deep.ipynb
The "entry script" links do not work (give 404).

loomlike:
oh yeah. seems our folder name has changed from 'aml' to 'azureml'. There might be another broken links in both wide-deep and azureml- notebooks because of the name changes. I will check them. Thanks @anargyri !

yueguoguo:
Fixed in #678

PR #678, opened 2019-03-24
wide-deep AML Hyperdrive notebook link fix

loomlike:
### Description
Fix broken links to script in markdown cell 
Also, found bugs in `azureml/wide_deep.py` cause by the recent refactor. So fixed it.
TODO: Should add unit-test for azureml entry script, because it seems like the `azureml/wide_deep.py` script has been broken since the recent refactor and we didn't have the unit-test to catch that.

### Related Issues
#655 

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/reco_utils/azureml/wide_deep.py
+++ b/reco_utils/azureml/wide_deep.py
@@ -33,11 +33,17 @@ OUTPUT_NOTEBOOK = "wide_deep.ipynb"
 
 
 def _log(metric, value):
+    """AzureML log wrapper.
+
+    Record list of int or float as a list metrics so that we can plot it from AzureML workspace portal.
+    Otherwise, record as a single value of the metric.
+    """
     if run is not None:
-        if isinstance(value, list) and len(value) > 0 and isinstance(value[0], (int, float))
+        if isinstance(value, list) and len(value) > 0 and isinstance(value[0], (int, float)):
             run.log_list(metric, value)
         else:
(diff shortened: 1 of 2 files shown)


reco_utils/azureml/wide_deep.py
```diff
@@ -33,11 +33,17 @@
 
 
 def _log(metric, value):
```
anargyri: Shouldn't this method be public (with run as an argument) so that it is used in other python files in the azureml directory?

loomlike:
@yueguoguo Can you specify which one? I tested three of them and all works. Are you getting 404 error?

reco_utils/azureml/wide_deep.py
loomlike: This wrapper is necessary only when we use a notebook with papermill on AzureML and want to record the same metrics multiple times within the notebook-context, e.g. record training error for every epoch. Because papermill will overwrite the values whenever we record by using the same key, we had to store the values into a list and record the list with the key. Once the notebook run is done, AzureML script (which is in this case, `azureml/wide_deep.py`) gets the list from papermill and write to AzureML via this `_log` function.

This approach is temporal because AzureML is planning to support papermill / notebooks sometime soon. If the day come, we will replace this method :-) So, I would keep this method private unless we have other examples that need this approach before the AzureML's update.

Issue #685, opened 2019-03-26
[BUG] Wide and Deep notebook is throwing an error when ITEM_FEAT_COL=None


loomlike:
### Description
Wide and Deep notebook is throwing an error when `ITEM_FEAT_COL=None`

### How do we replicate the issue?
set `ITEM_FEAT_COL=None` and run the notebook

### Expected behavior (i.e. solution)
The model should run w/ and w/o item features

### Other Comments
The bug fix will include some other improvements in coding style:
- Use tempfile instead of `shutil.rmtree` to clean things up
- Allow `EPOCHS=0` (do 1 batch) for quick testing of the notebook

PR #686, opened 2019-03-26
Wide-deep notebook bug fix

loomlike:
### Description
`ITEM_FEAT_COL=None` error fix
Ensure train at least 1 batch (when Epochs=0)
azureml_hyperdrive_wide_and_deep.ipynb link fix
Refactor dataloading and splitting cells
Refactor tests to use tempfile

### Related Issues
#685

### Checklist:

- [x] I have added tests.


tests/integration/test_notebooks_gpu.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-import shutil
+from tempfile import TemporaryDirectory
```
yueguoguo: Interestingly, when I do `pytest` with `-s` to print something to console it failed if `TemporaryDiretory` or `tempdir` is used; `gettempdir()` works.... 😂

miguelgfierro: does this work for you?
```
with TemporaryDirectory() as tmp_dir:

```

loomlike: lol ... w/o '-s' does this work?

miguelgfierro:
@loomlike do you know why the tests are not being executed?

tests/unit/test_notebooks_gpu.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-import shutil
+from tempfile import TemporaryDirectory
```
miguelgfierro: for the tests, we could use the pytest fixture tmpdir https://docs.pytest.org/en/latest/tmpdir.html. It's quite neat. There is also tmpdir_factory available in case we need to have more control on the temporary folders

loomlike: Good! I'm working on it.

loomlike:
@miguelgfierro no idea why the first commit didn't trigger the tests lol. But the second commit went through all of them.

tests/integration/test_notebooks_gpu.py
yueguoguo: Neither of the above worked for me... `pytest` fixture `tmpdir` works, so this is so far the best practice.

anargyri:
> This looks nice, just one question that I asked you in the other PR, is `str(tmpdir.mkdir("wide_deep_0"))` the recommended way of getting the string of the path?

Also worth noting that they use a python library "not recommended for new projects":
https://docs.pytest.org/en/latest/tmpdir.html
https://py.readthedocs.io/en/latest/path.html

notebooks/00_quick_start/wide_deep_movielens.ipynb
```diff
@@ -47,24 +47,23 @@
     "import sys\n",
```
anargyri: <p>You don't need MODEL_DIR = None</p><p>since you do</p><p>if MODEL_DIR is None:</p><p>&nbsp;&nbsp;tmp_dir = TemporaryDirectory()</p><p>&nbsp;&nbsp;MODEL_DIR = tmp_dir.name</p><p>afterwards.</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/686/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

anargyri: <p>You say "the hyperparameters are&nbsp;optimized for the training set."</p><p>But for the tuning there is also a validation set.</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/686/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

anargyri:
Part of this notebook is about preprocessing and encoding features. I suspect this can be used by other algorithms in the future, so it is probably better to separate it into a util instead.

notebooks/00_quick_start/wide_deep_movielens.ipynb
loomlike: good catch. more specifically, we further split the training set into training / validation (instead of split the data into three from the first place). The reason was to make sure azureml_w_d notebook uses the same set of training set as the other notebooks. I will put more contextual detail to this notebook.

loomlike: I will get `name 'MODEL_DIR' is not defined` error if I don't define it. Right?

loomlike:
@anargyri thanks for great suggestions! See my response here:
* preprocessing and encoding features: totally agree. Actually, I'm planning to work on utilizing other features as well, like **movie's released year vs rating timestamp** (assuming users rated movies by the time they "watched"). Let me create another issue for that and take care "adding util functions" from there.
* pytest's tmpdir fixture: Yeah pytest is using py.path. I think in the future version of pytest, they will (should) update the underneath module to pathlib, and we can keep using as it is (hopefully). Any suggestions?

loomlike:
@miguelgfierro Regarding `tmpdir`, here, I will get this error if I pass the object directly:
`Object of type 'LocalPath' is not JSON serializable`

I will update this accordingly once we have the conclusion from the other thread!

notebooks/00_quick_start/wide_deep_movielens.ipynb
anargyri: I mean, you just need 
`tmp_dir = TemporaryDirectory()
MODEL_DIR = tmp_dir.name`
You don't need the if statement because there is no alternative.

anargyri:
> * pytest's tmpdir fixture: Yeah pytest is using py.path. I think in the future version of pytest, they will (should) update the underneath module to pathlib, and we can keep using as it is (hopefully). Any suggestions?

See my comment in another PR; there is a factory that uses pathlib
https://docs.pytest.org/en/latest/reference.html#tmp-path-factory-factory-api

tests/integration/test_notebooks_gpu.py
loomlike: @yueguoguo Can you double check what was the reason for the failure? @miguelgfierro and I had discussed and we want to use `with TemporaryDirectory() as td:` back again, because:

pytest's fixture (tmpdir or tmp_path or tmp_factory) keeps 3-latest temp folders and we don't want that (also saw someone complaining about the old folders are not removed completely from their CI machine).

I tested using TemporaryDirectory from my side and both unit and smoke tests are passed, with `-s`

yueguoguo: Tested and it works with the context manager. Thanks!

notebooks/00_quick_start/wide_deep_movielens.ipynb
loomlike: Ah, I got your point. MODEL_DIR can be passed as a parameter when the notebook got called by papermill. Usecase is to store model checkpoints.
Will add comments to make it clear.

tests/integration/test_notebooks_gpu.py
loomlike: Let me refactor all tempdir things in a new issue and PR! I will merge the current PR as soon as possible since the main fix is the bugfix.

PR #687, opened 2019-03-27
Notebook showcasing NNI on Surprise SVD

anargyri:
### Description

NNI is a toolkit for hyperparameter tuning that incorporates several tuning methods from the literature.
This notebook shows how to apply NNI for tuning Surprise SVD on Movielens100K. 

### Checklist:

- [x] I have added tests.


notebooks/04_model_select_and_optimize/azureml_hyperdrive_surprise_svd.ipynb
```diff
@@ -197,46 +197,6 @@
     "print(cpu_cluster.status.serialize())"
    ]
   },
-  {
```
anargyri: I removed this cell which was not used anywhere.

reco_utils/nni/svd_training.py
```diff
@@ -0,0 +1,150 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: there is another svd_training for azureml, do you think it is possible to have a common one?

miguelgfierro: I think we should consider having a folder where we put all the code related to tuning, instead of `reco_utils/nni`, if tomorrow we have another hypterparameter tunning framework, I think the repo can get messy. Alternatively, we could have the nni code inside evaluation. This is were @yueguoguo added the other hyperparameter evaluation

anargyri: Not really, because Hyperdrive and NNI have differences and there are differences in how they parse arguments, the arguments are not the same in fact, reporting of the metrics differs etc. Even if it were possible to refactor some common code, it would be confusing to the developer.

anargyri: I don't think evaluation is the right place; after all tuning is a different thing. It would be more appropriate to have a folder called tuning and, under that folder, azureml and nni and the spark tuning code. This would require testing again all the tuning notebooks, so I would leave it for a separate PR. Hyperdrive and NNI rely on several path names, so they will break as soon as you change the folder structure.

miguelgfierro: that makes sense, added issue here: https://github.com/Microsoft/Recommenders/issues/689

miguelgfierro: ok

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
```diff
@@ -0,0 +1,835 @@
+{
```
miguelgfierro: <p>Title and initial explanation of nni and this notebook are missing</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/687/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

miguelgfierro: <p>maybe here we could also add the nni version</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/687/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

miguelgfierro: <p>one question, are you optimizing over precision at k? bellow you show the results of precision at k</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/687/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

miguelgfierro: <p>can we parametrize exp_directory?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/687/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

miguelgfierro: <p>I think it would be interesting to give a brief overview of each of the algos we are using</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/687/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

miguelgfierro: <p>this is amazing, probably it would be interesting to show the rest of the metrics. Also, we can discuss why are the business benefits of optimizing over precision@k and that the rest of the metrics won't be optimal</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/687/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='46'/>

anargyri: Yes, I can add these. I could also add the times but I need to dig them out of the logs, since there is no API for that.

anargyri: Good point.

anargyri: Unfortunately the only way to get this is to read the experiment ID and the ID of the best trial from the Web UI. They are not available programmatically.

anargyri: Yes, the script optimizes the primary metric, which is precision@k.
I should add some description to explain how this is done.

loomlike: Also could you add brief outline of the contents? Maybe simple 1, 2, 3 steps as an overview so that readers can easily see what's yaml file is doing, json is doing, where the compute gets done, what's the prerequisites , etc.

loomlike: Ah.. nvm Miguel's talking about the algos used in NNI. I was more like wonder about when to use hyperopt vs hyperdrive vs NNI

loomlike: You can simply add end_time - start_time as a metric, right?

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
```diff
@@ -0,0 +1,835 @@
...
+      "text/plain": [
+       "               precision@10\n",
+       "Metis              0.102345\n",
+       "Random Search      0.089339\n",
+       "TPE                0.084968\n",
+       "Annealing          0.069510\n",
+       "Hyperband          0.068443\n",
+       "SMAC               0.064179\n",
+       "Evolution          0.058849"
+      ]
```
loomlike: 😍 love this table. As you said, please add time too (time or steps which one is better? I think step (# params sets navigated to find this optimal set) will be better since time can be affected by other things too).

reco_utils/nni/svd_training.py
```diff
@@ -0,0 +1,150 @@
...
+
+def svd_training(params):
+    """
+    Train Surprise SVD using the given hyper-parameters
+    """
+    logger.debug("Start training...")
+    train_data = pd.read_pickle(path=os.path.join(params['datastore'], params['train_datapath']))
+    validation_data = pd.read_pickle(path=os.path.join(params['datastore'], params['validation_datapath']))
+
+    svd = surprise.SVD(random_state=params['random_state'], n_epochs=params['epochs'], verbose=params['verbose'],
```
loomlike: Can we use **kwargs here? Seems you can simply remove few keys `datastore`, `train_datapath`, `validation_datapath`, `surprise_reader`, `rating_metrics` ... (oh well.. more than I thought)

but would be nice to just pass like: `svb = surprise.SVB(**hyperparams)` if small refactor of this function can do that.

reco_utils/nni/svd_training.py
```diff
@@ -0,0 +1,150 @@
...
+        k = params['k']
+        for metric in ranking_metrics:
+            result = getattr(evaluation, metric)(validation_data, all_predictions, col_prediction='prediction', k=k)
+            logger.debug("%s@%d = %g", metric, k, result)
+            if metric == params['primary_metric']:
+                metrics_dict['default'] = result
+            else:
+                metrics_dict[metric] = result
+
+    if len(ranking_metrics) == 0 and len(rating_metrics) == 0:
```
loomlike: Let's early fail! Test this at the very beginning.

scripts/generate_conda_file.py
```diff
@@ -74,6 +75,7 @@
     "hyperopt": "hyperopt==0.1.1",
     "idna": "idna==2.7",
     "memory-profiler": "memory-profiler>=0.54.0",
+    "nni": "nni==0.5.2.1"
```
loomlike: so, as I mentioned somewhere, do we want to include this in the env? Our env is getting bigger and thus quite slow to install & activate it now.

One option is to remove not-very-common modules (e.g. `hyperopt` `nni` `memory-profiler`) and add a prerequisite step to each notebook that has dependency on such packages. Any thoughts? @miguelgfierro @yueguoguo

Crysple: BTW, it seems there is a comma missing here.

notebooks/04_model_select_and_optimize/README.md
```diff
@@ -34,3 +35,8 @@ To configure this notebook to communicate with your workspace, type in your Azur
     "workspace_name": "<workspace-name>"
 }
 ```
+
+### NNI Configuration
+The NNI command `nnictl` comes installed with the conda environment. 
+In order to use the SMAC tuner, it has to be installed first with the following command 
+`nnictl package install --name=SMAC`.
```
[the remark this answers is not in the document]
yueguoguo: Agree on cleaning this up sooner rather than later. Btw, do we want to support docker container image?

scripts/generate_conda_file.py
yueguoguo: I agree with @loomlike and it indeeds affects user experience in some way. It would be great to have a finer granularity of yaml files for different workloads. E.g., we may add options to yaml generation script to enable a desired env.

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
```diff
@@ -0,0 +1,835 @@
...
+   "source": [
+    "with open('../../reco_utils/nni/search_space_svd.json', 'w') as fp:\n",
+    "    json.dump(hyper_params, fp)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We also create a yaml file for the configuration of the trials and the tuning algorithm to be used (in this experiment we use TPE). The tuning trials will be executed locally on a [Standard_D16_v3 virtual machine](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes-general#dv3-series-1) (16 vcpus, 64 GB memory)."
```
yueguoguo: Not sure if there is any information about how NNI uses the underlying multi-core capability in a machine to perform the concurrent run, and how it is compared against a cluster of machines. The reason for the ask is that multi-node may not always be more computationally efficient to multi-core. @scarlett2018

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
yueguoguo: +1 ❤️

yueguoguo: * It would be great to have a discussion on why these calculated metrics for each tuning algorithm. The results of Random Search is not worse than most of the other algorithms. Maybe this is owing to some factors that may not be directly related to the algorithms themselves (e.g., # trials, initial state, etc.), but it would be good to understand better about this. 
* Would it be better to compare RMSE given SVD implementation in Surprise is meant for predicting ratings?

reco_utils/nni/svd_training.py
yueguoguo: Having a dedicated folder for tuning makes perfect sense.

notebooks/04_model_select_and_optimize/README.md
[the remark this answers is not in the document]
anargyri: I added it in `generate_conda.py`, so it is part of the conda environment. I could remove it however; it doesn't need to come with reco_base etc. by default.

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
anargyri: The number of HP configurations is the same for all methods (equal to the number of trials), except Hyperband, which tries fewer. 

About the results, I would not like to overstate them, because there is significant variation; when trying the notebook again, the order changes. Because of this variation, it is hard to tell which method performs better. In some sense it is not surprising that Random does well this time and badly other times; this is what randomness is, sometimes it will hit the right HPs. Also, the other tuning methods may be smarter in terms of sampling, but they do not exploit the nature of the ML algorithm i.e. the methods ignore whether you are doing SVD or Wide and Deep etc.

The good thing about NNI is that you have all these methods in your quiver and you can check in practice which one works best.

scripts/generate_conda_file.py
miguelgfierro: I think there is going to be a trade between how big the dependencies are vs how difficult to generate the conda file if we start adding many options. Many dependencies will make that when we install the env it takes longer, for the tests this can be annoying. On the other hand, it is easier for the user to have a simple conda generator so they can just do two steps: generate conda file and install conda file.

You know what I'll do? I will first make sure that we are using all the libraries in the generator, for example, are we using dask? Also, we are using gitpython and pymongo for tracking the metrics, but this is not related to the utilities. Maybe it would make sense to have an env file in the tracking metrics folder with the libraries that we need there. 
What do you guys think

notebooks/04_model_select_and_optimize/README.md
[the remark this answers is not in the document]
miguelgfierro: > Btw, do we want to support docker container image?
yeah that's something that we have discussed also with @msalvaris, I think it would be a nice addition and would make some developments easy

Furthermore, it would give @msalvaris a new incentive to start contributing to this repo :-)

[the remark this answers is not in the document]
msalvaris: https://imgflip.com/i/2x8vpp

scripts/generate_conda_file.py
anargyri: Definitely, for the tracking metrics the conda env could be different. I like Le's suggestion of using flags for the `generate_conda.py` script e.g. `--nni` would allow one to use NNI. That said, if we remove tracking metrics and debugging from the conda env, we are not left with many optional libraries other than NNI and hyperopt. For example, there is no rationale for separating recommendation algorithms, so you cannot e.g. make fastai an option.

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
anargyri: I leave the distributed cluster for another PR. It is worth noting that the time required is shorter than the Hyperdrive notebook (which ran a cluster) and NNI seemed to exploit the multiple cores well. I gave it a number of cores comparable to the Hyperdrive cluster, so maybe this favours NNI. I think there are cases when a single machine would be preferable and others when a cluster of nodes would be better, but probably the latter holds with larger data sets than ML100K.

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
anargyri: About RMSE, maybe it would make the function easier to the tuning methods to optimise. However, precision seems to occur more often in practical situations, since it correlates better with KPIs like CTR. Movie recommendations is more of this scenario rather than the goal of "understanding" the user.

scripts/generate_conda_file.py
miguelgfierro: > I like Le's suggestion of using flags for the generate_conda.py script e.g. --nni would allow one to use NNI

The danger we have with this is to end up with a super complex conda generator. Now we have 3 options: base, --gpu and --pyspark. If we start adding options for libraries I think for the user it will be more complicated. Imagine that we have the --nni option, I'm sure users will start installing the repo, will forget about the --nni option and will have to go back to the readme to see what happens

anargyri: Good point, it doesn't affect many libraries anyway. Also, once you have created the conda environment, it is fast to update it.

miguelgfierro:
hey @anargyri, nni team sent us a way of making tests, one question, do you want to create the tests in this PR or do you prefer to have them in a different one?

anargyri:
> hey @anargyri, nni team sent us a way of making tests, one question, do you want to create the tests in this PR or do you prefer to have them in a different one?

Yes, I'd rather address it within this PR.

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
yueguoguo: > About the results, I would not like to overstate them, because there is significant variation; when trying the notebook again, the order changes. Because of this variation, it is hard to tell which method performs better. In some sense it is not surprising that Random does well this time and badly other times; this is what randomness is, sometimes it will hit the right HPs. Also, the other tuning methods may be smarter in terms of sampling, but they do not exploit the nature of the ML algorithm i.e. the methods ignore whether you are doing SVD or Wide and Deep etc.

Interesting to know the impact of initial state of trials on the final result. I am not entirely sure but how does NNI does cross validation when it tunes the parameters?

anargyri: It is like Hyperdrive, NNI does not handle the evaluation, it is all implemented in the python script. Btw I do not do cross-validation, just plain holdout. I might get more reliable numbers with CV, however that would make testing slower and the notebook is for demonstration anyway.

reco_utils/nni/nni_utils.py
```diff
@@ -0,0 +1,58 @@
+import json
+import numpy as np
+import os
+import requests
+import time
+
+NNI_REST_ENDPOINT = 'http://localhost:8080/api/v1/nni'
```
miguelgfierro: @anargyri, one question, all these utils are for supporting the code or for the tests?

anargyri: They enable running the code automatically from the notebook.

miguelgfierro: ok cool

anargyri:
> 'took about 10' minutes right?

Fixed.

reco_utils/nni/nni_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def get_experiment_status(status_url):
+    nni_status = requests.get(status_url).json()
+    return nni_status['status']
+
+
+def check_experiment_status():
+    while True:
+        time.sleep(20)
```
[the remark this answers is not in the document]
gramhagen: also, it might be good to have a limit on the number of retries: 
```
while i < MAX_RETRIES:
  i += 1
  ...
```

scripts/generate_conda_file.py
```diff
@@ -52,6 +52,7 @@
     "scikit-learn": "scikit-learn==0.19.1",
     "scipy": "scipy>=1.0.0",
     "scikit-surprise": "scikit-surprise>=1.0.6",
+    "swig": "swig==3.0.12",
```
gramhagen: what are we using swig for?

reco_utils/nni/nni_utils.py
```diff
@@ -0,0 +1,58 @@
...
+            raise RuntimeError("NNI experiment failed to complete with status {}".format(status))
+
+
+def check_stopped():
+    while True:
+        time.sleep(20)
+        try:
+            get_experiment_status(NNI_STATUS_URL)
+        except:
+            break
```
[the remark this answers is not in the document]
msalvaris: Erm, if you get an error in the above call your try except will simply swallow it and never pass it on to the user. This will make it super hard to debug.

reco_utils/nni/nni_utils.py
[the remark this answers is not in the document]
msalvaris: To be honest I might move that kind of logic to get_experiment_status

scripts/generate_conda_file.py
anargyri: It is needed in order to install one of the tuners:
https://github.com/Microsoft/nni/blob/master/docs/en_US/Builtin_Tuner.md#SMAC

reco_utils/nni/nni_utils.py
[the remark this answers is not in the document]
msalvaris: https://www.calazan.com/retry-decorator-for-python-3/

reco_utils/nni/nni_utils.py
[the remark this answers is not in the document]
anargyri: It waits until the NNI experiment completes. NNI status will eventually become 'DONE' or 'TUNER_NO_MORE_TRIAL', or NNI will raise an error.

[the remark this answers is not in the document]
gramhagen: but what happens when status is 'running'?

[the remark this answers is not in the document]
anargyri: Someone needs to stop (i.e. kill) the experiment; while status is 'running' this function will wait.

reco_utils/nni/nni_utils.py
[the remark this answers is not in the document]
gramhagen: yeah i was thinking about a decorator, or using (backoff)[https://pypi.org/project/backoff/], but would probably take some adjustment to handle the different success / failure conditions

reco_utils/nni/nni_utils.py
```diff
@@ -0,0 +1,58 @@
...
+        time.sleep(20)
+        all_trials = requests.get(NNI_TRIAL_JOBS_URL).json()
+        if np.all(['finalMetricData' in trial for trial in all_trials]):
+            break
+
+
+def get_trials(optimize_mode):
+    if optimize_mode not in ['minimize', 'maximize']:
+        raise ValueError("optimize_mode should equal either 'minimize' or 'maximize'")
+    all_trials = requests.get(NNI_TRIAL_JOBS_URL).json()
```
gramhagen: why is eval used here?

another option which is a little cleaner:
```
sorted_trials = sorted(all_trials, key=lambda x: x[0]['default'], reverse=optimize_mode=='maximize')
trials =  [(eval(trial['finalMetricData'][0]['data']), trial['logPath'].split(':')[-1]) for trial in sorted_trials]
best_trial_path = trials[0][1]
with open(os.path.join(best_trial_path, 
```

reco_utils/nni/nni_utils.py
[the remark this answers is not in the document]
gramhagen: sorry, i guess i meant status = 'DONE', get_experiment_status will generate an error in that case? do you mean check_experiment_status?

scripts/generate_conda_file.py
gramhagen: oh, are they planning on including it as a requirement for the nni package so we don't need to add it separately?

anargyri: Good suggestion. @scarlett2018

scarlett2018: @gramhagen @anargyri - swig is only used by SMAC example, so it is only part of the SMAC example installation (`use nnictl package install --name=SMAC`) . Currently, we did not support install swig along with other SMAC required packages for the following reasons:
- it requires sudo
- different os may have differ installation approaches

we did not plan to do it in the future yet for the same reasons (above). cc @QuanluZhang  in case you have further technical detail to discussion.

anargyri: I could move the installation of swig from `generate_conda_file.py` to the notebook.
@gramhagen  What do you think?

reco_utils/nni/nni_utils.py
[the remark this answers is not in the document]
anargyri: If `get_experiment_status()` returns any status it means that the experiment is not "stopped" (in the sense that the server process and the REST interface are active). This is what `check_stopped()` checks. The experiment may have completed (status == 'DONE') but the `nnictl` command still returns info about the experiment and the URL is accessible. 
The purpose of `check_stopped()` is to be called right after `nnictl stop` in order to verify that this command worked. You need to ensure this before you create a new tuning experiment.

reco_utils/nni/nni_utils.py
anargyri: `trial['finalMetricData'][0]['data']` is a string. `eval` here is doing the same thing as `json.loads()` and returns a dict.

scripts/generate_conda_file.py
gramhagen: i'm ok with keeping it here, if we want to support using nni, i was just curious how this was being used.

reco_utils/nni/nni_utils.py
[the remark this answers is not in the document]
gramhagen: ah, ok i guess the context for how this was to be used was what I was missing. alright, maybe just a note in the docstring that you must call this after nnictl stop is issued otherwise it will hang.

anargyri:
@miguelgfierro There is a spark test failing that should be irrelevant to this PR.

anargyri:
I was trying pytest-httpserver 
https://pytest-httpserver.readthedocs.io/en/latest/index.html
https://pypi.org/project/pytest-httpserver/
which mocks an HTTP server, for testing the NNI utils.
What do you think? Should we have this type of tests? It would require adding pytest-httpserver to the conda env though.

gramhagen:
i'd recommend just leveraging mock to patch requests and provide the desired results, we already have that in our environment.

https://stackoverflow.com/questions/15753390/how-can-i-mock-requests-and-the-response

reco_utils/nni/nni_utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+NNI_REST_ENDPOINT = 'http://localhost:8080/api/v1/nni'
+NNI_STATUS_URL = NNI_REST_ENDPOINT + '/check-status'
+NNI_TRIAL_JOBS_URL = NNI_REST_ENDPOINT + '/trial-jobs'
+WAITING_TIME = 20
+MAX_RETRIES = 5
+
+
+def get_experiment_status(status_url):
+    nni_status = requests.get(status_url).json()
```
miguelgfierro: there are some docstrings missing

notebooks/04_model_select_and_optimize/nni_surprise_svd.ipynb
```diff
@@ -0,0 +1,1066 @@
...
+    "check_metrics_written()\n",
+    "svd = surprise.dump.load(os.path.join(get_trials('maximize')[3], \"model.dump\"))[1]\n",
+    "test_results_metis = compute_test_results(svd)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Hyperband follows a different style of configuration from other tuners. See [the NNI documentation](https://nni.readthedocs.io/en/latest/hyperbandAdvisor.html). Note that the [training script](../../reco_utils/nni/svd_training.py) needs to be adjusted as well, since each Hyperband trial receives an additional parameter `STEPS`, which corresponds to the resource allocation _r<sub>i</sub>_ in the [Hyperband algorithm](https://arxiv.org/pdf/1603.06560.pdf). In this example, we used `STEPS` in combination with `R` to determine the number of epochs that SVD will run for in every trial."
```
QuanluZhang: We will change `STEPS` to `TRIAL_BUDGET` for easy understanding in the next release v0.7, very sorry for this broken change. We will inform you when v0.7 is released. Thanks.

miguelgfierro: Thanks for the notice

miguelgfierro: Thanks for the notice

anargyri:
I added tests except `get_trials()` which is not clear how to do. It requires both the `patch` decorator and `TemporaryDirectory`, which don't seem to combine.

Issue #690, opened 2019-03-28
[BUG] Error running spark notebooks unit tests on Windows DSVM


roalexan:
### Description
Running pspark unit tests as described [here ](https://github.com/Microsoft/Recommenders/tree/staging/tests)as follows: pytest tests/unit -m "notebooks and spark and not gpu". Getting error: 3 failed, 2 passed, 129 deselected, 4 warnings. Please see attached file of full stacktrace [here](https://github.com/Microsoft/Recommenders/files/3019448/stacktrace-pyspark-unittests-windowsdsvm.txt) for details.

### In which platform does it happen?
Windows DSVM

### How do we replicate the issue?
Create Windows DSVM, create and activate reco_spark (or reco_full) Python environment, run unit test: pytest tests/unit -m "notebooks and spark and not gpu"

### Expected behavior (i.e. solution)
should not have any errors

loomlike:
changing spark driver's local file path as `file:///` which is a shorten form of `file://localhost/` I will create PR soon once I tested all on Databricks too

PR #693, opened 2019-03-28
Movielens bug fix on Windows-spark

loomlike:
### Description
Fix local file path to make work on Windows spark
Change movielens tests to use pytest tmpdir fixture

### Related Issues
#690 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/tests/integration/test_movielens.py
+++ b/tests/integration/test_movielens.py
@@ -3,7 +3,6 @@
 
 import os
 import pytest
-from tempfile import TemporaryDirectory
 from reco_utils.dataset.movielens import (
     load_pandas_df,
     load_spark_df,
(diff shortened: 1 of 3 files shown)


tests/integration/test_movielens.py
```diff
@@ -68,49 +67,51 @@ def test_load_pandas_df(
...
     # Test if correct data are loaded and local_cache_path works
-    with TemporaryDirectory() as tmp_dir:
-        # Test if can handle different size of header columns
-        header = ["a"]
-        df = load_pandas_df(size=size, local_cache_path=tmp_dir, header=header)
-        assert len(df) == num_samples
-        assert len(df.columns) == max(
-            len(header), 2
-        )  # Should load at least 2 columns, user and item
+    tmp_dir = str(tmpdir.mkdir("load_pandas_df"))
```
yueguoguo: It is worth considering a `try ... finally ...` or `with open ... as` for testing with a temp directory. The reason for this is that if the testing fails somewhere in the middle (after the temporary directory is created), the temporary directory will not be removed.

miguelgfierro: I think this is the pytest fixture `tmpdir`, it is supposed to be erased after the test is finished. It would be good to make sure that this is the behavior

miguelgfierro: some questions about this:
1) is tmpdir a pathlib object?
2) is `str(pathlib_path)` the recommended way to get the string from the path object?
3) do you actually need to pass a str or can you directly pass the object

loomlike: Found interesting fact from the pytest page:

>Temporary directories are by default created as sub-directories of the system temporary directory. The base name will be pytest-NUM where NUM will be incremented with each test run. Moreover, **entries older than 3 temporary directories will be removed**.

So, you don't need to remove manually, but worth to note that it won't be removed right away.

Also, tmp-dir is unique for each test. E.g. `test_load_item_df(tmpdir)`,
tmpdir is `.... Temp\pytest-of-jumin\pytest-1\test_load_item_df_100k_1682_1_0/`

As you can see,  it creates a folder "pytest-of-USER' and create subfolder `pytest-1` (2, 3, ,...) where the last three of them will be kept, and create the children folders under the subfolder for each test function for each parameter. In the example, the test function is `test_load_item_df` and the params are `(100k, 1682, 1, 0)`

loomlike: 1: py.path.local path object
2: tested and we can just pass the object in this case (which will be converted as a string implicitly). If we have the case where implicit conversion doesn't work, then we should use str() that case.
3: answered in 2

miguelgfierro: >  entries older than 3 temporary directories will be removed.

let's say we download movielens in 4 of the tests within the same file, then is the first directory going to disappear? or is it 4 directories in different files?

loomlike: Ok, there is a way to get the string path in py.path, which is `strpath` property. But if pytest changes py.path to pathlib, it will break the test because pathlib doesn't have strpath. I think just explicitly calling `str()` is the easiest way to make it work on any cases. What do you think?

loomlike: As long as the tests are under the same pytest run, it should be there. 
So, for your case, you will have one folder e.g. `pytest-n` and have 4 test folders under it.
the folder `pytest-n` will be removed after 2 additional pytest runs.

loomlike: ah... you know what. if pytest updates to use pathlib, anyway mkdir or other functions will not work lol per @anargyri's point about pytest is using py.path which is not recommended to the production  code, what will be the best way then?

anargyri: They already have a factory built on pathlib:
https://docs.pytest.org/en/latest/tmpdir.html
"The tmp_path_factory is a session-scoped fixture which can be used to create arbitrary temporary directories from any other fixture or test.

It is intended to replace tmpdir_factory, and returns pathlib.Path instances."
https://docs.pytest.org/en/latest/reference.html#tmp-path-factory-factory-api

miguelgfierro: yeah, I think what @anargyri suggested is what @loomlike and I were looking for https://docs.pytest.org/en/latest/tmpdir.html#the-tmpdir-factory-fixture

loomlike: @miguelgfierro @anargyri unfortunately, no it is not. I tested and still see all the **last three** temp folders and files in those three folders. They are not cleaned up after the session done. "Session-scoped" means just the temp-factory "object" is session-scoped, not the physical file-system folders. So, still our concerns remain the same -- the garbage files on the machine.

We can manually clean up, but to do that, should use `shutil.rmtree` again, because `pathlib.Path` doesn't have the function to delete non-empty folders.

So far, `with TemporaryDirectory() as tmp_dir:` works best for our purpose: remove the folders and files with context manager + can create many folders within the same test-function safely.

loomlike:
Based on our discussion and my investigation, don't see any benefit of using pytest's temp path fixtures over `tempfile.TemporaryDirectory`. If we want to share a folder and files across the multiple tests or for the session, then yes we should use pytest's tmp_path_factory, which is intended to use for that specific use-cases. But my test functions are to see if the module download and unzip data. Let me know if this make sense.

miguelgfierro:
we have 3 PRs with the issue of temporary paths lol https://github.com/Microsoft/Recommenders/pull/661#pullrequestreview-220998802

roalexan:
yes, this should fix it.

Issue #677, opened 2019-03-22
[BUG] Error running test notebook als_deep_dive on Windows DSVM


roalexan:
### Description
Get this error running unit test:

`E           papermill.exceptions.PapermillExecutionError:
E           ---------------------------------------------------------------------------
E           Exception encountered at "In [15]":
E           ---------------------------------------------------------------------------
E           TypeError                                 Traceback (most recent call last)
E           <ipython-input-15-9eddaaa7187e> in <module>
E                 4               relevancy_method='top_k', k=TOP_K)
E                 5
E           ----> 6 rank_metrics = {'MAP': map_at_k(*args, **kwargs),
E                 7                 'NDCG': ndcg_at_k(*args, **kwargs),
E                 8                 'Precision': precision_at_k(*args, **kwargs),
E
E           ~\notebooks\Recommenders\reco_utils\evaluation\python_evaluation.py in check_column_dtypes_wrapper(rating_true, rating_pred, col_user, col_item, col_rating, col_prediction, *args, **kwargs)
E                57             raise TypeError(
E                58                 "data types of column {} are different in true and prediction".format(
E           ---> 59                     col_user
E                60                 )
E                61             )`

### In which platform does it happen?
Windows DSVM

### How do we replicate the issue?
The steps are exactly the same as issue https://github.com/Microsoft/Recommenders/issues/616, except the test to run is:
pytest tests/unit -m "not notebooks and not spark and not gpu"
or more specifically:
pytest tests/unit -m "notebooks and not spark and not gpu" -k vw

### Expected behavior (i.e. solution)
should run successfully without any errors

### Other Comments
Had a discussion with Jun and Scott about this. One possible workaround is to make this change to: Recommenders\reco_utils\evaluation\python_evaluation.py:

```
from: if rating_true[col_user].dtypes != rating_pred[col_user].dtypes:
to: if rating_true[col_user].dtype.type.__base__ != rating_pred[col_user].dtype.type.__base__:
```
and
```
from: if rating_true[col_item].dtypes != rating_pred[col_item].dtypes:
to: if rating_true[col_item].dtype.type.__base__ != rating_pred[col_item].dtype.type.__base__:
```
Scott made the following suggestion, which makes a lot of sense:

I think the right solution is to create a couple utility functions to check that columns exists, and another to check that two dataframes have the same base types for columns of interest, or alternatively to remove unnecessary type checks

Another open question is how pervasive this type of check happens - it would be nice to fix this across the repo.

[...]

Issue #648, opened 2019-03-14
[FEATURE] Optimize column checking codes in each function with decorator


yueguoguo:
### Description
Decorator pattern is used to check columns in Python evalutor (https://github.com/Microsoft/Recommenders/blob/master/reco_utils/evaluation/python_evaluation.py). It would be good to have it used across the repo where column checking is a common task for properly using a function.

### Expected behavior with the suggested feature
The code for functions that may need to check columns can be optimized from
```
def func_old(df, col_user, col_item, col_rating):
    columns = df.columns
    if col_user not in columns:
        raise ValueError("Column {} is not in the input data".format(col_user))
    ...
```
to
```
@check_column
def function_new(df, col_user, col_item, col_rating):
    ...
```

gramhagen:
what can we do if the required columns are missing? catching it and raising an error seems the same as not checking at all and raising the error where it occurs in the function

yueguoguo:
@gramhagen Yea it should check the existence of columns as well. If it occurs in the function, the error might be different. See #543 for example. The input datasets to the evaluator may be different, and the difference in dtypes may result in some unexpected behaviors.

gramhagen:
yeah, you're right. I am working on some changes to the python evaluation code, and will push a PR for review of that approach. Changes to address type checking can be put there.

PR #699, opened 2019-03-30
Gramhagen/python eval class

gramhagen:
### Description
Changing python evaluation to use class based approach similar to spark evaluation. This reduces redundant operations and should accelerate testing. Plus it provides consistency between spark and python evaluation approaches.

Also added a couple methods to pandas_df_utils: 
- has_columns: to check if dataframe has specified columns
- has_same_base_dtype: to check if specified columns across two dataframes have the same base type (can be used as keys for joining). This will ensure the types are both string, or int, or float without raising unneccesary errors if they differ within a base dtype (e.g. int32 vs int64)

noticed that Miguel's url_utils refactor is showing up here ? disregard for now, will clean up the branch in the final PR

### Related Issues
#677 
#648 

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/reco_utils/dataset/pandas_df_utils.py
+++ b/reco_utils/dataset/pandas_df_utils.py
@@ -270,7 +270,7 @@ def has_same_base_dtype(df_1, df_2, columns=None):
 
     result = True
     for c in columns:
-        if df_1[c].dtype.__base__ != df_2[c].dtype.__base__:
+        if df_1[c].dtype.base != df_2[c].dtype.base:
             print('Columns {} do not have the same base dtype'.format(c))
             result = False
     return result
--- a/reco_utils/dataset/pandas_df_utils.py
+++ b/reco_utils/dataset/pandas_df_utils.py
@@ -255,3 +255,39 @@ def negative_feedback_sampler(
     )
 
     return df_sample
+
+
+def has_same_base_dtype(df_1, df_2, columns=None):
+    """Check if specified columns have the same base dtypes
+
+    Args:
+        df_1 (pd.DataFrame): first dataframe
+        df_2 (pd.DataFrame): second dataframe
+        columns (list(str)): columns to check, None checks all columns
+    Returns:
+        bool: True if dataframe columns have the same base dtype
+    """
+
+    result = True
+    for c in columns:
+        if df_1[c].dtype.__base__ != df_2[c].dtype.__base__:
+            print('Columns {} do not have the same base dtype'.format(c))
+            result = False
+    return result
+
+
+def has_columns(df, columns):
+    """Check if dataframe has necessary columns
+    Args:
+        df (pd.DataFrame): dataframe
+        columns (list(str)): columns to check for
+    Returns:
+        bool: True if dataframe has specified columns
+    """
+
+    result = True
+    for c in columns:
+        if c not in df.columns:
+            print('Missing column: {}'.format(c))
+            result = False
+    return result
--- a/reco_utils/evaluation/python_evaluation.py
+++ b/reco_utils/evaluation/python_evaluation.py
@@ -3,7 +3,6 @@
 
 import numpy as np
 import pandas as pd
-from functools import wraps
 from sklearn.metrics import (
     mean_squared_error,
     mean_absolute_error,
@@ -13,6 +12,7 @@ from sklearn.metrics import (
     log_loss
 )
 
+from reco_utils.dataset.pandas_df_utils import has_columns, has_same_base_dtype
 from reco_utils.common.constants import (
     DEFAULT_USER_COL,
     DEFAULT_ITEM_COL,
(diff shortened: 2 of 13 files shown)


gramhagen:
side note - we should disable tests on draft PRs

loomlike:
I personally like this suggestion and I had similar PR before,
but remember we wanted `sklearn`-like apis where each evaluation metrics exist as a separate function... 

I know many parts/calculations are redundant and not very efficient... specifically top-k selection part which involve sorting. One way to improve the efficiency is to make evaluation functions to accept pre-calculated intermediate dataframes (outputs of `merge_ranking_true_pred()`) that can be shared across those functions. If those dataframes are provided, skip the re-calculation.

anargyri:
> I personally like this suggestion and I had similar PR before,
> but remember we wanted `sklearn`-like apis where each evaluation metrics exist as a separate function...
> 
> I know many parts/calculations are redundant and not very efficient... specifically top-k selection part which involve sorting. One way to improve the efficiency (without making them as a class) is to make evaluation functions to accept pre-calculated intermediate dataframes (outputs of `merge_ranking_true_pred()`) that can be shared across those functions. If those dataframes are provided, skip the re-calculation.

Yes, as Jun Ki says, there has been a certain amount of discussion around this and we have ended up with the current approach i.e. follow the scikit-learn style for python evaluation and the Spark MLLib style for Spark evaluation. One reason is conformity with those libraries, another is that the scikit-learn style is more pythonic and there are benefits to the user when the python functions are explicit (see Zen of python etc.).

I think there are other ways to reduce redundancies within the pythonic framework e.g. refactoring using private functions, while keeping the interface to the user explicit and more readable.

reco_utils/evaluation/python_evaluation.py
```diff
@@ -23,663 +23,347 @@
 )
 
 
-def check_column_dtypes(f):
-    """
-    Checks columns of dataframe inputs.
+class Evaluation:
```
miguelgfierro: hey @gramhagen we had a discussion about the metrics interfaces months ago. We decided to follow the sklearn/TF/Pytorch convention of using functions instead of classes for Python. In PySpark we will follow the PySpark convention of using classes. There are multiple reasons to follow this:
1) We want to follow a data-driven software design [based on evidence](https://github.com/Microsoft/Recommenders/wiki/Coding-Guidelines#evidence-based-software-design). In this case the evidence is that sklearn/TF/Pytorch uses this convention. A user that uses sklearn will find natural to use functions in our metrics.
2) There's no need for classes when using metrics, you just need to make an operation with 2 df. Abusing classes increases complexity and reduce readability.
3) As @loomlike noted, we discussed that when performing some metrics, we are repeat certain operations. The way for maintaining single responsibility and at the same time being efficient is be using cache (same as we do in the github metrics). This is something that we haven't done so far and would be nice to implement.
4) We also thought about changing the PySpark interfaces and make then more pythonic (which would be having functions instead of classes), however, we left it as classes. Here the same principle applies, a PySpark user will find more natural to use classes on everything (we are not saying that this is the recommended approach, in fact it is not, but PySpark code have a strong Java background so to follow their convention we code in this line)

yueguoguo: I think the general principle is to make the interface and design pattern (if there is 😂) mimic the existing packages we are using. In case there is no existing approaches, we will look for evidence to support any proposed idea. 

Regarding the replicated calculation of metrics, it makes sense to have some preliminary steps (e.g., data join) before certain metrics of the same type are calculated to eliminate redundant computation. 

Would love to have everyone's opinion.

gramhagen: I understand, and agree with the concept of using existing tools / apis. While I don't feel that this implementation is an abuse of python classes as it actually simplifies the current usage I think @loomlike has the right idea. We can reduce redundant operations (i.e. greatly speed up evaluation) by exposing the methods for extracting the truth and prediction arrays. Honestly at that point there's little need for any of our wrapper methods as we can use sklearn methods directly.

I can refactor this towards that goal if that sounds reasonable to everyone.

anargyri: We still need to keep the wrappers, because sklearn does not have all the metrics we want (or that we will build in the future). Also, the one that sklearn has may be unsuitable (e.g. precision), because it was not designed with recommendation use cases in mind.

But I agree we should delegate `merge_rating_true_pred` and `merge_ranking_true_pred` to the user, for efficiency.

miguelgfierro: > One way to improve the efficiency (without making them as a class) is to make evaluation functions to accept pre-calculated intermediate dataframes (outputs of merge_ranking_true_pred()) that can be shared across those functions. If those dataframes are provided, skip the re-calculation.

Not sure if you are proposing to have a preprocessing step and then call the interface that we have. Something like:

```
x_true, x_pred = preprocess(df_true, df_pred)
p = precision_at_k(x_true, x_pred)
```
is this what @loomlike  and @gramhagen  are saying?

anargyri: It is more like 
`merged_true_pred, df_hits = preprocess_ranking(df_true, df_pred)
p = precision_at_k(merge_true_pred, df_hits)`

loomlike: @miguelgfierro Rather, introducing additional arg for cached df so that users can use the evaluation functions as usual **or** pre-compute the ranked-df and provide it to ranking evals as follows:
```
def precision_at_k(..., ranking_true_pred=None):
    if ranking_true_pred is None:
        ranking_true_pred = merge_ranking_true_pred(...)
    
    _, df_hit, n_users = ranking_true_pred

    ....

>>> cached_ranking = merge_ranking_true_pred(...)
>>> p_k = precision_at_k(..., cached_ranking)
>>> n_k = ndcg_at_k(..., cached_ranking)
...
````

If we want to use this approach, can we rename `merge_ranking_true_pred` function to be more descriptive? I think naming after the goal of the function make more sense rather than how the function achieve that goal. We can describe about 'how' part at the docstring. 

Also adding "Examples:" to all ranking evaluation function docstrings will be great!

Comment (anargyri) [reply reco_utils/evaluation/python_evaluation.py]: Also, it probably makes sense for `cached_ranking` to be a class rather than a tuple, since it will be a black box for the users anyway.

Comment (miguelgfierro) [reply reco_utils/evaluation/python_evaluation.py]: ok what I'm thinking has the same result but doesn't need to add another argument, it would be something like this:

```
@lru_cache()
def merge_ranking_true_pred(df_true, df_pred):
    ...

def precision_at_k(df_true, df_pred):
    ranking_true_pred = merge_ranking_true_pred(...)
    ...
    ...

>>> p_k = precision_at_k(df_true, df_pred )
>>> n_k = ndcg_at_k(df_true, df_pred)
```
Note that the only thing I added was the decorator, this is catching the computation the first time it is done. Second time merge_ranking_true is called, it gets the cached value. I've never done it with code, we we have this in the repo metrics: https://github.com/Microsoft/Recommenders/blob/master/scripts/repo_metrics/git_stats.py#L61 and I guess it should be the same.

I think @msalvaris has done something similar in the past, is this doable?

Comment (gramhagen) [reply reco_utils/evaluation/python_evaluation.py]: Lru cache stores inputs as keys in a dict, so they need to be hashable. I believe this will be a problem for dataframes and mutable arrays. There may be other cache implementations that provide workarounds? We should time the whole process to figure out where bottlenecks really are. Could be caching isn't a huge help, if we can avoid recalling the merging.

Comment (gramhagen) [reply reco_utils/evaluation/python_evaluation.py]: Lru cache stores inputs as keys in a dict, so they need to be hashable. I believe this will be a problem for dataframes and mutable arrays. There may be other cache implementations that provide workarounds? We should time the whole process to figure out where bottlenecks really are. Could be caching isn't a huge help, if we can avoid recalling the merging.

Comment (msalvaris) [reply reco_utils/evaluation/python_evaluation.py]: Dataframes should be hashable, we have used lru_cache with dataframes before. But I agree that only add caching when you know you need it.

File: reco_utils/dataset/pandas_df_utils.py [hunk 2]
```diff
@@ -255,3 +255,39 @@ def negative_feedback_sampler(
...
+        df (pd.DataFrame): dataframe
+        columns (list(str)): columns to check for
+    Returns:
+        bool: True if dataframe has specified columns
+    """
+
+    result = True
+    for c in columns:
+        if c not in df.columns:
+            print('Missing column: {}'.format(c))
```
msalvaris: Would Python logging not be a better fit?

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -255,3 +255,39 @@ def negative_feedback_sampler(
...
+        columns (list(str)): columns to check, None checks all columns
+    Returns:
+        bool: True if dataframe columns have the same base dtype
+    """
+
+    result = True
+    for c in columns:
+        if df_1[c].dtype.base != df_2[c].dtype.base:
+            print('Columns {} do not have the same base dtype'.format(c))
+            result = False
```
msalvaris: Maybe simply return False in the loop or break, it seems that going through all the columns is unecessary

anargyri: Or compare the results of `df.dtypes.values` with `numpy.all()`.

reco_utils/evaluation/python_evaluation.py
anargyri: In [14]: a = np.array([1,2,3])

In [15]: hash(a)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-15-57b555d30865> in <module>
----> 1 hash(a)

TypeError: unhashable type: 'numpy.ndarray'

In [16]: df = pd.DataFrame(data={'a': 1, 'b': 'ewiu'}, index=[0])

In [17]: hash(df)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-17-5e54663ef3fb> in <module>
----> 1 hash(df)

~/anaconda3/envs/reco_pyspark/lib/python3.6/site-packages/pandas/core/generic.py in __hash__(self)
   1814     def __hash__(self):
   1815         raise TypeError('{0!r} objects are mutable, thus they cannot be'
-> 1816                         ' hashed'.format(self.__class__.__name__))
   1817
   1818     def __iter__(self):

TypeError: 'DataFrame' objects are mutable, thus they cannot be hashed

reco_utils/dataset/pandas_df_utils.py
miguelgfierro: +1 to using logging, we have it in other parts of the code in fact

reco_utils/evaluation/python_evaluation.py
anargyri: https://docs.python.org/3.1/glossary.html
"Objects which are instances of user-defined classes are hashable by default; they all compare unequal, and their hash value is their id()"
So, we could wrap the dataframes inside a class we define, but this would not be a good idea, because:
- if you change the values in the encapsulated dataframe, the hash does not change (I double checked)
- it becomes implicit and obfuscates the user.

miguelgfierro: but have you tried to check if lru_cache works with that function? that would just accelerate things and the code would be the same

anargyri: https://docs.python.org/3/library/functools.html
"Since a dictionary is used to cache results, the positional and keyword arguments to the function must be hashable."

msalvaris: Thanks for checking @anargyri. Evindently forgot what we did :D.  You can use hash_pandas_object and simply implement lru_cache for dfs

msalvaris: https://stackoverflow.com/questions/31567401/get-the-same-hash-value-for-a-pandas-dataframe-each-time

reco_utils/dataset/pandas_df_utils.py
msalvaris: You have to make sure that both dataframes columns are ordered the same way if you do that @anargyri

msalvaris:
> One way to improve the efficiency (without making them as a class) is to make evaluation functions to accept pre-calculated intermediate dataframes (outputs of `merge_ranking_true_pred()`) that can be shared across those functions. If those dataframes are provided, skip the re-calculation.

I am going to play the devils advocate here and say do we need to improve speed? Is the perfromance unacceptable? Maybe this is premature optimization?

anargyri:
> > One way to improve the efficiency (without making them as a class) is to make evaluation functions to accept pre-calculated intermediate dataframes (outputs of `merge_ranking_true_pred()`) that can be shared across those functions. If those dataframes are provided, skip the re-calculation.
> 
> I am going to play the devils advocate here and say do we need to improve speed? Is the perfromance unacceptable? Maybe this is premature optimization?

We don't need it so much for small data sets, but when running the notebooks with a large users-items matrix, the overhead is significant. We need to compute the top-k recommendations from the entire users-items matrix, so this can be heavy and we don't want to do it more than once: 
https://github.com/Microsoft/Recommenders/blob/d98ea81e25d51c794510c25c75ce2c7837768735/reco_utils/evaluation/python_evaluation.py#L265

msalvaris:
> 
> 
> > > One way to improve the efficiency (without making them as a class) is to make evaluation functions to accept pre-calculated intermediate dataframes (outputs of `merge_ranking_true_pred()`) that can be shared across those functions. If those dataframes are provided, skip the re-calculation.
> > 
> > 
> > I am going to play the devils advocate here and say do we need to improve speed? Is the perfromance unacceptable? Maybe this is premature optimization?
> 
> We don't need it so much for small data sets, but when running the notebooks with a large users-items matrix, the overhead is significant. We need to compute the top-k recommendations from the entire users-items matrix, so this can be heavy and we don't want to do it more than once:
> 
> [Recommenders/reco_utils/evaluation/python_evaluation.py](https://github.com/Microsoft/Recommenders/blob/d98ea81e25d51c794510c25c75ce2c7837768735/reco_utils/evaluation/python_evaluation.py#L265)
> 
> Line 265 in [d98ea81](/Microsoft/Recommenders/commit/d98ea81e25d51c794510c25c75ce2c7837768735)
>  df_hit = get_top_k_items(dataframe=df_pred, col_rating=self.col_prediction, k=self.k)

OK, then I guess you will have to choose some option to maintain the calculated values.

loomlike:
@msalvaris @anargyri when we do hyperparameter tuning, we may need to do 100 or more runs... then yes we need performance improvement! .. (but that is a special case)

anargyri:
> @msalvaris @anargyri when we do hyperparameter tuning, we may need to do 100 or more runs... then yes we need performance improvement! .. (but that is a special case)

Actually this is a very good point. Because for HP tuning with AzureML Hyperdrive (and more generally when one wants to do distributed evaluation) the LRU cache will not be sufficient (unless there is a way to access the cached result and distribute it to all the nodes).

anargyri:
> @msalvaris @anargyri when we do hyperparameter tuning, we may need to do 100 or more runs... then yes we need performance improvement! .. (but that is a special case)

This is a very good point because in such cases, with many different nodes, it may make sense to distribute some time consuming tasks like the computation of predictions (not part of evaluation) or the computation of top-k (part of ranking evaluation) to exploit the availability of the nodes. This probably requires additional functionality to the LRU cache.

reco_utils/evaluation/python_evaluation.py
miguelgfierro: For this discussion, see this PR that speeds up the rating computation https://github.com/Microsoft/Recommenders/pull/707 and this issue to improve it even further https://github.com/Microsoft/Recommenders/issues/708

gramhagen:
ok, thanks for the useful feedback, i believe most of the items here are being / have been handled in different PRs (#707, #712, #713) and remaining issues logged (#708) so I will close this draft

Issue #701, opened 2019-04-01
[Feature] pytest temp directory


loomlike:
#### Description
Benefit of using pytest fixture `tmp_path` is that pytest handles temp directory creation and cleanup for us. E.g., `tempfile.TemporaryDirectory` may fail to cleanup the directory if the files are being used. On the other hand, pytest's tmpdir module will revisit the existing temp directories next time we run tests and cleanup.

One drawback of using pytest's tmpdir is, it keeps the **last three directories**.  But we want to clean up the directory right after we use it. 

I can think about two options to solve our issue:
##### 1. Use `TemporaryDirectory` with pytest's `tmp_path`:
```
def test_A(tmp_path):
    with TemporaryDirectory(dir=tmp_path) as td:
        ....
```
Pros - TemporaryDirectory will take care of cleanup. If it fails, the worst case is the dir will last for next three tests (and then tmp_path will be cleaned up by the pytest module)
Cons - Every test scripts that uses tmp dir should import TemporaryDirectory 

##### 2. Use `tmp_path_factory` and cleanup by ourselves
```
def test_A(tmp_path_factory):
    td = tmp_path_factory.mktemp("test_1")
    ....
    shutils.rmtree(os.path.abspath(str(td)))
```
Pros - No need to import TemporaryDirectory
Cons - Have to use `rmtree` if we want to clean up right after we use the dir; `pathlib.Path.resolve()` doesn't seem reliable across different platforms so that we need to use `os.path.abspath`

I will probably want to use 1st option....but feel free to share thoughts!

yueguoguo:
+1 on option1. Is it heavy to import Temporary Directory? Given we don't have every testing script importing TemporaryDirectory, maybe the overhead is negligible...

miguelgfierro:
@loomlike, with TemporaryDirectory, when is the directory erased?

loomlike:
@miguelgfierro you're right! Thought we wanted to use *context manager*(which I was obsessed about!) from our test functions which doesn't work with the fixtures, but we don't need to use context manager.

I will use 1st approach with our own fixture (what Miguel showed) for extra safety. Also will use function scope instead of module.

miguelgfierro:
@loomlike, about the context manager thing, there is an alternative in pytests: https://pytest.readthedocs.io/en/2.8.7/yieldfixture.html

maybe yield fixture will allow you to do this:
```
with tmp_dir:
   # do something with tmp_dir
[...]

PR #706, opened 2019-04-03
pytest tmp_dir fixture

loomlike:
### Description
* Introduce tmp_dir fixture
* Refactor tests to use the fixture
* Fix a bug in movielens load_spark_df - could throw an error when item_df is None. Now it handles the case.
* Make movielens loader require at least two columns in the header, user and item

### Related Issues
#701 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/dataset/movielens.py
+++ b/reco_utils/dataset/movielens.py
@@ -136,12 +136,12 @@ DEFAULT_HEADER = (
 
 # Warning and error messages
 WARNING_MOVIE_LENS_HEADER = """MovieLens rating dataset has four columns
-    (user id, movie id, rating, and timestamp), but more than four column headers are provided.
-    Will only use the first four column headers."""
+    (user id, movie id, rating, and timestamp), but more than four column names are provided.
+    Will only use the first four column names."""
 WARNING_HAVE_SCHEMA_AND_HEADER = """Both schema and header are provided.
     The header argument will be ignored."""
 ERROR_MOVIE_LENS_SIZE = "Invalid data size. Should be one of {100k, 1m, 10m, or 20m}"
-ERROR_NO_HEADER = "No header (schema) information"
+ERROR_NO_HEADER = "No header (schema) information. At least user and movie column names should be provided"
 
 
 def load_pandas_df(
@@ -187,13 +187,14 @@ def load_pandas_df(
     size = size.lower()
     if size not in DATA_FORMAT:
         raise ValueError(ERROR_MOVIE_LENS_SIZE)
-    if header is None or len(header) == 0:
-        raise ValueError(ERROR_NO_HEADER)
 
-    if len(header) > 4:
+    if header is None or len(header) < 2:
+        raise ValueError(ERROR_NO_HEADER)
+    elif len(header) > 4:
         warnings.warn(WARNING_MOVIE_LENS_HEADER)
         header = header[:4]
-    movie_col = DEFAULT_ITEM_COL if len(header) < 2 else header[1]
+
+    movie_col = header[1]
 
     with download_path(local_cache_path) as path:
         filepath = os.path.join(path, "ml-{}.zip".format(size)) 
@@ -205,10 +206,6 @@ def load_pandas_df(
         )
 
         # Load rating data
-        if len(header) == 1 and item_df is not None:
-            # MovieID should be loaded to merge rating df w/ item_df
-            header = [header[0], movie_col]
-
         df = pd.read_csv(
             datapath,
             sep=DATA_FORMAT[size].separator,
@@ -268,11 +265,11 @@ def load_item_df(
 
 def _load_item_df(size, item_datapath, movie_col, title_col, genres_col, year_col):
     """Loads Movie info"""
-    item_header = []
-    usecols = []
-    if movie_col is not None:
-        item_header.append(movie_col)
-        usecols.append(0)
+    if title_col is None and genres_col is None and year_col is None:
+        return None
+
+    item_header = [movie_col]
+    usecols = [0]
 
     # Year is parsed from title
     if title_col is not None or year_col is not None:
@@ -290,9 +287,6 @@ def _load_item_df(size, item_datapath, movie_col, title_col, genres_col, year_co
         else:
             item_header.append(genres_col)
             usecols.append(2)  # genres column
-
-    if len(item_header) == 0:
-        return None
 
     item_df = pd.read_csv(
         item_datapath,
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -23,6 +23,15 @@
     pass  # so the environment without spark doesn't break
 
 
+@pytest.fixture
+def tmp_dir(tmp_path_factory):
```
miguelgfierro: a note of naming, the [pytest temporary dir](https://docs.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture) is called `tmpdir` which is almost the same as ours. Maybe we should rename it to avoid users mistake pytest fixture with our fixture. Maybe `temporary_dir`?

tests/smoke/test_movielens.py
```diff
@@ -50,53 +49,50 @@ def test_load_pandas_df(
     title_example,
     genres_example,
     year_example,
+    tmp_dir,
```
miguelgfierro: there is no need to import this since it is a fixture that it is in conftest, see this: https://github.com/Microsoft/Recommenders/blob/master/tests/conftest.py#L4

tests/conftest.py
```diff
@@ -23,6 +23,15 @@
     pass  # so the environment without spark doesn't break
 
 
+@pytest.fixture
```
miguelgfierro: I also like the context manager option, maybe it can be implemented as we are discussing here: https://github.com/Microsoft/Recommenders/issues/701#issuecomment-479436972

tests/smoke/test_movielens.py
loomlike: ohhhh good! will update

tests/conftest.py
loomlike: Let me try out that. looks promising.

tests/conftest.py
loomlike: good suggestion. maybe even more specific, something like `fn_scope_tmpdir`?

loomlike: or does it sound like some kind of function because of some name convention?

gramhagen: i'm missing where tmp_path_factory is coming from?

loomlike: @gramhagen it is pytest's fixture: https://docs.pytest.org/en/latest/tmpdir.html

tests/conftest.py
loomlike: checked and turns out normal fixture does the same thing since newer version of pytest.
> Since pytest-3.0, fixtures using the normal fixture decorator can use a yield statement to provide fixture values and execute teardown code, exactly like yield_fixture in previous versions. Marking functions as yield_fixture is still supported, but deprecated and should not be used in new code.

tests/smoke/test_movielens.py
loomlike: ah, nvm. it means you don't need to do `import your_fixture_module`. You still need to pass the fixture object as an argument. I think the comments contain misleading contents.

tests/conftest.py
loomlike: name changed to just `tmp`, named after linux tmp dir. I picked that for simplicity, but if you think it is not very descriptive, feel free to let me know or give any suggestions. `temporary_dir` looks (visually) unbalanced to me...haha

tests/smoke/test_movielens.py
miguelgfierro: good point, yeah, what you said is more detailed. It would be nice to have it in the code

tests/conftest.py
miguelgfierro: good, so if we use the fixture as you program it using: `with temp_dir: ` is the folder erased after the test function is finished?

tests/conftest.py
miguelgfierro: one question, do we foresee any case where we want to have a directory per (pytest) module or even per (pytest) session?

reco_utils/dataset/movielens.py
```diff
@@ -136,12 +136,12 @@ def item_has_header(self):
...
 WARNING_MOVIE_LENS_HEADER = """MovieLens rating dataset has four columns
-    (user id, movie id, rating, and timestamp), but more than four column headers are provided.
-    Will only use the first four column headers."""
+    (user id, movie id, rating, and timestamp), but more than four column names are provided.
+    Will only use the first four column names."""
 WARNING_HAVE_SCHEMA_AND_HEADER = """Both schema and header are provided.
     The header argument will be ignored."""
 ERROR_MOVIE_LENS_SIZE = "Invalid data size. Should be one of {100k, 1m, 10m, or 20m}"
-ERROR_NO_HEADER = "No header (schema) information"
+ERROR_NO_HEADER = "No header (schema) information. At least user and movie column names should be provided"
```
yueguoguo: Really like these constant string of error/warning message. Do we later want to consolidate this into somewhere in `common/constants`? Or, at least, adopt this practice in other utility codes.

reco_utils/dataset/movielens.py
```diff
@@ -268,11 +265,11 @@ def load_item_df(
 
 def _load_item_df(size, item_datapath, movie_col, title_col, genres_col, year_col):
     """Loads Movie info"""
-    item_header = []
-    usecols = []
-    if movie_col is not None:
-        item_header.append(movie_col)
-        usecols.append(0)
+    if title_col is None and genres_col is None and year_col is None:
```
yueguoguo: Looks like the check now is different from before? The previous checks the `movie_col` and return `None` if it does not exist, whilst now it checks other three columns.

tests/conftest.py
```diff
@@ -23,6 +23,12 @@
...
+def tmp(tmp_path_factory):
```
yueguoguo: This looks nice and clean ❤️

tests/conftest.py
loomlike: We can make fixtures for them when we need, like `sess_tmp` and `module_tmp`.

tests/conftest.py
loomlike: **yes**. Even try/finally does remove the dir too, but using the context manager made the code simpler and give us peace of mind :-)

reco_utils/dataset/movielens.py
loomlike: If we need, sure why not. We can even generalize further, e.g. use `obj.__name__` or to accept names as args.

reco_utils/dataset/movielens.py
loomlike: yeah, made `movie_col` as a default (must-have) column, checking earlier as: `if header is None or len(header) < 2:` (2 is the first two columns, user and movie), because there is no point to load only user ids.

tests/conftest.py
loomlike: kudos to @miguelgfierro @anargyri!

PR #707, opened 2019-04-03
faster ratings

miguelgfierro:
### Description

Refactoring and speeding up the rating evaluation

In my experiments the speed up is around half:
```
%timeit rmse(rating_true_big, rating_pred_big) #the original one
1.05 s ± 17.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit rmse_faster(rating_true_big, rating_pred_big) #the new one
474 ms ± 13.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

The bottleneck was in the function `merge_rating_true_pred` that consisted of 2 operations, a merge and a rename. Curiously, the rename took even more time that the merge. @msalvaris and I just realized that we don't really need the rename so we can pass the name of the true and prediction columns to the sklearn metric

### Checklist:

- [ ] I have added tests.


reco_utils/evaluation/python_evaluation.py
```diff
@@ -95,30 +102,25 @@ def merge_rating_true_pred(
...
-    # Select the columns needed for evaluations
-    rating_true = rating_true[[col_user, col_item, col_rating]]
-    rating_pred = rating_pred[[col_user, col_item, col_prediction]]
-
-    if col_rating == col_prediction:
-        rating_true_pred = pd.merge(
+    suffixes = ["_true", "_pred"]
+    # Apart from merging both dataframes, pd.merge will rename the columns with the suffixes only if the rating
+    # column name of rating_true is the same as the name rating column name in rating_pred
+    rating_true_pred = pd.merge(
```
gramhagen: yes, this is similar to what I did in #699, nice to know the suffixes are only added as needed, at that point it's probably not necessary to even create a separate variable for them.

I think we should go further though, there's nothing else in the merged dataframe that is important other than the truth and prediction columns, so I just extracted them and passed them back as two pd.Series. That way we don't need to keep track of the name of the columns.

miguelgfierro: oh that's a great suggestion and make the code cleaner, I'll make the changes

miguelgfierro: fixed

reco_utils/evaluation/python_evaluation.py
```diff
@@ -94,31 +102,26 @@ def merge_rating_true_pred(
...
+        column_select_pred = col_prediction + suffixes[1]
     else:
-        rating_true_pred = pd.merge(rating_true, rating_pred, on=[col_user, col_item])
-        rating_true_pred.rename(columns={col_rating: DEFAULT_RATING_COL}, inplace=True)
-        rating_true_pred.rename(columns={col_prediction: PREDICTION_COL}, inplace=True)
-
-    return rating_true_pred
+        column_select_true = col_rating
+        column_select_pred = col_prediction
+    return rating_true_pred[column_select_true], column_select_true[column_select_pred]
```
gramhagen: typo on second variable there - should be rating_true_pred[column_select_pred]

miguelgfierro: thanks, fixed now

miguelgfierro:
I don't understand why from time to time github doesn't execute the tests o_o

miguelgfierro:
This can be improved further with memoization: https://github.com/Microsoft/Recommenders/issues/708

PR #712, opened 2019-04-04
Fixing datatype checking issue and adding column checker method

gramhagen:
adding utils to check pandas columns and base datatypes, updating python evaluation to use those in check_column_dtypes decorator

two new methods in reco_utils.dataset.pandas_df_utils:
has_columns(df, columns) - ensures df has the list of columns specified
has_same_base_dtype(df1, df2, columns) - for each entry in columns checks that the base datatype is the same across df1 and df2

python_evaluation is updated to use these two functions

### Related Issues
#648 
#677 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/dataset/pandas_df_utils.py
+++ b/reco_utils/dataset/pandas_df_utils.py
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+import logging
+
 import pandas as pd
 import numpy as np
 
@@ -10,6 +12,9 @@ from reco_utils.common.constants import (
     DEFAULT_RATING_COL,
     DEFAULT_LABEL_COL
 )
+
+
+logger = logging.getLogger(__name__)
 
 
 def user_item_pairs(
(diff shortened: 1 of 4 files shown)


reco_utils/dataset/pandas_df_utils.py
```diff
@@ -315,3 +320,51 @@ def negative_feedback_sampler(
...
     return df_sample
+
+
+def has_columns(df, columns):
+    """Check if DataFrame has necessary columns
+
+    Args:
+        df (pd.DataFrame): DataFrame
+        columns (list(str): columns to check for
+    Returns:
```
miguelgfierro: minor comment, when I was using sphinx for another project, I realized that if we don't have a new line between the end of Args and Returns, sphinx won't parse the documentation correctly. I had to change all my code adding the new line, which was very annoying

reco_utils/evaluation/python_evaluation.py
```diff
@@ -50,32 +46,23 @@ def check_column_dtypes_wrapper(
...
+        Args:
+            rating_true (pd.DataFrame): True data
+            rating_pred (pd.DataFrame): Predicted data
+            col_user (str): column name for user
+            col_item (str): column name for item
+            col_rating (str): column name for rating
+            col_prediction (str): column name for prediction
+        """
+
+        assert has_columns(rating_true, [col_user, col_item, col_rating])
```
miguelgfierro: I think if we are going to use asserts, we should at least add a message for the user. One question here, why using asserts instead of throwing an exception?

miguelgfierro: FYI @msalvaris

msalvaris: It is up to how you think these exceptions will be caught. The best practice is to have a unique exception for different types of errors then functions that try except them can handle them as required. If they are all AssertionError it is going to be hard to do that.

msalvaris: Assertions are fine for scripts and notebooks but not ideal for libs

gramhagen: asserts are easy, there is already logging inside the function to display an error, but i'm fine putting this back to raising a ValueError.

msalvaris: I actually don't agree with the blanket ValueError either tbh. If we are going to use blanket ValueError without any distiction between the errors then asserts are just as good.

reco_utils/dataset/pandas_df_utils.py
gramhagen: ah, that is annoying. ok I'll put the spaces back. I wish black fixed that.

reco_utils/evaluation/python_evaluation.py
```diff
@@ -115,7 +102,7 @@ def merge_rating_true_pred(
     """
     suffixes = ["_true", "_pred"]
     # Apart from merging both dataframes, pd.merge will rename the columns with the suffixes only if the rating
-    # column name of rating_true is the same as the name rating column name in rating_pred
+    # column name of rating_true is the same as the name prediction column name in rating_pred
```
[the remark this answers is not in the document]
gramhagen: How about:
If dataframes have columns with the same names, pd.merge will append the provided suffixes to columns from each dataframe.

reco_utils/evaluation/python_evaluation.py
gramhagen: ahahaha, yes. that was my thinking. but really create a specific exception for this case seems like overkill

reco_utils/evaluation/python_evaluation.py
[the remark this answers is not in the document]
miguelgfierro: that's much better!

Issue #709, opened 2019-04-04
[BUG] Error in mmlspark-lightgbm on Windows


roalexan:
### Description
I'm getting an error when running:

pytest tests/unit -m "notebooks and spark and not gpu"
	1 failed, 5 passed, 130 deselected, 4 warnings in 6478.96 seconds

pytest tests/smoke -m "smoke and spark and not gpu"
	1 failed, 3 passed, 20 deselected, 5 warnings in 1011.78 seconds

More specifically, the errors occur in these tests:

	pytest tests/unit/test_notebooks_pyspark.py -k test_mmlspark_lightgbm_criteo_runs
	pytest tests/smoke/test_notebooks_pyspark.py -k test_mmlspark_lightgbm_criteo_smoke

The full stack traces, which are essentially the same for both, are attached:

[error6.txt](https://github.com/Microsoft/Recommenders/files/3043949/error6.txt)
[error7.txt](https://github.com/Microsoft/Recommenders/files/3043950/error7.txt)

### In which platform does it happen?
Windows DSVM

### How do we replicate the issue?
Create Windows DSVM, clone repo, create reco_pyspark Python environment, and run the tests in the description.

### Expected behavior (i.e. solution)
Should run without any errors.

miguelgfierro:
it looks it is not resolving correctly the mmlspark version: `com.microsoft.ml.spark#mmlspark_2.11;0.16.dev8+2.g6a5318b: not found` it is weird that this is happening on windows. @mhamilton723 any idea where the problem could be?

roalexan:
Another possibility is that it's related to a Java pathing issue - e.g. "The installation directory for Java must not have blanks in the path such as in C:\Program Files" as mentioned [here](https://stackoverflow.com/questions/31841509/pyspark-exception-java-gateway-process-exited-before-sending-the-driver-its-po#36367669). Agree that the "unsolved dependency" might be the real problem.

roalexan:
tested with JAVA_HOME=C:\java\Zulu\zulu-8\ on one of my Windows DSVMs - no space in path - still didn't work.

roalexan:
BTW, I created a new Windows DSVM and this time did 'conda update conda' then 'conda install anaconda'. This brings the conda version to 4.6.9, and apparently fixes the problem where the pip installs weren't working. Unfortunately, this issue persists.

gramhagen:
[...]

PR #714, opened 2019-04-05
fixing access to mml spark repo access from databricks_install in mml…

gramhagen:
### Description
fixing issue pulling in maven repository link in mmlspark lightgbm notebook
i'm still seeing issues locally on windows, but we should try this fix on windows dsvm because my environment is probably not very clean.

### Related Issues
#709 

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/notebooks/02_model/mmlspark_lightgbm_criteo.ipynb
+++ b/notebooks/02_model/mmlspark_lightgbm_criteo.ipynb
@@ -80,7 +80,7 @@
     "    # get the maven coordinates for MML Spark from databricks_install script\n",
     "    from scripts.databricks_install import MMLSPARK_INFO\n",
     "    packages = [MMLSPARK_INFO[\"maven\"][\"coordinates\"]]\n",
-    "    repo = MMLSPARK_INFO[\"maven\"].get(\"repositories\", None)\n",
+    "    repo = MMLSPARK_INFO[\"maven\"][\"repo\"]\n",
     "    spark = start_or_get_spark(packages=packages, repository=repo)\n",
     "    dbutils = None\n",
     "    print(\"MMLSpark version: {}\".format(MMLSPARK_INFO['maven']['coordinates']))\n",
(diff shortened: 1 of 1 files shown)


notebooks/02_model/mmlspark_lightgbm_criteo.ipynb
```diff
@@ -80,7 +80,7 @@
     "    # get the maven coordinates for MML Spark from databricks_install script\n",
     "    from scripts.databricks_install import MMLSPARK_INFO\n",
     "    packages = [MMLSPARK_INFO[\"maven\"][\"coordinates\"]]\n",
-    "    repo = MMLSPARK_INFO[\"maven\"].get(\"repositories\", None)\n",
+    "    repo = MMLSPARK_INFO[\"maven\"][\"repo\"]\n",
```
miguelgfierro: good catch, however, we have to add `.get("repo", None)`, if not spark function will break if we use something similar to what we have originally, where there was only coordinates.

miguelgfierro:
it happened again that the tests hasn't triggered, this is weird

gramhagen:
I pushed several prs in close succession, i wonder if there's a limit to the queue of pending tests or some kind of time limit before the test status is no longer accessible?

notebooks/02_model/mmlspark_lightgbm_criteo.ipynb
gramhagen: i wanted to force an error in that case, should we carry forward if they get out of synch?
also, fwiw if you use dict.get(), None is automatically provided as the default. [reference](https://docs.python.org/3/library/stdtypes.html#dict.get)

miguelgfierro: > if you use dict.get(), None is automatically provided as the default. 

yeah, so the idea is if we are using the old dictionary, that doesn't have repo, the spark loader doesn't fail:

```
MMLSPARK_INFO = {"maven": {"coordinates": "Azure:mmlspark:0.16"}}
```
I guess that the dictionary that we have now is temporal:
```
MMLSPARK_INFO = {"maven": {"coordinates": "com.microsoft.ml.spark:mmlspark_2.11:0.16.dev8+2.g6a5318b",
                          "repositories": "https://mmlspark.azureedge.net/maven"}
                }
```

> i wanted to force an error in that case

don't understand very well what you are thinking, can you elaborate?

gramhagen: I see so, you're just saying we shouldn't bother updating this notebook if we drop the repo piece in the future (which we're planning on doing), i'm fine with that.

PR #715, opened 2019-04-05
Benchmark partial

miguelgfierro:
### Description

Benchmark of some algos in Movielens, this was a request from @nikhilrj and @wutaomsft 

As discussed, this will be temporal, in the future we will use directly the notebooks to execute the benchmarks

### Related Issues

https://github.com/Microsoft/Recommenders/issues/464

### Checklist:

- [ ] I have added tests.


anargyri:
Do we want to include some tuning for the benchmarking (either as part of this code or as a preprocessing step)?

miguelgfierro:
@anargyri that's a good point, I guess in the benchmark we should have some "optimal" parameters precomputed via hyperparameter tuning, so we can claim that the results of the benchmark are reasonable enough.

benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,297 @@
...
+                                        col_prediction=PREDICTION_COL)
+    return {
+        "RMSE": rating_eval.rmse(),
+        "MAE": rating_eval.mae(),
+        "R2": rating_eval.exp_var(),
+        "Explained Variance": rating_eval.rsquared()
+    }
+    
+    
+def ranking_metrics_pyspark(test, predictions, k=10):
```
yueguoguo: Can we just directly call the method?

benchmarks/movielens.ipynb
```diff
@@ -0,0 +1,942 @@
+{
```
yueguoguo: Maybe hyperparater tuning can be incorporated given that we have already developed notebooks for some (unfortunately not all) algorithms.

benchmarks/movielens.ipynb
```diff
@@ -0,0 +1,942 @@
...
+       "      <td>0.175191</td>\n",
+       "      <td>0.160295</td>\n",
+       "      <td>0.053431</td>\n",
+       "    </tr>\n",
+       "  </tbody>\n",
+       "</table>\n",
+       "</div>"
+      ],
+      "text/plain": [
+       "    Data    Algo   K      Train time     Rating time      RMSE       MAE  \\\n",
```
yueguoguo: Maybe better to rename "Rating time" and "Ranking time" to "Predicting time" and "Recommending time", respectively?

benchmarks/movielens.ipynb
```diff
@@ -0,0 +1,942 @@
...
+       "1   100k     als  10  0:00:04.525615  0:00:00.037825  0.961177  0.749250   \n",
```
yueguoguo: It's interesting to see that SAR does not perform better than the others, which is not in line with the results we had previously. It would be great to understand more about this.

benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,297 @@
+import sys
```
yueguoguo: Many of the utilities are used only in the context of benchmarking (they wrap our existing utilities), and some of them have very similar functionalities. Do we want to combine in a way to enhance the readability?

benchmarks/movielens.ipynb
yueguoguo: Just saw @anargyri comment 😂

miguelgfierro: not sure if you guys mean to have the tuning in this same notebook or to perform the tuning outside and then add the optimal parameters here?

benchmarks/movielens.ipynb
miguelgfierro: yeah, this is interesting, in this notebook we are using stratified splitting. The reason to use that, is because it is the only one (afaik) that fastai accepts.

benchmarks/benchmark_utils.py
miguelgfierro: I was thinking the same, but I wasn't sure to be honest how to extract the utilities to not create a black box for the user. I think we need to find a middle ground between adding too much content and having only few high level functions (load, train, evaluate)

benchmarks/movielens.ipynb
anargyri: The former requires writing a python script for each method. It is a lot of work and will take a long time to run. Probably the latter is more reasonable but can be done gradually, not all of it in this PR. You could start from ALS and SVD since we have automated tuning for those.

miguelgfierro: that makes sense

gramhagen:
I thought from #589 we were going to have separate notebooks / scripts for each algorithm? did i misinterpret what was meant there for scaling this out as we add new algorithms?

benchmarks/movielens.ipynb
loomlike: that is interesting... does stratified-splitting splits chronologically or random within a user's log?

benchmarks/movielens.ipynb
```diff
@@ -0,0 +1,942 @@
...
+    "                                                min_rating=1, \n",
+    "                                                filter_by=\"item\", \n",
+    "                                                col_user=USER_COL, \n",
+    "                                                col_item=ITEM_COL\n",
+    "                                                )\n",
+    "    print(\"Train set size: {}\".format(df_train.shape))\n",
+    "    print(\"Test set size: {}\".format(df_test.shape))\n",
+    "   \n",
+    "    # Loop through the algos\n",
+    "    for algo in algorithms:\n",
```
loomlike: Each iteration takes long and all are independent, which is a perfect scenario for parallel / distributed processing. Can we use Azure remote compute for this or (if refactoring is not trivial or already in the next todo list), maybe simply use multiprocessing at least for CPU algos?

benchmarks/movielens.ipynb
```diff
@@ -0,0 +1,942 @@
...
+       "      <td>0.140403</td>\n",
+       "      <td>0.056657</td>\n",
+       "    </tr>\n",
+       "    <tr>\n",
+       "      <th>6</th>\n",
+       "      <td>1m</td>\n",
+       "      <td>als</td>\n",
+       "      <td>10</td>\n",
+       "      <td>0:00:03.275541</td>\n",
+       "      <td>0:00:00.013025</td>\n",
```
loomlike: Just double check if als' rating and ranking time were correctly measured. Isn't the transform lazy operation? 

Also would be better to use more 'human-friendly' time format, maybe 'secs'?

miguelgfierro:
> I thought from #589 we were going to have separate notebooks / scripts for each algorithm? did i misinterpret what was meant there for scaling this out as we add new algorithms?

you are right, this was requested by Tao for having an initial benchmark until we start developing the benchmark through the notebooks

benchmarks/benchmark_utils.py
miguelgfierro: Hey guys, I've been looking at this. I think it would be too much to add all the metrics code here, right now this is the code in functions:
```
def rating_metrics_pyspark(test, predictions):
    rating_eval = SparkRatingEvaluation(test, 
                                        predictions, 
                                        col_user=USER_COL, 
                                        col_item=ITEM_COL, 
                                        col_rating=RATING_COL, 
                                        col_prediction=PREDICTION_COL)
    return {
        "RMSE": rating_eval.rmse(),
        "MAE": rating_eval.mae(),
        "R2": rating_eval.exp_var(),
        "Explained Variance": rating_eval.rsquared()
    }
    
    
def ranking_metrics_pyspark(test, predictions, k=10):
    rank_eval = SparkRankingEvaluation(test, 
                                       predictions, 
                                       k=k, 
                                       col_user=USER_COL, 
                                       col_item=ITEM_COL, 
                                       col_rating=RATING_COL, 
                                       col_prediction=PREDICTION_COL, 
                                       relevancy_method="top_k")
    return {
        "MAP": rank_eval.map_at_k(),
        "nDCG@k": rank_eval.ndcg_at_k(),
        "Precision@k": rank_eval.precision_at_k(),
        "Recall@k": rank_eval.recall_at_k()
    }
    
    
def rating_metrics_python(test, predictions):
    cols = {
        "col_user": USER_COL, 
        "col_item": ITEM_COL, 
        "col_rating": RATING_COL, 
        "col_prediction": PREDICTION_COL
    }
    return {
        "RMSE": rmse(test, predictions, **cols),
        "MAE": mae(test, predictions, **cols),
        "R2": rsquared(test, predictions, **cols),
        "Explained Variance": exp_var(test, predictions, **cols)
    }
    
    
def ranking_metrics_python(test, predictions, k=10):
    cols = {
        "col_user": USER_COL, 
        "col_item": ITEM_COL, 
        "col_rating": RATING_COL, 
        "col_prediction": PREDICTION_COL
    }
    return {
        "MAP": map_at_k(test, predictions, k=k, **cols),
        "nDCG@k": ndcg_at_k(test, predictions, k=k, **cols),
        "Precision@k": precision_at_k(test, predictions, k=k, **cols),
        "Recall@k": recall_at_k(test, predictions, k=k, **cols)
    }
```
then apart from that we need to select which algos are spark and which cpu. If we add all these code in the loop is going to be very big. What do you guys think?

benchmarks/movielens.ipynb
miguelgfierro: added in the next commit

benchmarks/benchmark_utils.py
anargyri: Are you asking about the size of the code or about computational time? The latter should be faster now with the LRU cache.

benchmarks/movielens.ipynb
```diff
@@ -0,0 +1,942 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Model parameters\n",
+    "EPOCHS_CPU = 30\n",
```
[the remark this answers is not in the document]
miguelgfierro: I don't understand very well what you are asking, can you put an example?

benchmarks/movielens.ipynb
```diff
@@ -0,0 +1,942 @@
...
+     "output_type": "stream",
+     "text": [
+      "4.93MB [00:01, 3.87MB/s]                            \n"
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Size of Movielens 100k: (100000, 4)\n",
```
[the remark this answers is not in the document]
miguelgfierro: removing that in the next commit

benchmarks/movielens.ipynb
miguelgfierro: yeah I think the next step on this would be to do that, if we do the benchmarks through notebooks we can do them in parallel. It would be nice. We also should start thinking on how to perform the tests in parallel, which is blocking us a little bit

benchmarks/benchmark_utils.py
miguelgfierro: size of the code, I think adding all the metrics in the main loop will be too much

benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,297 @@
...
+        StructField(ITEM_COL, IntegerType()),
+        StructField(RATING_COL, FloatType()),
+        StructField(TIMESTAMP_COL, LongType()),
+    )
+    )
+    spark = start_or_get_spark()
+    return spark.createDataFrame(train, schema)
+
+
+def train_als(params, data):
```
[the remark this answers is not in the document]
miguelgfierro: the wrappers are just simple functions to surface the code based of factories. I'm sure there are other ways. The idea in these functions is to split the functionality for each part, following single responsibility

benchmarks/benchmark_utils.py
anargyri: You could also define RATING_METRICS and RANKING_METRICS as dicts and then use for loops (as in the tuning notebooks). It would remove some stuff that repeats.

miguelgfierro: this is how it is done, you can see this in cell 13:
```
rating_evaluator = {
    "als": lambda test, predictions: rating_metrics_pyspark(test, predictions),
    "svd": lambda test, predictions: rating_metrics_python(test, predictions),
    "fastai": lambda test, predictions: rating_metrics_python(test, predictions)
}
    
    
ranking_evaluator = {
    "als": lambda test, predictions, k: ranking_metrics_pyspark(test, predictions, k),
    "sar": lambda test, predictions, k: ranking_metrics_python(test, predictions, k),
    "svd": lambda test, predictions, k: ranking_metrics_python(test, predictions, k),
    "fastai": lambda test, predictions, k: ranking_metrics_python(test, predictions, k),
    "ncf": lambda test, predictions, k: ranking_metrics_python(test, predictions, k),
}
```
Le and Jun Ki were proposing to call the metrics directly, but I think it is too much code

anargyri: You could reduce some repetition here too using two dictionaries of the form
`{"als": rating_metrics_pyspark, "svd": rating_metrics_python, "fastai": rating_metrics_python}`
`{"als": ranking_metrics_pyspark, "sar": ranking_metrics_python, "svd": ranking_metrics_python, "fastai": ranking_metrics_python, "ncf": ranking_metrics_python}` etc.

anargyri: It is good to have these dictionaries anyway, because they can be reused by many of the notebooks we have written.

benchmarks/movielens.ipynb
miguelgfierro: changed in the next commit

benchmarks/movielens.ipynb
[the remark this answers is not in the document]
miguelgfierro: changed all to 10

benchmarks/benchmark_utils.py
miguelgfierro: we can review this in a future issue

benchmarks/movielens.ipynb
miguelgfierro: I think random: https://github.com/Microsoft/Recommenders/blob/master/reco_utils/dataset/python_splitters.py#L202

benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,297 @@
...
+    return preds, t
+
+
+def recommend_k_als(model, test, train):
+    with Timer() as t:
+        # Get the cross join of all user-item pairs and score them.
+        users = train.select(USER_COL).distinct()
+        items = train.select(ITEM_COL).distinct()
+        user_item = users.crossJoin(items)
+        dfs_pred = model.transform(user_item)
```
gramhagen: wouldn't it be better to filter out seen items before prediction?

also it would be awesome to extract thisout as a function that can be used by all algorithms: get_all_user_items(data, user_col, item_col, remove_seen=True)

benchmarks/benchmark_utils.py
miguelgfierro: hey @anargyri, I added a col dictionary, will be there in the next commit. I left the metrics as:
```
def rating_metrics_python(test, predictions):
    return {
        "RMSE": rmse(test, predictions, **COL_DICT),
        "MAE": mae(test, predictions, **COL_DICT),
        "R2": rsquared(test, predictions, **COL_DICT),
        "Explained Variance": exp_var(test, predictions, **COL_DICT)
    }
```
I think this is a good compromise between what Jun Ki and Le were thinking and having the metrics as a dict comprehension. We could argue that this is more explicit

benchmarks/benchmark_utils.py
miguelgfierro: Moving this to a future issue since there was a similar ask from Le https://github.com/Microsoft/Recommenders/issues/733

benchmarks/benchmark_utils.py
loomlike: @miguelgfierro  Hey Miguel, sorry for un-resolving the conversation, but what do you think about adding functions "all_rating_metrics" and "all_ranking_metrics" to our evaluation module? All of our notebooks do that anyway. They will be the same functions as these that you have in benchmark_utils.

benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def train_svd(params, data):
+    model = surprise.SVD(**params)
+    with Timer() as t:
+        model.fit(data)
+    return model, t
+
+
+def predict_svd(model, test):
```
anargyri: We have some utilities for doing these with Surprise:
https://github.com/Microsoft/Recommenders/blob/master/reco_utils/recommender/surprise/surprise_utils.py

benchmarks/benchmark_utils.py
miguelgfierro: I'm not super in favor tbh, I think it is the same reason that Le said, we are trying to do things explicitely over implicitely. In general situations like the notebooks, we have only 4 lines of code:
```
rmse(test, predictions, **cols),
mae(test, predictions, **cols),
rsquared(test, predictions, **cols),
exp_var(test, predictions, **cols)
```
This is not too much and I think it is nice to show it explicitly as opposite of having 1 line of implicit `rating_metrics`.

In the case of this notebook, where we have to add all the metrics at the same time, I think it could make sense to have it, just because if not it gets super messy. 

Does that make sense?

anargyri: Makes sense, I think it is flexible how much one may want to refactor a notebook, since it is partly development and partly demonstration.

benchmarks/benchmark_utils.py
miguelgfierro: very good catch, fixed in the next commit

PR #717, opened 2019-04-06
Spark SETUP doc update

loomlike:
### Description
* "Install All environments" actually installs one env, pyspark-gpu. Updated setup doc accordingly.
* Backup PYSPARK_HOME before unset and recover when deactivate the env, so that users don't need to restart the session.
* Spark env setup to be more visible: That section is related to both reco_pyspark and reco_full. So pulled out from spark-install section to end of the env installation.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -41,7 +41,7 @@ We install the dependencies with Conda. As a pre-requisite, we want to make sure
 
 ```{shell}
 conda update conda -n root
-conda update anaconda
+conda update anaconda        # use 'conda install anaconda' if the package is not installed
 ```
 
 We provide a script, [generate_conda_file.py](scripts/generate_conda_file.py), to generate a conda file, depending of the environment we want to use. This will create the environment using the Python version 3.6 with all the correct dependencies.
--- a/SETUP.md
+++ b/SETUP.md
@@ -32,7 +32,7 @@ Depending on the type of recommender system and the notebook that needs to be ru
 
 * Machine running Linux, Windows Subsystem for Linux ([WSL](https://docs.microsoft.com/en-us/windows/wsl/about)) or macOS
 * Anaconda with Python version >= 3.6.
-  * This is pre-installed on Azure DSVM, for local setup [Miniconda](https://docs.conda.io/en/latest/miniconda.html) is a quick way to get started.
+  * This is pre-installed on Azure DSVM. For local setup [Miniconda](https://docs.conda.io/en/latest/miniconda.html) is a quick way to get started.
 * [Apache Spark](https://spark.apache.org/downloads.html) (this is only needed for the PySpark environment).
 
 ### Dependencies setup
--- a/SETUP.md
+++ b/SETUP.md
@@ -23,9 +23,8 @@ This document describes how to setup all the dependencies to run the notebooks i
 
 Depending on the type of recommender system and the notebook that needs to be run, there are different computational requirements. Currently, this repository supports the following environments:
 
-* Python CPU
-* Python GPU
-* PySpark
+* Python CPU / GPU
+* PySpark CPU / GPU
 
 ## Setup guide for Local or DSVM
 
@@ -74,7 +73,7 @@ Assuming that you have a GPU machine, to install the Python GPU environment, whi
 </details>
 
 <details>
-<summary><strong><em>PySpark environment</em></strong></summary>
+<summary><strong><em>PySpark CPU environment</em></strong></summary>
 
 To install the PySpark environment, which by default installs the CPU environment:
 
@@ -89,7 +88,7 @@ Additionally, if you want to test a particular version of spark, you may pass th
 </details>
 
 <details>
-<summary><strong><em>All environments</em></strong></summary>
+<summary><strong><em>PySpark GPU environment</em></strong></summary>
 
 To install the PySpark GPU environment:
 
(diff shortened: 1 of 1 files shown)


SETUP.md
```diff
@@ -23,17 +23,16 @@ This document describes how to setup all the dependencies to run the notebooks i
...
+* PySpark CPU / GPU
 
 ## Setup guide for Local or DSVM
 
 ### Requirements
 
 * Machine running Linux, Windows Subsystem for Linux ([WSL](https://docs.microsoft.com/en-us/windows/wsl/about)) or macOS
 * Anaconda with Python version >= 3.6.
-  * This is pre-installed on Azure DSVM, for local setup [Miniconda](https://docs.conda.io/en/latest/miniconda.html) is a quick way to get started.
+  * This is pre-installed on Azure DSVM. For local setup [Miniconda](https://docs.conda.io/en/latest/miniconda.html) is a quick way to get started.
```
yueguoguo: Perhaps change "This is pre-installed on Azure DSVM" to "This is pre-installed on Azure DSVM such that one can run the following steps directly". The reasons is previously someone tries the repo after installing miniconda manually and that leads to some path issue.

SETUP.md
```diff
@@ -86,37 +85,43 @@ Additionally, if you want to test a particular version of spark, you may pass th
...
-```bash
-#!/bin/sh
-unset PYSPARK_PYTHON
-unset PYSPARK_DRIVER_PYTHON
-```
 </details>
 
 <details>
-<summary><strong><em>All environments</em></strong></summary>
+<summary><strong><em>PySpark GPU environment</em></strong></summary>
```
yueguoguo: How about "PySpark & GPU environment" given that it actually means both of the two are supported in one conda env? PySpark GPU sounds like running GPU in a Spark cluster though I am not sure if it is doable 😂

SETUP.md
loomlike: make sense

SETUP.md
```diff
@@ -23,17 +23,16 @@ This document describes how to setup all the dependencies to run the notebooks i
 
 Depending on the type of recommender system and the notebook that needs to be run, there are different computational requirements. Currently, this repository supports the following environments:
 
-* Python CPU
```
miguelgfierro: I have a similar concern to @yueguoguo, having `PySpark CPU / GPU` seems to indicate that we are doing pyspark on GPU. What about something like this:
* Python CPU
* Python GPU
* PySpark
* PySpark & Python GPU

SETUP.md
```diff
@@ -42,7 +41,7 @@ We install the dependencies with Conda. As a pre-requisite, we want to make sure
 
 ```{shell}
 conda update conda -n root
-conda update anaconda
```
miguelgfierro: about this @roalexan had problems in windows, we were planning to remove these two lines:

```
conda update conda -n root
conda update anaconda 
```
and then add a maximum conda version to the conda file generator. Robert do you have the info about this?

SETUP.md
```diff
@@ -86,37 +85,43 @@ Additionally, if you want to test a particular version of spark, you may pass th
...
-unset PYSPARK_DRIVER_PYTHON
-```
 </details>
 
 <details>
-<summary><strong><em>All environments</em></strong></summary>
+<summary><strong><em>PySpark & GPU environment</em></strong></summary>
 
-To install all three environments:
+To install the PySpark GPU environment:
```
miguelgfierro: This is again confusing, it looks that we are doing Spark on GPU :-) 
Maybe we say in the title `PySpark & Python GPU environment` and then in the description below state clearly that the user can run PySpark **OR** Python GPU

SETUP.md
```diff
@@ -86,37 +85,43 @@ Additionally, if you want to test a particular version of spark, you may pass th
...
+> ```bash
+> #!/bin/sh
+> unset PYSPARK_PYTHON
+> unset PYSPARK_DRIVER_PYTHON
+> export SPARK_HOME=$SPARK_HOME_BACKUP
+> unset SPARK_HOME_BACKUP
+> ```
+
+
 ### Register the conda environment as a kernel in Jupyter
```
miguelgfierro: I think we need to clarify this. We only need to register the kernels for Jupyterhub, not normal jupyter. For some reason many people start a jupyter kernel in the DSVM instead of using the Jupyterhub that comes by default in the DSVM. 

We can say that if you are using the DSVM, you can access a fully functional jupyterhub by going to `IP_or_DNS:8000`. In order to install any of the reco environments to appear automatically in Jupyterhub, you have to type ...

SETUP.md
yueguoguo: Would be better to add a "-" between Python and CPU/GPU. What do you @Microsoft/miguelgfierro @loomlike think?

loomlike: Make sense. I think Miguel's suggestion (4 bullets) better explains the actual environments. I would add env name as well. Please check new changes.

loomlike: In that sense, removing the bullet points look more simple to me.
> Currently, this repository supports **Python CPU**, **Python GPU** and **PySpark**.

SETUP.md
loomlike: Adding to conda file generator sounds a simple & good solution.
For windows, I didn't have any issues to install the environments except `conda update anaconda` which threw the package not installed error. (I ran the script on Administrator-terminal)

SETUP.md
loomlike: Doesn't the `python -m ipykernel install ...` set the kernel for both Jupyterhub and Jupyter?

loomlike:
Thanks for all the comments. Revised based on the comments.

In addition to them, I would like to unfold one of the four environments setups for better visibility, while leave other three environment-setups as it is (advanced readings). Check and let me know if this looks good or prefer the previous or have any better ideas!

SETUP.md
loomlike: Btw, please remove this part once we have updated conda file generator.

PR #718, opened 2019-04-08
Docker Support

yueguoguo:
### Description
Initialize a PR of Docker support for PySpark environment. This PR is for discussion with team to brainstorm and optimize Docker support for the repo.

**NOTE** 
* I did not use conda yaml file in the repo to build conda env because the base image from https://github.com/jupyter/docker-stacks handles Jupyter kernel separately and has already installed many packages that exist in our yaml file.
* To make the image light weighted, the conda/pip packages duplicated in both the base image and those in our yaml file are removed

**TODO**
* Create pre-built image and publish in Docker hub
* Test running the Docker container - I see [this](https://github.com/jupyter/docker-stacks/tree/master/test) a good example. Maybe we want to adopt it? 
* Finish the rest of the three Docker images, i.e., CPU and GPU

**HOW-TO**
A sample image is created and publicitized on my own Docker hub account (we can and should create one for the team later on)
In a Linux terminal or Windows powershell (assume Docker is pre-installed in the machine)
```{shell}
docker pull yueguoguo/reco_pyspark:latest
docker run --rm -p 8888 yueguoguo/reco_pyspark
```
Open browser and go to localhost:8888 with the token generated in the above run of the image. 

**UPDATE**
**2019-04-08** 
* The pre-built image refers to a branch in the repo where there are only notebooks executable in the environment. For example, in the PySpark environment, the deep learning notebooks which are supposed to run in a GPU environment, are removed, because the Python packages for these notebooks are not installed for the light-weight consideration. 

**2019-06-21** 
* "one to bind all". The same Dockerfile can be used for building an image with different environment, i.e., CPU, GPU, and Spark. This can be done by using the specific build args, i.e., `cpu`, `gpu`, and `pyspark`, respectively
* `SETUP.md` is updated accordingly
* Master branch of the repo will be cloned

### Related Issues
Discussed in #687 

### Checklist:

- [ ] I have added tests.


reco_utils/docker/pyspark/Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: are you planning to do 3 PRs: one for each env? or are you planning to use this one?

reco_utils/docker/pyspark/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    "azureml-sdk[notebooks,tensorboard,contrib]==1.0.18" \
+    "idna==2.7" \
+    "memory-profiler>=0.54.0" \
+    "papermill>=0.15.0" \
+    "pydocumentdb>=2.3.3" \
+    "tqdm==4.31.1"
+
+RUN fix-permissions $CONDA_DIR && \
+    fix-permissions /home/$NB_USER && \
+    fix-permissions $RECO_PATH
```
miguelgfierro: I think it would be nice to have our notebooks in docker, similarly to what MMLSpark guys are doing: https://github.com/Azure/mmlspark/blob/master/tools/docker/Dockerfile

reco_utils/docker/pyspark/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ARG RECO_PATH=$WORK_PATH/Recommenders
+
+USER root
+
+# Clone repo
+RUN git clone https://github.com/Microsoft/Recommenders.git && \
+    cp -rf Recommenders $WORK_PATH 
+
+# Install the needed packages
+RUN conda install --quiet --yes \
```
miguelgfierro: I think we should have a way to add the libraries programmatically, I've seen that other libs use scripts: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/Dockerfile.gpu#L24. We could use python to get the libs

reco_utils/docker/pyspark/Dockerfile
miguelgfierro: This is great, we should have the images in dockerhub, maybe even we can have the official release and a dev release :-)

SETUP.md
```diff
@@ -269,3 +269,29 @@ Additionally, you must install the [spark-cosmosdb connector](https://docs.datab
...
 
 </details>
+
+## Setup guide for Docker
+
+Docker images of the repo for different environment are provided. Dockerfiles can be found at `Recommenders/reco_utils/docker` if one would like to build Docker images on his/her own.
+
+Assuming Docker is pre-installed and configured (**NOTE** `docker` command is already available in the Azure Data Science Virtual Machine. For installation of Docker in on a Linux machine, check instructions [here](https://docs.docker.com/install/)), for example, a PySpark Docker image can be built by the following command line
+>   ```{shell}
+>   cd Recommenders/reco_utils/docker/pyspark
```
miguelgfierro: I've been looking at where other libs put the docker files, both [TF](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/ci_build) and [Pytorch](https://github.com/pytorch/pytorch/tree/master/tools/docker) put it under the folder tools. 

Maybe we should consider renaming scripts to tools, this was discussed several times @bethz

reco_utils/docker/pyspark/Dockerfile
yueguoguo: Ah currently we can directly run the notebooks in the browser after running a container of the image

reco_utils/docker/pyspark/Dockerfile
yueguoguo: It is a good idea for better management.

SETUP.md
yueguoguo: +1 on renaming

reco_utils/docker/pyspark/Dockerfile
msalvaris: Could we not pull the dependencies from the github repo? That way we won't have to change versions in two locations.

reco_utils/docker/pyspark/Dockerfile
msalvaris: You can also add a juptyer_notebook_config.py for default configurations

reco_utils/docker/pyspark/Dockerfile
yueguoguo: Will investigate this again. Thanks @Microsoft/miguelgfierro and @msalvaris 

The reason for not using our yaml is that the base image has some pip/conda dependencies installed which are duplicated in our case. Another reason is that the base image has its own kernel installation and registration style which is different from our approach...

reco_utils/docker/pyspark/Dockerfile
msalvaris: Yea, the only thing you have to be mindful of is that we have to be careful of what we are allowed to redistribute. You may have to talk with CELA :(

SETUP.md
```diff
@@ -269,3 +269,29 @@ Additionally, you must install the [spark-cosmosdb connector](https://docs.datab
    7. Restart the cluster.
 
 </details>
+
+## Setup guide for Docker
+
+Docker images of the repo for different environment are provided. Dockerfiles can be found at `Recommenders/reco_utils/docker` if one would like to build Docker images on his/her own.
```
gramhagen: suggested wording change and link addition (if we land on a link, maybe leave it out till then):

Docker images of the repo for different environments are available (provide link). To build Docker images yourself use the Dockerfiles in `reco_utils/docker`

SETUP.md
gramhagen: this could just be 
```docker build -t <image_name>/<tag> -f reco_utils/docker/pyspark```

File: reco_utils/docker/pyspark/Dockerfile (hunk 3 above)
Comment (gramhagen) [reply reco_utils/docker/pyspark/Dockerfile]: it would be great to adjust the conda file generator script to accept a docker flag that would handle the adjusting the yaml file as needed for these cases.

File: reco_utils/docker/pyspark/Dockerfile [hunk 6]
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+ARG BASE_CONTAINER=jupyter/pyspark-notebook
```
gramhagen: are you intending to allow these ARGs to be changed at build time?

reco_utils/docker/pyspark/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+LABEL maintainer="Microsoft Recommender Project <RecoDevTeam@service.microsoft.com>"
+
+ARG WORK_PATH=/home/$NB_USER/work
+ARG RECO_PATH=$WORK_PATH/Recommenders
+
+USER root
+
+# Clone repo & discard notebooks that cannot be run in this environment
+RUN git clone --single-branch --branch only_pyspark https://github.com/Microsoft/Recommenders.git && \
```
gramhagen: why not set working directory to work_path prior to git clone so you don't need to copy the directory?

gramhagen: it seems like retaining seperate branches for notebooks that are relevant for different environments will be hard to maintain. Do we need to remove them? Right now we're assuming the user understands and has set up the correct environment (and ideally that environment is documented clearly in each notebook - this may not be the case but should be)

SETUP.md
loomlike: +1 on the name `tools`

reco_utils/docker/pyspark/Dockerfile
yueguoguo: @gramhagen having some issues installing the conda dependencies re AML but agree this is a desirable approach.

loomlike:
@yueguoguo @miguelgfierro What's the status of this?

Also wonder about if we want to have non-pyspark (gpu or cpu) docker support. I like to have that for two reasons:
1. No hassle to install env (I heard one of our Garage intern says having some issues of installing env and running notebooks).
2. Make it easy to use container orchestration solutions or even AzureML

miguelgfierro:
I guess we can have 3 versions (cpu, gpu, spark), not sure if we want to also have the full environment

yueguoguo:
@loomlike I am revisiting this part these days. 

Quick ones re your ask:
> No hassle to install env (I heard one of our Garage intern says having some issues of installing env and running notebooks).
The Docker support in this PR is based on the Jupyter Dockerfile so it is for launching a Jupyter notebook on whatever desired platform where Docker is run, to run the notebooks in our repo, without upfront environment setup. So this scenario will be nice fit.

>  Make it easy to use container orchestration solutions or even AzureML
Need modification to better support this scenario.

gramhagen:
Another thought here. It might actually be preferable to avoid images that have libraries pre installed as that will introduce potential differences from our existing setup.

yueguoguo:
Currently, using the setup instruction directly in the docker image build yields an image with size of 22GB...

miguelgfierro:
> Currently, using the setup instruction directly in the docker image build yields an image with size of 22GB...


22? that's crazy. Do you know what is making the image so big?

gramhagen:
I was looking into building a docker image for the reco_base env. and then thought I would build off what you have here to try out working from minimal ubuntu image. The final spark image is only 6.5GB which seems reasonable. I did have to make some changes to the spark test setup though (something bad happens if you allocate too little memory in docker).

I put some of the work I've been doing here: https://github.com/microsoft/recommenders/tree/gramhagen/spark_test_fix

there's a tools/docker folder with two images. Maybe we can consolidate these two branches?

yueguoguo:
@gramhagen FYI I updated the Dockerfile to reduce the dependencies of the packages in the built image. It also allows users to specify argument value for building image with different environment. 

Sure let's see how to merge the work together.

SETUP.md
miguelgfierro: @yueguoguo  is the plan to finally create the `tools` folder and add the docker there?

docker/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    "pydocumentdb>=2.3.3" \
+    "tqdm==4.31.1"
+
+# Install conda & pip packages
+RUN if ["$ENVIRONMENT" = "pyspark"]; then \
+        # Install the needed packages for pyspark environment
+        conda install --quiet --yes \
+        "pyarrow>=0.8.0" \
+        "pyspark==2.3.1"; \
+    elif ["$ENVIRONMENT" = "tensorflow"]; then \
```
miguelgfierro: maybe for consistency we should say "gpu" instead of  "tensorflow"

miguelgfierro:
do we have a channel on dockerhub where we can host the images? maybe we should create a pipeline for building the docker images of the releases and another one for master

gramhagen:
Hosting images is a nightmare because every installed package / library can have different licenses. It's better to just provide the dockerfiles.

miguelgfierro:
> Hosting images is a nightmare because every installed package / library can have different licenses. It's better to just provide the dockerfiles.

What is the problem with the licenses? In MMLSpark they have a docker image under the microsoft account https://hub.docker.com/r/microsoft/mmlspark

miguelgfierro:
> If any package included has a strict copyleft license then arguably the entire image would be distributed under that license, not to mention if licenses or restrictions on usage are in conflict with each other. We should figure out what we can do in terms of redistribution, hopefully it's not a problem but the alternative of making the dockerfile available allows for users to properly manage packages / license issues.

Yeah, makes sense, maybe it would be good to check with CELA, in case there is any problem with the licenses. I'll send an email

docker/Dockerfile
yueguoguo: Hmmmm yea fair point. The reason is that the name of the base image uses `tensorflow` instead of `gpu`. Conditional load of image is a little bit messy in Dockerfile.

reco_utils/docker/pyspark/Dockerfile
yueguoguo: Now the base image is fixed. Previously I was thinking having different base for different images.

docker/Dockerfile
yueguoguo: @miguelgfierro I moved the Dockerfile to the top of the repo, as I noticed that quite a few repos put Dockerfile at the top. E.g., pytorch, horovod, etc.

simonzhaoms: @yueguoguo I didn't come across this situation before.  If you don't like if-else, maybe you can use case:

```
RUN case "$ENVIRONMENT" in \
    "pyspark") \
        # Install the needed packages for pyspark environment
        conda install --quiet --yes \
            "pyarrow>=0.8.0" \
            "pyspark==2.3.1" \
	;; \
    "tensorflow") \
        # Install the needed packages for GPU environment
        conda install --quiet --yes \
            "numba>=0.38.1" \
            "pytorch>=1.0.0" \
            "tensorflow-gpu==1.12.0"
	;; \
    esac
```

But there are still duplicated `conda install --quiet --yes`.

miguelgfierro: > @miguelgfierro I moved the Dockerfile to the top of the repo, as I noticed that quite a few repos put Dockerfile at the top. E.g., pytorch, horovod, etc.

That's true. I also saw Lightgbm does the same. TF, on the contrary has it in an internal folder called [tools](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools). 

I thought that `tools` was the "wild card" to put things that are not tests, utilities and examples. My only concern is that we should be careful not to make the first level of folders too complex. 

But again, up to you, there is more evidence for the first approach

gramhagen:
> @gramhagen FYI I updated the Dockerfile to reduce the dependencies of the packages in the built image. It also allows users to specify argument value for building image with different environment.
> 
> Sure let's see how to merge the work together.

I think maybe best to decouple, I'll remove the docker files from my pr and just focus on the spark testing. It would be helpful to try and run the spark tests with your docker image as well. I'd prefer to have a single image that can be used for testing as well as running the notebooks.

yueguoguo:
@gramhagen Make sense. Then try mine as it is for running both scripts and notebooks (interactively).

yueguoguo:
@miguelgfierro updated to use environment variable values that correspond to the ones in our repo, i.e., `cpu`, `gpu`, and `pyspark`.

gramhagen:
have you tried to run this on a dsvm? I just built the image and ran into a few problems. It looks like jupyter is setup to require a token, but it is also listening to traffic on 127.0.0.1, which will cause problems if you're running on a vm. I think the jupyter ip address should be set to 0.0.0.0.

also, it might be easier to just disable the token based authentication for jupyter?

I tried to run some of the tests and they failed because they could not find the reco_utils package, maybe this needs to be added to the pythonpath?

on the plus side the image size is down to ~6GB, so close to what I was seeing with the image built from scratch. I'm still a bit concerned about introducing so many discrepancies btw the image and the default installation based on full conda install. even the version of python used is different (3.7 vs 3.6).

Other than speed of install do you think there's an advantage to going with this approach vs. clean install off ubuntu image?

yueguoguo:
> have you tried to run this on a dsvm? I just built the image and ran into a few problems. It looks like jupyter is setup to require a token, but it is also listening to traffic on 127.0.0.1, which will cause problems if you're running on a vm. I think the jupyter ip address should be set to 0.0.0.0.

I tried running it on the local laptop, and yea it listens to 127.0.0.1. When it is run, the token will be generated which is directly embedded in the url. What problem did you get? 

> also, it might be easier to just disable the token based authentication for jupyter?

Yea I think it is doable. This is the default setting in the original base image.

> I tried to run some of the tests and they failed because they could not find the reco_utils package, maybe this needs to be added to the pythonpath?

This is strange. All should be the same as clone into a local environment... 

> on the plus side the image size is down to ~6GB, so close to what I was seeing with the image built from scratch. I'm still a bit concerned about introducing so many discrepancies btw the image and the default installation based on full conda install. even the version of python used is different (3.7 vs 3.6). Other than speed of install do you think there's an advantage to going with this approach vs. clean install off ubuntu image?

Ah just realized that the Python version is 3.7. They just updated the Python version in the base image... Probably a tag pointing to the right version needs to be used. IMHO, the current base image is for the convenience to use Jupyter notebooks which are the major assets in our repository. We can definitely do everything from scratch from a ubuntu base image, but maybe that will be to reinvent the wheel... What do you think?

gramhagen:
I wasn't able to get to the jupyter server on the container when I ran it remotely, I created an ssh tunnel fwding port 8888 to my machine. I suspect using 0.0.0.0 would work better. I can't remember if you see output from the container if you run it in detached mode? If not then you won't be able to see the token. I'm not sure we're worried too much about security inside the container?

I think it's worth building from scratch as we can lock down exactly what's used and we can leverage the same setup as is done locally. I don't think there's a whole lot going on that would be reinvented. Basically all that's needed is some apt-installs, install spark, anaconda, then grab our repo setup an environment and adjust jupyter config settings. The only disadvantage is it's slower to build the image.

loomlike:
@yueguoguo Can we have another `ARG` to be able to specify a specific branch of the recommenders repo?

Example use case will be -- If I'm working on something on my branch, and I want to use this Dockerfile to create a docker image containing all of my changes (in my branch).

yueguoguo:
@gramhagen 

> I wasn't able to get to the jupyter server on the container when I ran it remotely, I created an ssh tunnel fwding port 8888 to my machine. I suspect using 0.0.0.0 would work better. I can't remember if you see output from the container if you run it in detached mode? If not then you won't be able to see the token. I'm not sure we're worried too much about security inside the container?

Ah I see. I have not tried forward a container run on a remote machine (say a VM) and tunnel it to a local one. I ran it on the local machine directly with `docker run --rm -p 8888 yueguoguo/reco_pyspark
` and it generated the output, thus token. I believe running in a detached mode will not give the interactive output. The token and security come from the base image, and sure if there is no particular concern about it we can remove that. 

> I think it's worth building from scratch as we can lock down exactly what's used and we can leverage the same setup as is done locally. I don't think there's a whole lot going on that would be reinvented. Basically all that's needed is some apt-installs, install spark, anaconda, then grab our repo setup an environment and adjust jupyter config settings. The only disadvantage is it's slower to build the image.

Yea make sense. Using the same approach to set up the whole environment as we do in the repo generated a super big image (like what I tried previously) that needs to be optimized

yueguoguo:
> @yueguoguo Can we have another ARG to be able to specify a specific branch of the recommenders repo? Example use case will be -- If I'm working on something on my branch, and I want to use this Dockerfile to create a docker image containing all of my changes (in my branch)

Yea sure actually it was included in the last PR but I removed it, as the idea is to provide user (of the Docker image) a stable version. It certainly makese sense to have the variable to allow developers to customize an image for their own use.

gramhagen:
@yueguoguo here's a gist for 3 images, they work fine from both the notebook and pytest perspectives (running locally or on remote vm). The final images are ~5-8GB. Thoughts on how what we should incorporate here?

https://gist.github.com/gramhagen/c572d9677ccfb43a323f7e42b071c059

yueguoguo:
@gramhagen It looks great. Just one general question, do we want to have just one Dockerfile with probably multiple stages in it, such that user can build with a desirable environment by passing the corresponding argument - there is something in common in all of the three Dockerfiles. What do you think?

gramhagen:
@yueguoguo yes I think that's a good idea, I can try using your conditional logic to consolidate them. Would be nice to only have one to maintain.

docker/Dockerfile
```diff
@@ -1,71 +1,98 @@
...
-        conda install --quiet --yes \
-        "numba>=0.38.1" \
-        "pytorch>=1.0.0" \
-        "tensorflow-gpu==1.12.0"; \
-    fi
-
-RUN fix-permissions $CONDA_DIR && \
-    fix-permissions /home/$NB_USER && \
-    fix-permissions $RECO_PATH
+ARG HOME
```
yueguoguo: What is the difference from line 5?

gramhagen: specifying the ARG before a FROM means it won't retain scope inside different FROM statements, so you have to re-declare it. kind of like python global. See the scope section under [Dockerfile ARG](https://docs.docker.com/engine/reference/builder/#arg) docs

docker/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+WORKDIR ${HOME}
+
+COPY --from=base ${HOME} .
+
+# Setup Conda environment
+ENV PATH="${HOME}/conda/bin:${PATH}"
+RUN python recommenders/scripts/generate_conda_file.py --name base --gpu
+
+
```
miguelgfierro: I think we should also consider to have the full environment: gpu+spark

SETUP.md
miguelgfierro: @gramhagen what is the final decision for the location of docker?

docker/Dockerfile
gramhagen: ok, i think this is doable

SETUP.md
gramhagen: What do people think about renaming scripts -> tools
then the structure for this would be

```
- tools/
    - docker/
        - Dockerfile
    - repo_metrics/
    - databricks_install.py
    - generate_conda_file.py
```

yueguoguo: I think this looks good. Alternatively, we may want to put `docker` at the top-level? I see many other repos do this.

gramhagen: fair, I've seen this quite a bit too. ok. I'm leaning towards keeping this in docker folder at the top level, that's not too cluttered.

we can rename scripts in a separate PR.

docker/Dockerfile
gramhagen: add the full environment for gpu + spark

yueguoguo:
@gramhagen FYI I added a note for exporting the environment variable of `DOCKER_BUILDKIT` to make sure the multi-stage build runs well. 

In general, I think the PR is good to go, and maybe we want to merge it into master?

gramhagen:
Ok, cool. Merged this to staging, will create a PR for merging to master.

PR #719, opened 2019-04-08
delete alpha parameter from pyspark ALS

tandav:
### Description
[ALS PySpark example](https://github.com/Microsoft/Recommenders/blob/master/notebooks/00_quick_start/als_movielens.ipynb) has parameters `implicitPrefs=False` and `alpha=0.1`. But 
> `alpha` is a parameter applicable to the implicit feedback variant of ALS that governs the baseline confidence in preference observations.
[source](https://spark.apache.org/docs/2.3.0/mllib-collaborative-filtering.html#collaborative-filtering)

So if explicit / `implicitPrefs=False` variant of ALS is used (movielens ratings are explicit), `alpha` should not be used

### Checklist:
- [ ] I have added tests.


miguelgfierro:
great catch @tandav, thanks for your help

tandav:
but maybe `alpha` works even with explicit ratings. 
It kinda coverts ratings range `[min_rating, max_rating]`:
```
alpha = 0.1
alpha * [0, 10] = [0, 1]
```

PR #723, opened 2019-04-09
more readable collect_set in _get_top_k_items

tandav:
### Description
add more readable `collect_set` using `groupBy` instead of window functions

it gives same result.  
But if there are recommendations with equal ratings  
then recommendations can be different between runs, because there is no "right way to order equal numbers"

### Checklist:

- [ ] I have added tests.


miguelgfierro:
adding @loomlike @gramhagen and @yueguoguo to review this

PR #727, opened 2019-04-09
Added azureml logging for top-k and data-size

heatherbshapiro:
### Description

Adding 2 lines to log the top-k and data-size on the run. This will help users keep track of them when using AzureML and will make it easier to compare runs

### Checklist:

- [x] My code follows the code style of this project, as detailed in our [contribution guidelines](../CONTRIBUTING.md).
- [ ] I have added tests.
- [x] I have updated the documentation accordingly.

PR #728, opened 2019-04-10
Gramhagen/sar on aml update

gramhagen:
### Description
adding config.json to .gitignore
moving workspace get/create script to python utility
updating sar on aml notebook

### Related Issues
#727 

### Checklist:

- [ ] I have added tests.


notebooks/00_quick_start/sar_movielens_with_azureml.ipynb
```diff
@@ -115,59 +117,33 @@
...
-     "output_type": "stream",
-     "text": [
-      "You are currently using version 1.0.21 of the Azure ML SDK\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
-    "print(\"You are currently using version\", azureml.core.VERSION, \"of the Azure ML SDK\")"
+    "ws = get_or_create_workspace()"
```
heatherbshapiro: We chose to do it with the magic load so that users do not have to leave this notebook to add the subscription info or resource group. This will then create the config.json. If we have a utility function, they will still need to go elsewhere to upload a config.json.

gramhagen: this function works the same way as the load script, if the user already has a config it will use it, otherwise it will use environment variables (or inputs to the method) to look up the workspace, and lastly it will create the workspace with the settings if needed. in all cases the config.json is saved locally and the user isn't required to go to the portal to retrieve them.

heatherbshapiro: Oh I see, I didn't realize you could put parameters. Can we explicitly show the parameters and explain that if it is a user's first time they can fill them in? `get_or_create_workspace(
    config_path=None,
    subscription_id=None,
    resource_group=None,
    workspace_name=None,
    workspace_region=None,
)`

gramhagen: sure makes sense, i'll add a comment that the user either needs to have config.json already, or set the env vars, or pass them in.

Issue #549, opened 2019-02-20
[FEATURE]Add "similar-item recommendations" in SAR


feihe6:
### Description

Our team's data policy won’t allow us to use any ID information for recommendation at this point. We would love to provide dynamic recommendation based on user's new choice of items rather his ID.
### Expected behavior with the suggested feature

In our scenario, we would like to recommend user new template to try. We are interested in use both limited user profile information and user's template download history to provide recommendation. Since we are only allowed to use very few attributes related to user and no ID information is allowed at this time point, I am thinking of taking advantage of item co-occurrence similarity to provide more dynamic in the recommendation. 

https://github.com/Microsoft/Product-Recommendations/blob/master/doc/sar.md#frequently-occurring-together-recommendations 
### Other Comments
I have used the "similar-item recommendation" feature of SAR from another repo before: https://github.com/Microsoft/Product-Recommendations

miguelgfierro:
Thanks for the request. We will evaluate internally and provide a solution to your request.

This is in line with what @loomlike suggested in the past.

yueguoguo:
Add @maxkazmsft as well. This is something useful as we have seen item-to-item typed recommendation scenario a few times.

yueguoguo:
@feihe6 not sure if your problem has been resolved?

feihe6:
@yueguoguo Hi Le I used this piece of code to get the item-similarity with itemId and userId. Not sure this is the right solution.

### output item_similarity matrix 
item_similarity_matrix = model.item_similarity
item_similarity = pd.DataFrame(item_similarity_matrix)
item_similarity["ItemIndex"] = item_similarity.index
item_similarity_long = pd.melt(item_similarity, id_vars=["ItemIndex"]) 
item_similarity_long = item_similarity_long.convert_objects(convert_numeric=True)
item_similarity_long = item_similarity_long.rename(index=str, columns={"variable": "ItemIndex2"})
### Add item index
item_index = model.index2item
item_index = pd.DataFrame.from_dict(item_index, orient='index')
item_index["ItemIndex"] = item_index.index
item_index = item_index.rename(index=str, columns={0: "TemplateId"})
item_index2 = item_index
item_index2 = item_index2.rename(index=str, columns={"ItemIndex": "ItemIndex2", "TemplateId": "TemplateId2"})
### output the templateId with similarity value
item_similarity_1 = item_similarity_long.merge(item_index, on="ItemIndex", how="left")
[...]

PR #734, opened 2019-04-11
Gramhagen/sar item sim

gramhagen:
### Description
Addition of similar item recommendations to SAR implementation
This lets you create a set of seed items and get the top-k recommended items similar to that set of items (excluding the ones provided).
You can put in multiple seed sets by adding a user column, and you can weight the seeds by providing a rating column as well.
This essentially provides a way to generate top-k recommendations for new users, but does not update the model.

```
    sar = SARSingleNode()
    sar.fit(data)

    items = pd.DataFrame(dict(itemID=[1, 5, 10]))
    results = sar.recommend_similar_items(items, top_k=3)
```

results will be a dataframe with user, item and prediction columns (user is 0 unless multiple users are provided in items dataframe)

### Related Issues
#549 

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/recommender/sar/sar_singlenode.py
+++ b/reco_utils/recommender/sar/sar_singlenode.py
@@ -103,6 +103,7 @@ class SARSingleNode:
             df (pd.DataFrame): Indexed df of users and items.
             n_users (int): Number of users.
             n_items (int): Number of items.
+
         Returns:
             sparse.csr: Affinity matrix in Compressed Sparse Row (CSR) format.
         """
@@ -121,6 +122,7 @@ class SARSingleNode:
             df (pd.DataFrame): Indexed df of users and items.
             n_users (int): Number of users.
             n_items (int): Number of items.
+
         Returns:
             np.array: Co-occurrence matrix
         """
@@ -251,6 +253,7 @@ class SARSingleNode:
         Args:
             test (pd.DataFrame): user to test
             remove_seen (bool): flag to remove items seen in training from recommendation
+
         Returns:
             np.ndarray
         """
@@ -278,14 +281,81 @@ class SARSingleNode:
 
         return test_scores
 
+    def recommend_similar_items(self, items, top_k=10, sort_top_k=False):
+        """Recommend top K items similar to the provided seed items
+
+        Args:
+            items (pd.DataFrame): dataframe with item, user (optional), and rating (optional) columns
+            top_k (int): number of top items to recommend
+            sort_top_k (bool): flag to sort top k results
+
+        Returns:
+            pd.DataFrame: sorted top k recommendation items
+        """
+
+        # convert item ids to indices
+        item_ids = items[self.col_item].map(self.item2index)
+
+        # if no ratings were provided assume they are all 1
+        if self.col_rating in items.columns:
+            ratings = items[self.col_rating]
+        else:
+            ratings = pd.Series(np.ones_like(item_ids))
+
+        # create local map of user ids
+        if self.col_user in items.columns:
+            test_users = items[self.col_user]
+            user2index = {x[1]: x[0] for x in enumerate(items[self.col_user].unique())}
+            user_ids = test_users.map(user2index)
+        else:
+            # if no user column exists assume all entries are for a single user
+            test_users = pd.Series(np.zeros_like(item_ids))
+            user_ids = test_users
+        n_users = user_ids.drop_duplicates().shape[0]
+
+        # generate pseudo user affinity using seed items
+        pseudo_affinity = sparse.coo_matrix(
+            (ratings, (user_ids, item_ids)),
+            shape=(n_users, self.n_items),
+        ).tocsr()
+
+        # calculate raw scores with a matrix multiplication
+        test_scores = pseudo_affinity.dot(self.item_similarity)
+
+        # remove items in the seed set so recommended items are novel
+        test_scores[user_ids, item_ids] = -np.inf
+
+        # ensure we're working with a dense matrix
+        if isinstance(test_scores, sparse.spmatrix):
+            test_scores = test_scores.todense()
+
+        return self.get_top_k_items(users=test_users, scores=test_scores, top_k=top_k, sort_top_k=sort_top_k)
+
     def recommend_k_items(self, test, top_k=10, sort_top_k=False, remove_seen=False):
         """Recommend top K items for all users which are in the test set
 
         Args:
-            test (pd.DataFrame): user to test
+            test (pd.DataFrame): users to test
             top_k (int): number of top items to recommend
             sort_top_k (bool): flag to sort top k results
             remove_seen (bool): flag to remove items seen in training from recommendation
+
+        Returns:
+            pd.DataFrame: top k recommendation items for each user
+        """
+
+        test_scores = self.score(test, remove_seen=remove_seen)
+        return self.get_top_k_items(users=test[self.col_user], scores=test_scores, top_k=top_k, sort_top_k=sort_top_k)
+
+    def get_top_k_items(self, users, scores, top_k, sort_top_k=False):
+        """Extract top K items from score matrix, optionally sort results per user
+
+        Args:
+            users (pd.Series): test users
+            scores (np.array): score matrix
+            top_k (int): number of top items to recommend
+            sort_top_k (bool): flag to sort top k results
+
         Returns:
             pd.DataFrame: top k recommendation items for each user
         """
@@ -294,14 +364,13 @@ class SARSingleNode:
             logger.warning('Number of items is less than top_k, limiting top_k to number of items')
         k = min(top_k, self.n_items)
 
-        test_scores = self.score(test, remove_seen=remove_seen)
-        test_user_idx = np.arange(test_scores.shape[0])[:, None]
+        test_user_idx = np.arange(scores.shape[0])[:, None]
 
         # get top K items and scores
         logger.info("Getting top K")
         # this determines the un-ordered top-k item indices for each user
-        top_items = np.argpartition(test_scores, -k, axis=1)[:, -k:]
-        top_scores = test_scores[test_user_idx, top_items]
+        top_items = np.argpartition(scores, -k, axis=1)[:, -k:]
+        top_scores = scores[test_user_idx, top_items]
 
         if sort_top_k:
             sort_ind = np.argsort(-top_scores)
@@ -311,7 +380,7 @@ class SARSingleNode:
         df = pd.DataFrame(
             {
                 self.col_user: np.repeat(
-                    test[self.col_user].drop_duplicates().values, k
+                    users.drop_duplicates().values, k
                 ),
                 self.col_item: [
                     self.index2item[item] for item in np.array(top_items).flatten()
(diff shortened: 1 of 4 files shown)


gramhagen:
@heatherbshapiro this is the item similarity method we can use for demonstrating SAR recommendations

heatherbshapiro:
I'm a little naive here - so the itemIds are say movies. We are putting in 3 movie IDs (regardless of what the new user rates them) and getting the top -k  similar movies to those 3?

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -281,6 +284,63 @@ def score(self, test, remove_seen=False):
...
+                    self.index2item[item] for item in np.array(top_items).flatten()
+                ],
+                self.col_prediction: np.array(top_scores).flatten(),
+            }
+        )
+
+        # drop invalid items
+        return df.replace(-np.inf, np.nan).dropna()
+
+    def recommend_popular_items(self, top_k=10, sort_top_k=False):
```
miguelgfierro: I don't understand very well this function and the reason why we are making get_topk_items private. Why is it only for user `test_users = pd.Series([0])`? also, don't we want to support the option of getting the top k items given some users?

FYI @anargyri @yueguoguo

anargyri:
> I'm a little naive here - so the itemIds are say movies. We are putting in 3 movie IDs (regardless of what the new user rates them) and getting the top -k similar movies to those 3?

I think it would be good to talk about the motivation / why we may want to do this. Also to emphasise that this type of recommendation is _not personalised_ (i.e. it would probably appear as a separate carousel / frame). There are some cases I can think of:
- Increase diversity; say, there is a popular TV show, then the TV station may want to find which shows "co-occur" and recommend them along with the popular one. 
- Interpretability; this type of recommendation is easy to explain and understand, and could be helpful to business / product development etc.
- In cases with cold users, when user features are not available, one of the few options is to take a baseline that is not personalised (e.g. popularity, or rule-based logic) and generate recommendations based on the top items of the baseline.

anargyri:
> @heatherbshapiro this is the item similarity method we can use for demonstrating SAR recommendations


Incidentally, SAR also has a variant of item similarity that is "catalogue-based"  (i.e. based on item features). We may want to consider that as a future issue.

reco_utils/recommender/sar/sar_singlenode.py
anargyri: Don't we use `recommend_k_items()` in the SAR notebooks?

anargyri: Also `remove_seen` is a flag we need.

anargyri: Since the PR is about new functionality, I don't see why the old functionality and signature should change at all.

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -323,11 +329,96 @@ def recommend_k_items(self, test, top_k=10, sort_top_k=False, remove_seen=False)
...
+        # ensure we're working with a dense matrix
+        if isinstance(test_scores, sparse.spmatrix):
+            test_scores = test_scores.todense()
+
+        return self._get_top_k_items(
+            users=test_users, scores=test_scores, top_k=top_k, sort_top_k=sort_top_k
+        )
+
+    def recommend_similar_items(self, items, top_k=10, sort_top_k=False):
+        """Recommend top K items similar to the provided seed items
```
anargyri: Maybe be a bit more specific and say "k most similar items in terms of co-occurrence" or something along these lines.

reco_utils/recommender/sar/sar_singlenode.py
anargyri: Maybe pass `users` in `_get_top_k_items()` as a list or numpy array would have been simpler.

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -323,11 +329,96 @@ def recommend_k_items(self, test, top_k=10, sort_top_k=False, remove_seen=False)
...
+            test_users = items[self.col_user]
+            user2index = {x[1]: x[0] for x in enumerate(items[self.col_user].unique())}
+            user_ids = test_users.map(user2index)
+        else:
+            # if no user column exists assume all entries are for a single user
+            test_users = pd.Series(np.zeros_like(item_ids))
+            user_ids = test_users
+        n_users = user_ids.drop_duplicates().shape[0]
+
+        # generate pseudo user affinity using seed items
```
anargyri: I don't understand what the role of users is in this function. Why should users even enter the picture? Don't you just want to get the k largest entries from the item similarity matrix?

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: this is just refactoring to avoid repeating the same functionality in all 3 recommend methods. maybe the name isn't great. alternatively we could adjust the input / output further and replace the get_top_k_items method in python_evaluation?

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: this is so you can input multiple cold users, or multiple item seeds

gramhagen:
@heatherbshapiro 

recommend_similar_items will take a dataframe with the following options:
1. Just items without ratings (ratings are filled in as 1 so all items are considered equal)
2. Items and users without ratings (sames as above but supports multiple seed sets of items)
3. Items with ratings (this would be how you could simulate a new user)
4. Items, ratings, and users (this is how you could simulate multiple new users)

also I added the recommend_popular_items, so if you don't know anything about the user you could just get most commonly occurring items.

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: since _get_top_k_items needs to support cases where there are multiple users I need to create a series for the users (test_users), this is kind of hacky. I suppose a better way would be to remove users from get_top_k, and handle the generation of the dataframe with outside of that function.

anargyri: I am not sure what you mean by "multiple users"; isn't this the standard case that was supported before?
I agree that https://github.com/Microsoft/Recommenders/blob/9d76d1c81b3650326988b61094959cbd96efe709/reco_utils/evaluation/python_evaluation.py#L685
looks similar and could be refactored so that we have a single public function in the common utilities. Also @miguelgfierro has a point about making `_get_top_k_items()` public, because there is no fucntional dependence on SAR any more (only the column names).

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -323,11 +329,96 @@ def recommend_k_items(self, test, top_k=10, sort_top_k=False, remove_seen=False)
...
+
+        # ensure we're working with a dense matrix
+        if isinstance(test_scores, sparse.spmatrix):
+            test_scores = test_scores.todense()
+
+        return self._get_top_k_items(
+            users=test_users, scores=test_scores, top_k=top_k, sort_top_k=sort_top_k
+        )
+
+    def recommend_similar_items(self, items, top_k=10, sort_top_k=False):
```
miguelgfierro: We already have several function names with recommend. Can we say something different to avoid confusion? Maybe `find_topk_similar_items` or `get_similar_items` ...

reco_utils/recommender/sar/sar_singlenode.py
miguelgfierro: We have `recommend_popular_items(self, top_k=10, sort_top_k=False)` and `recommend_k_items(self, test, top_k=10, sort_top_k=False, remove_seen=False)`. I have two questions:
1) what is the difference between both?
2) Can't we have recommend_popular_items as a specific case of recommend_k_items?

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         self.model_str = "sar_ref"
         self.user_affinity = None
         self.item_similarity = None
+        self.item_frequencies = None
```
miguelgfierro: I think we can consider (in this PR or another one) to show the functionality of getting similar items in the deep dive notebook

reco_utils/recommender/sar/sar_singlenode.py
anargyri: And what does this mean more precisely? How do you handle cold users, or item seeds? Without reading the code one cannot understand what these functions do from the docstring. 
What scenarios are you covering with this function? We have an item-item similarity matrix already computed with `fit()`. One scenario is to recommend similar items to a given set of items, which can be simply read from this matrix. But clearly you want to cover other scenarios too.

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: @miguelgfierro  ```recommend_popular_items``` just gets the most frequently occurring items across all users. we could use it as a fallback case if all the items in ```recommend_similar_items``` are cold, or the user is cold in recommend_k_items. Either way it seems useful to maintain as it's own method in case the user wants to access that information or if the method is shared across other functions. happy to change the names of this if you think something else would be clearer

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: ok, will do

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: i'll try to clarify the options in the docstrings

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: my thought was they are all providing similar output so it made sense to keep the names the same, but I'm fine with whatever name is clearer. ```get_item_based_topk``` ?

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: yeah, good idea

reco_utils/recommender/sar/sar_singlenode.py
miguelgfierro: Oh I see, yeah I would suggest to rename it to something like what you said: `recommend_most_popular_items` and then add `gets the most frequently occurring items across all users.` in the docstrings

reco_utils/recommender/sar/sar_singlenode.py
miguelgfierro: that sounds good

reco_utils/recommender/sar/sar_singlenode.py
anargyri: I am wondering, since `recommend_most_popular_items` functionality is useful more generally with cold users, regardless of SAR; what would be a good way to make it available in other scenarios too e.g. when they train an ALS or SVD model for warm users but want a most popular model for cold ones.

gramhagen: we could certainly offer a utility to determine that based on a dataframe of user/items, that might be more useful than this version since I'm only looking at counts not rating values, which could promote frequently occurring low-ranked items above moderately occurring high-ranked items

PR #735, opened 2019-04-11
lgbm o16n tuning pass

jreynolds01:
### Description
Adjusted spark serving params to be better tuned.

Adjusted AksWebservice configuration to meet a load demand and added documentation about how to decide what values to choose.

### Related Issues
N/A

### Checklist:

- [X ] My code follows the code style of this project, as detailed in our [contribution guidelines](../CONTRIBUTING.md).
- [x] I have added tests.
- [X ] I have updated the documentation accordingly.


notebooks/05_operationalize/lightgbm_criteo_o16n.ipynb
```diff
@@ -122,6 +122,8 @@
     "from azureml.core.image import ContainerImage\n",
```
miguelgfierro: <p>This is so cool man, I don't think there something like this anywhere</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/735/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

miguelgfierro: <p>I think it would be nice to have some of the results from Devops of request time. We can also add that image to the OKR document</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/735/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='40'/>

jreynolds01: <p>Thanks! That estimation was hidden in docs (at the link I referenced), but it **is** really useful. Was wondering if this should actually be a function in common or in azureml that we could just import.</p>

jreynolds01: <p>I think the scenario doc is the better place for that, but it's probably worth adding a cell with the results here at the very end. Will try to get to that later today.</p><p><br></p>

gramhagen: <p>suggest using slug style variables instead of camelcase</p><p><br></p><p>also, is rps standard? I've seen qps more often, but wasn't sure if that was industry specific</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/735/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

gramhagen: <p>a jupyter widget would be really slick here, letting people adjust the parameters and see the output curves for qps vs # replicas</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/735/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

gramhagen: <p>cpu_cores is a little confusing to me (being a fraction), how should I think about what that number means? where did 0.3 come from?</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/735/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

gramhagen: <p>why not combine this logic into the first method, and provide some messaging around adjustments made if estimated cores is lower than minimum</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/735/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

gramhagen: <p>this is good stuff, super useful!</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/735/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

jreynolds01: it's the number of cores you want to allocate per replica, and since you can have multiple replicas / agent, it can be fractional. See [here](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#cpu-units) for more info.

jreynolds01: and agreed - it is non-intuitive

jreynolds01: sure, i'll update to slug. 

Don't know if rps is standard, I just re-used what was in the [docs page](https://docs.microsoft.com/en-us/azure/machine-learning/service/how-to-deploy-and-where#autoscaling). Let's use qps if you think it's more common

jreynolds01: <p>one is going from qps -&gt; replicas, this one is going from replicas -&gt; qps. I suppose one could estimate  replicas if qps was passed, and estimate qps if replicas was passed. I feel like that violates single responsibility, but there are a lot of replicated characters here.</p>

miguelgfierro: I guess if the estimation is only related to azureml, it should go in azureml folder

gramhagen: got it, I was thinking of replicas as separate vms, not containers on the same vm.

jreynolds01: <p>ok, moved these, renamed them, and added tests.</p>

jreynolds01: <p>fixed.</p>

jreynolds01: <p>yeah, agreed. that could be really cool. One issue is that we are expecting this to be run in Databricks, so I think that makes it challenging. So, maybe we can set that as a future iteration.</p><p><br></p>

jreynolds01: <p>updated with more explanation and link to kube docs.</p>

jreynolds01: <p>thanks!</p>

gramhagen: makes sense, it actually probably sounds cooler than it would look since this is all linear, showing the tabular for like you have seems fine

gramhagen: ok, np. this is like fine tuning the estimate, makes sense

reco_utils/azureml/aks_utils.py
```diff
@@ -0,0 +1,57 @@
+from math import ceil, floor
+
+def qps_to_replicas(target_qps, processing_time, max_qp_replica=1, target_utilization=0.7, verbose=True):
+    """
```
miguelgfierro: A minor comment. In the other code we are using the docstring format like this:
```
"""Provide a rough estimate ...
```
Also, another annoying thing is that if we don't add a new line between the end of `Args` and `Returns`, when we use sphinx, it won't parse the docstrings. I learned this the hard way because I have to add a new line in all the code of a library that I had, and it was very time consuming.

PR #741, opened 2019-04-16
Gramhagen/merge master patch

gramhagen:
### Description
addressing code review comments from #720 

cleanup of sar_on_aml notebook, suggested change to readme algo list to move links to specific notebooks under the environments column (so we can showcase the same algo with different implementatation, i.e. LightGBM on Spark / CPU)

### Checklist:

- [ ] I have added tests.


jingyanwangms:
https://github.com/Microsoft/Recommenders/blob/staging/notebooks/run_notebook_on_azureml.ipynb used same text about AmlCompute. And I only noticed the PR is merged after a refresh, I will create a new one to fix both comment.

PR #743, opened 2019-04-17
Fix minor typos

az0:
### Description

This change is required to solve minor typographical errors. These changes do not affect any code.

### Related Issues

Not applicable

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/notebooks/03_evaluate/evaluation.ipynb
+++ b/notebooks/03_evaluate/evaluation.ipynb
@@ -121,7 +121,7 @@
    "source": [
     "For illustration purpose, a dummy data set is created for demonstrating how different evaluation metrics work. \n",
     "\n",
-    "The data has the schema that can be frequently found in a recommendation problem, that is, each row in the dataset is a (user, item, rating) tuple, where \"rating\" can be an ordinal rating score (e.g., discrete integers of 1, 2, 3, etc.) or an numerical float number that quantitatively indicates the preference of the user towards that item. \n",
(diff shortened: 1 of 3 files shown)


notebooks/03_evaluate/evaluation.ipynb
```diff
@@ -121,7 +121,7 @@
    "source": [
```
gramhagen: <p>I think there's still a type in the first sentence here, maybe a better version would be:</p><p>"For comparison, a similar process is used with a threshold value of 3 to create a more balanced dataset."</p>

 _Reply via <a href='https://app.reviewnb.com/Microsoft/Recommenders/pull/743/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='64'/>

gramhagen:
Thanks for the updates, one thing to note is we push PRs to staging instead of master, so I've changed the base used here. In this case I think it's okay because the differences are small, but normally I would suggest working off staging and pushing PRs from branches based on changes to the staging branch.

notebooks/03_evaluate/evaluation.ipynb
az0: Great suggestion. I used it.

az0:
I will try to be more careful about the branch. Thanks for the kind advice

az0:
Replaced by https://github.com/Microsoft/Recommenders/pull/744

Issue #736, opened 2019-04-12
[FEATURE] Use rand-seed for reproducibility


loomlike:
### Description
Use seed for NN-based models - both at notebooks and tests

### Expected behavior with the suggested feature
Produce same results from notebooks
Assert exact value from tests

miguelgfierro:
we are using seed in the algos: https://github.com/Microsoft/Recommenders/blob/master/reco_utils/recommender/rbm/rbm.py#L74. I'm not sure if there is anyone missing. 

When trying to address reproducibility, @anargyri @msalvaris  and I had some fun trying to achieve it. In DL is very difficult, since the optimization is stochastic and there are some asynchronous processes in the GPU that make difficult to have complete reproducibility.

loomlike:
hmmm seems like there are many threads discussing about non-deterministic results from tf out there. 
In fastai, you can set several seeds to make the results reproducible.

Have you tried no-multi-threading with all the seeds set including `tf.random.set_random_seed`? Still no luck?

miguelgfierro:
yeah we have that function in all the DL algos: https://github.com/Microsoft/Recommenders/search?q=tf.random.set_random_seed&unscoped_q=tf.random.set_random_seed

miguelgfierro:
see this, just yesterday we got an error in the nightly builds for xdeepfm: https://msdata.visualstudio.com/DefaultCollection/AlgorithmsAndDataScience/_build/results?buildId=2873512

loomlike:
yeah... we should handle this randomness, but I couldn't find time to work on this issue. If this is urgent thing, can anybody take this item and fix? If not, we can temporally loose the tolerance value for NN algos and I will take care in the next few weeks.

miguelgfierro:
adding more folks to discuss this @anargyri @yueguoguo @gramhagen. I agree with Jun Ki, I think it is a very annoying issue. Many times our GPU nightly builds fails because of this

miguelgfierro:
One example of a nightly build that failed:

```
tests/smoke/test_deeprec_model.py ..                                     [ 25%]
tests/smoke/test_notebooks_gpu.py ...F..                                 [100%]

=================================== FAILURES ===================================
____________________________ test_notebook_xdeepfm _____________________________

[...]

PR #748, opened 2019-04-18
Update xDeepFM to work with random seed

loomlike:
### Description
Changes include:
* Refactor xDeepFM to fully use the seed
* Update xdeepfm smoke test
* Add xdeepfm unit and integration tests
* Change xDeepFM quickstart notebook name to xdeepfm_criteo
* Update the notebook contents
 
### Related Issues
#736 

### Checklist:

- [x] I have added tests.


reco_utils/recommender/deeprec/models/base_model.py
```diff
@@ -14,7 +12,7 @@
 
 
 class BaseModel(object):
-    def __init__(self, hparams, iterator_creator, graph=None, seed=42):
+    def __init__(self, hparams, iterator_creator, graph=None, seed=None):
```
miguelgfierro: why `None` here? in the rest of the notebooks we have 42

loomlike: The default `seed` should be the same as the functions' that we use in the module, which is `None`. See, e.g., [tensorflow](https://www.tensorflow.org/api_docs/python/tf/random/normal).

This follows the rational that if you don't pass the seed explicitly, the behavior should be 'random' (produce different results every time you run it).
e.g. Python's random behavior: 
> random.seed(a=None, version=2): If a is omitted or None, the current system time is used.

tests/smoke/test_notebooks_gpu.py
```diff
@@ -36,9 +36,9 @@ def test_ncf_smoke(notebooks):
     assert results["recall"] < 0.10
 
 
-@pytest.mark.notebooks
+@pytest.mark.smoke
 @pytest.mark.gpu
-def test_ncf_deep_dive(notebooks):
+def test_ncf_deep_dive_smoke(notebooks):
```
gramhagen: why change the all test names? they're already in a separate folder and marked as smoke ?

also, here we want to keep @pytest.make.notebooks as well right? hmm, odd we don't have notebooks marked on the rest of the tests ?

gramhagen: also, can we reduce the tolerance levels now?

loomlike: Changed the name to be the same as the rest of the tests in the module.
Don't know who starts first. haha
I kinda like it, because when it fails, it is easy to **read** from the pytest log what's failing, unit vs smoke vs integration.

Regarding the tolerance levels, I believe NNs models have been using their own numbers (hard-coded), correct me if I'm wrong, while the others use `TOL` and `ABS_TOL` constants. So, we can leave the numbers as it is. But, actually, I want to use TOL only, not ABS_TOL, since different metrics' range could be different for each other and thus using absolute tolerance doesn't make sense (TOL is relative).

loomlike: If we agree to not use ABS_TOL, we can create a separate issue to take care of that.

gramhagen: ok, makes sense as long as they're consistently named

yeah, whatever tolerance is appropriate is fine, but it would be good to be small enough to detect an potential error, vs random noise now that that is fixed. fine with creating a separate issue for that

loomlike: Also, I was wondering about *notebooks* mark too. Integration and smoke tests don't have that mark. Maybe because we don't use notebooks mark when we run smoke and integration tests?

gramhagen: yes, that's true, for nightly smoke and integration tests we run 3 versions: spark, gpu, base, there's no differentiation for notebooks so I think it's fine to remove it.

reco_utils/recommender/deeprec/models/base_model.py
miguelgfierro: if this is the defualt behavior we want to follow, then we should have the same idea in all the other algos, which should be changed. Do we want to follow this idea instead of fixing the seed? @anargyri @yueguoguo @gramhagen

loomlike: I think it is a good practice to use `seed=None` for algo modules, and set seed explicitly from example notebooks.

gramhagen: This sounds reasonable, i guess right now you would have to set the seed to None to actually get random values, which may be unexpected. I think we should adopt this as it matches approaches from other libraries and create a separate ticket to change the notebooks / code / tests as needed.

tests/smoke/test_notebooks_gpu.py
miguelgfierro: we have to be carefull with `ABS_TOL`, this was added because rel_tol was not enough when testing very small metric results (in the order of 0.001 or so) I wouldn't remove it unless we are sure that these problems are gone

reco_utils/recommender/deeprec/models/base_model.py
miguelgfierro: in numpy the default is None: https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.seed.html#numpy.random.seed, so I think it makes sense to change all to None

@loomlike do you want to homogenize None in all the utilities and then set 42 explicitly in the notebooks? Not sure if on this PR or in a different want

loomlike: @miguelgfierro I will create a separate issue to handle them

tests/smoke/test_notebooks_gpu.py
loomlike: Ah, yeah. Make sense.

Issue #747, opened 2019-04-18
[BUG] No matching distribution found for nni==0.5.2.1 on Windows DSVM


roalexan:
### Description
Seeing an error when installing nni on Windows. This is the error that I see when setting up the environment:

DEBUG menuinst_win32:__init__(196): Menu: name: 'Anaconda${PY_VER} ${PLATFORM}', prefix: 'C:\Anaconda\envs\nightly_reco_base', env_name: 'nightly_reco_base', mode: 'system', used_mode: 'system'
DEBUG menuinst_win32:create(320): Shortcut cmd is C:\Anaconda\python.exe, args are ['C:\\Anaconda\\cwp.py', 'C:\\Anaconda\\envs\\nightly_reco_base', 'C:\\Anaconda\\envs\\nightly_reco_base\\python.exe', 'C:\\Anaconda\\envs\\nightly_reco_base\\Scripts\\jupyter-notebook-script.py', '"%USERPROFILE%/"']
  Could not find a version that satisfies the requirement nni==0.5.2.1 (from -r C:\Users\recocat\agent\_work\1\s\condaenv.ieyjoeec.requirements.txt (line 11)) (from versions: 0.3.2, 0.3.3, 0.3.4)
No matching distribution found for nni==0.5.2.1 (from -r C:\Users\recocat\agent\_work\1\s\condaenv.ieyjoeec.requirements.txt (line 11))

CondaValueError: pip returned an error

### In which platform does it happen?
Windows DSVM

### How do we replicate the issue?
Follow the setup instructions for testing on Windows. Creating the reco_base python environment will cause the error.

### Expected behavior (i.e. solution)
no error. not sure about solution - we'll probably have to conditionally disable this somehow on Windows.

gramhagen:
it seems as though they are not supporting windows past version 0.3.4 (at least from what is availabily on PyPI)

i'm not sure if the repo is cloned if it can still be built on windows, but this introduces a challenge both on the yaml file generation side and the unit testing side. Should we check the OS system and remove nni / nni tests?

roalexan:
In addition to not running the nni tests on Windows, we'll have to see how to create the python environment, minus this one package. Or maybe use the latest installable nni package on either environment and see if the tests run.

gramhagen:
yeah, it's feasible to remove this package in the generate_conda_file script if it detects a windows system, but it seems less than ideal. Let's hold until @anargyri can weigh in.

[...]

Issue #746, opened 2019-04-18
[BUG] in NNI integration test


miguelgfierro:
### Description

```
tests/integration/test_movielens.py .........                            [ 62%]
tests/integration/test_notebooks_python.py .....F                        [100%]

=================================== FAILURES ===================================
_____________________________ test_nni_tuning_svd ______________________________

notebooks = {'als_deep_dive': '/data/home/recocat/cicd/7/s/notebooks/02_model/als_deep_dive.ipynb', 'als_pyspark': '/data/home/rec...baseline_deep_dive.ipynb', 'data_split': '/data/home/recocat/cicd/7/s/notebooks/01_prepare_data/data_split.ipynb', ...}
tmp = '/tmp/pytest-of-recocat/pytest-200/tmpuj9_hwzq'

    @pytest.mark.integration
    def test_nni_tuning_svd(notebooks, tmp):
        notebook_path = notebooks["nni_tuning_svd"]
        # First stop NNI in case it is running
        subprocess.run([sys.prefix + '/bin/nnictl', 'stop'])
        check_stopped()
        pm.execute_notebook(notebook_path, OUTPUT_NOTEBOOK, kernel_name=KERNEL_NAME,
                            parameters=dict(MOVIELENS_DATA_SIZE="100k",
                                            SURPRISE_READER="ml-100k",
                                            TMP_DIR=tmp,
                                            MAX_TRIAL_NUM=1,
>                                           NUM_EPOCHS=1))

tests/integration/test_notebooks_python.py:164: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/anaconda/envs/nightly_reco_base/lib/python3.6/site-packages/papermill/execute.py:94: in execute_notebook
    raise_for_execution_errors(nb, output_path)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

nb = {'cells': [{'cell_type': 'code', 'metadata': {'inputHidden': True, 'hide_input': True}, 'execution_count': None, 'sour...d_time': '2019-04-18T12:09:01.272263', 'duration': 345.187067, 'exception': True}}, 'nbformat': 4, 'nbformat_minor': 2}
output_path = 'output.ipynb'

[...]

PR #749, opened 2019-04-19
Gramhagen/nni windows patch

gramhagen:
### Description
changes to conda file generator to skip nni installation if os is windows
similar changes to skip integration tests if os is windows (nni_utils unit tests are ok since they don't import from nni)
also changed the nni_tuning notebook to use the temporary folder to store experiment results, and skipped SMAC optimization for now due to #746 

### Related Issues
#746 
#747 

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/scripts/generate_conda_file.py
+++ b/scripts/generate_conda_file.py
@@ -83,12 +83,16 @@ PIP_BASE = {
     "tqdm": "tqdm==4.31.1",
 }
 
-PIP_PYSPARK = {}
 PIP_GPU = {}
+PIP_PYSPARK = {}
 
+PIP_DARWIN = {
+    "nni": "nni==0.5.2.1",
+}
 PIP_LINUX = {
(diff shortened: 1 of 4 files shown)


scripts/generate_conda_file.py
```diff
@@ -76,15 +77,22 @@
...
     "papermill": "papermill==0.18.2",
     "pydocumentdb": "pydocumentdb>=2.3.3",
     "tqdm": "tqdm==4.31.1",
 }
 
-PIP_PYSPARK = {}
 PIP_GPU = {}
+PIP_PYSPARK = {}
+
+PIP_DARWIN = {
```
miguelgfierro: do we need different options for mac and linux, shouldn't be the same?

gramhagen: not sure if there are any cases where these might differ? I kept it separate just for clarity.

Issue #597, opened 2019-03-01
Docstring to NCF


loomlike:
### Description
NCF example implementation doesn't have docstring. It does have enough comments. We can move those comments to docstring format.

### Expected behavior (i.e. solution)
NCF class with good docstrings

### Other Comments
Also would be good to have gpu check codes at its example notebook. FYI, gpu check codes are in wide-deep quickstart notebook.

yueguoguo:
Not sure if @AaronHeee has the bandwidth on this - I can take it if not.

AaronHeee:
Sure, I will take it @yueguoguo

PR #759, opened 2019-04-25
Ncf docs

miguelgfierro:
### Description

Updated docstrings. 

There is one change that I'm not sure if it will change the results of the [benchmark](https://github.com/Microsoft/Recommenders/blob/master/benchmarks/movielens.ipynb), the results of NCF are very good, not sure because of this. The original code had a flag for `implicit=True` by default, which was inaccessible, because it was in a private function. I added it to the constructor, so now we can access it. The problem is that this flag actually is doing this: `If true, set rating > 0 to rating = 1`. However, Movielens is not implicit, it is explicit, and here we are treating it like implicit. Any feedback about this @AaronHeee?

### Related Issues

https://github.com/Microsoft/Recommenders/issues/597

### Checklist:

- [ ] I have added tests.

### DIFF
--- a/reco_utils/recommender/ncf/dataset.py
+++ b/reco_utils/recommender/ncf/dataset.py
@@ -12,9 +12,7 @@ from reco_utils.common.constants import (
 
 
 class Dataset(object):
-    """
-    classdocs
-    """
+    """Dataset class for NCF"""
 
     def __init__(
         self,
@@ -28,9 +26,22 @@ class Dataset(object):
         col_timestamp=DEFAULT_TIMESTAMP_COL,
         seed=42,
     ):
-        """
-        Constructor
-        """
+    """Constructor 
+    
+    Args:
+        train ([type]): [description]
+        test ([type], optional): [description]. Defaults to None.
+        n_neg (int, optional): [description]. Defaults to 4.
+        n_neg_test (int, optional): [description]. Defaults to 100.
+        col_user ([type], optional): [description]. Defaults to DEFAULT_USER_COL.
+        col_item ([type], optional): [description]. Defaults to DEFAULT_ITEM_COL.
+        col_rating ([type], optional): [description]. Defaults to DEFAULT_RATING_COL.
+        col_timestamp ([type], optional): [description]. Defaults to DEFAULT_TIMESTAMP_COL.
+        seed (int, optional): [description]. Defaults to 42.
+    
+    Returns:
+        [type]: [description]
+    """
 
         # initialize user and item index
         self.user_idx = None
--- a/reco_utils/recommender/ncf/dataset.py
+++ b/reco_utils/recommender/ncf/dataset.py
@@ -24,25 +24,25 @@ class Dataset(object):
         col_item=DEFAULT_ITEM_COL,
         col_rating=DEFAULT_RATING_COL,
         col_timestamp=DEFAULT_TIMESTAMP_COL,
+        implicit=True,
         seed=42,
     ):
     """Constructor 
     
     Args:
-        train ([type]): [description]
-        test ([type], optional): [description]. Defaults to None.
-        n_neg (int, optional): [description]. Defaults to 4.
-        n_neg_test (int, optional): [description]. Defaults to 100.
-        col_user ([type], optional): [description]. Defaults to DEFAULT_USER_COL.
-        col_item ([type], optional): [description]. Defaults to DEFAULT_ITEM_COL.
-        col_rating ([type], optional): [description]. Defaults to DEFAULT_RATING_COL.
-        col_timestamp ([type], optional): [description]. Defaults to DEFAULT_TIMESTAMP_COL.
-        seed (int, optional): [description]. Defaults to 42.
+        train (pd.DataFrame): Training data with at least columns (col_user, col_item, col_rating).
+        test (pd.DataFrame): Test data with at least columns (col_user, col_item, col_rating). test can be None, 
+            if so, we only process the training data.
+        n_neg (int): Number of negative samples for training set.
+        n_neg_test (int): Number of negative samples for test set.
+        col_user (str): User column name.
+        col_item (str): Item column name.
+        col_rating (str): Rating column name. 
+        col_timestamp (str): Timestamp column name.
+        implicit (bool): Ff true, set rating>0 to rating = 1 
+        seed (int): Seed.
     
-    Returns:
-        [type]: [description]
     """
-
         # initialize user and item index
         self.user_idx = None
         self.item_idx = None
@@ -55,24 +55,24 @@ class Dataset(object):
         self.col_rating = col_rating
         self.col_timestamp = col_timestamp
         # data preprocessing for training and test data
-        self.train, self.test = self._data_processing(train, test)
+        self.train, self.test = self._data_processing(train, test, implicit)
         # initialize negative sampling for training and test data
         self._init_train_data()
         self._init_test_data()
         # set random seed
         random.seed(seed)
 
-    def _data_processing(self, train, test, implicit=True):
-        """ process the dataset to reindex userID and itemID, also set rating as implicit feedback
+    def _data_processing(self, train, test, implicit):
+        """Process the dataset to reindex userID and itemID, also set rating as implicit feedback
 
         Args:
-            train (pandas.DataFrame): training data with at least columns (col_user, col_item, col_rating) 
-            test (pandas.DataFrame): test data with at least columns (col_user, col_item, col_rating)
-                    test can be None, if so, we only process the training data
-            implicit (bool): if true, set rating>0 to rating = 1 
+            train (pd.DataFrame): Training data with at least columns (col_user, col_item, col_rating). 
+            test (pd.DataFrame): Test data with at least columns (col_user, col_item, col_rating)
+                    test can be None, if so, we only process the training data.
+            implicit (bool): If true, set rating>0 to rating = 1.
 
         Returns:
-            list: train and test pandas.DataFrame Dataset, which have been reindexed.
+            list: train and test pd.DataFrame Dataset, which have been reindexed.
         
         """
         # If testing dataset is None
@@ -106,7 +106,7 @@ class Dataset(object):
         return self._reindex(train, implicit), self._reindex(test, implicit)
 
     def _reindex(self, df, implicit):
-        """ process dataset to reindex userID and itemID, also set rating as implicit feedback
+        """Process dataset to reindex userID and itemID, also set rating as implicit feedback
 
         Args:
             df (pandas.DataFrame): dataframe with at least columns (col_user, col_item, col_rating) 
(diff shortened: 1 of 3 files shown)


loomlike:
Here are my thought:

This is because our ranking metrics don't care about if user will like the movie or not,
while our models (w/ explicit feedback) learn like-or-not.
(note, our implementation users binary relevancy score).
 
E.g. Rating data of a user: `train = [5, 1, 1, x, x, x, x]`,  `test = [1, 1, 1, x, x, x, x, x]`,
x means the user did not watched (or not-rated).

And now let's say we have a **perfect model** M. Because M is so good, it predicts the future,
which will be `test = [1, 1, 1, 3, 4, 5, 2, 2]`.
You can see, our ranking metrics (w/ k=3) will give *F* (F for fail) to our model because the movies user watched (rated) in our test data will not be in top-3.
On the other hand, if the model learned 'watch-or-not', it would get higher ranking metrics.

loomlike:
Similarly, when I tried ALS w/ `implicit feedback = True` on Movielens, it performed better than those of ALS w/ explicit-feedback-setting.

yueguoguo:
This is related to the "negative sampling" feature in the implementation. To account for the negative (implicit) feedback, @AaronHeee has sampled 1 negative sample in the data, and this is from the movies that have been watched by the users. This is what people usually do in processing the Movielens data for implicit feedback typed settings. @AaronHeee correct me if there is anything. 

I think @loomlike comment explained well about "implicit feedback does not predict the preference (e.g., rating) but the likelihood of interaction (in our context, watch a movie)". So in the ranking metrics, the relevant items, i.e., the list of items that will be very likely to be watched by a user, will be evaluated against the actual items.

AaronHeee:
> This is related to the "negative sampling" feature in the implementation. To account for the negative (implicit) feedback, @AaronHeee has sampled 1 negative sample in the data, and this is from the movies that have been watched by the users. This is what people usually do in processing the Movielens data for implicit feedback typed settings. @AaronHeee correct me if there is anything.

That's right, and you can view the number of negative sample as a hyper-parameter, tune it for the best performance.

reco_utils/recommender/ncf/ncf_singlenode.py
```diff
@@ -204,40 +219,46 @@ def _create_model(self,):
...
     def load(self, gmf_dir=None, mlp_dir=None, neumf_dir=None, alpha=0.5):
-        """ load model parameters for further use.
-            GMF model --> load parameters in `gmf_dir`
-            MLP model --> load parameters in `mlp_dir`
-            NeuMF model --> load parameters in `neumf_dir` or in `gmf_dir` and `mlp_dir`
-            Args:
-                gmf_dir, mlp_dir, neumf_dir ( str or None ): model parameters directory name
-            Returns:
-                load parameters in this model
+        """Load model parameters for further use.
```
gramhagen: it seems like it would be better to have a static method to load the model, where the model type is specified, and a single directory is provided. even if it's kept as a class method which uses model_type defined, I still think there should only be a single directory given.

miguelgfierro: I was looking how other libs do the loading and saving, in LightGBM they do:
```
lgb_model.save_model(save_file)
loaded_model = lgb.Booster(model_file=save_file)
```
Are you thinking on something like this or different?

gramhagen: yeah, I think there are two straightforward ways with the current implementation:
```
old_model.save(model_dir=model_dir)
// keep load as an instance method
model = NCF(model_type="NeuMF").load(model_dir=model_dir)
... or ...
// make load a class method
model = NCF.load(model_type="NeuMF", model_dir=model_dir)
```
It seems like less work to go with option 1, plus it will let you modify the model parameters if needed and just reload the weights (I'm assuming that's what tf.train.Saver().restore does anyway)

either way the model_dir variable doesn't really need to be tied to the type of model

reco_utils/recommender/ncf/dataset.py
```diff
@@ -26,12 +24,25 @@ def __init__(
...
+            train (pd.DataFrame): Training data with at least columns (col_user, col_item, col_rating).
+            test (pd.DataFrame): Test data with at least columns (col_user, col_item, col_rating). test can be None, 
+                if so, we only process the training data.
+            n_neg (int): Number of negative samples for training set.
+            n_neg_test (int): Number of negative samples for test set.
+            col_user (str): User column name.
+            col_item (str): Item column name.
+            col_rating (str): Rating column name. 
+            col_timestamp (str): Timestamp column name.
+            implicit (bool): If true, set rating > 0 to rating = 1.
```
anargyri: Do you mean "If true, set rating > 0 else set rating = 1." ?

anargyri:
@miguelgfierro I think there is an issue that relates to the benchmark more generally and not only to NCF. There are a couple of instances where it is not explicit what happens and also it is not explicit that we don't violate any machine learning "firewall". Specifically, in line
`train, test = prepare_metrics_data.get(algo, lambda x,y:(x,y))(df_train, df_test)`
one would be more confident if we separated this in two lines, one with `df_train` only and one with `df_test`, unless there is a good reason. 
And line 
`top_k_scores, time_ranking = ranking_predictor[algo](model, test, train)`
should not have `test` normally. What happens here implicitly is not transparent to the reader (and indeed someone may find it suspect).

miguelgfierro:
@anargyri, I know what you mean, the problem is that it is very difficult to have a common interface for all the algos. I struggle a lot when creating the interfaces, and had to do several passes. Many times it is due to how the different functions are implemented.

About the other comment, do you propose something like:
```
train = prepare_metrics_train_set.get(algo, lambda x:x)(df_train)
test = prepare_metrics_test_set.get(algo, lambda  x:x)(df_test)
```

anargyri:
> @anargyri, I know what you mean, the problem is that it is very difficult to have a common interface for all the algos. I struggled a lot when creating the interfaces, and had to do several passes. Many times it is due to how the different functions are implemented.
> 
> About the other comment, do you propose something like:
> 
> ```
> train = prepare_metrics_train_set.get(algo, lambda x:x)(df_train)
> test = prepare_metrics_test_set.get(algo, lambda  x:x)(df_test)
> ```

Yes, this is exactly what I propose.
About the ranking_prediction, which of the algorithms requires the test set? This would be an issue with the design of the specific algorithm, and should be consistent with the others.

miguelgfierro:
> About the ranking_prediction, which of the algorithms requires the test set? This would be an issue with the design of the specific algorithm, and should be consistent with the others.

bufff, there is a mess here. In SAR for example, we are using the users in the test set to recommend items, whereas in  SVD we are using the users in the train set. I think this need a deep review I created a new issue to address this https://github.com/Microsoft/Recommenders/issues/768

anargyri:
> > About the ranking_prediction, which of the algorithms requires the test set? This would be an issue with the design of the specific algorithm, and should be consistent with the others.
> 
> bufff, there is a mess here. In SAR for example, we are using the users in the test set to recommend items, whereas in SVD we are using the users in the train set. I think this need a deep review I created a new issue to address this #768

Totally agree that we should strive to make the different algos consistent to a common syntax.
However, the SAR method accepts any dataframe you give it, it doesn't need to be the test data set. So for the benchmark you just need to pass a dataframe that contains the users for which you want recommendations. You could compute this dataframe on a separate line.

miguelgfierro:
> About the other comment, do you propose something like:
> train = prepare_metrics_train_set.get(algo, lambda x:x)(df_train)
> test = prepare_metrics_test_set.get(algo, lambda  x:x)(df_test)
> Yes, this is exactly what I propose.

I think in fact that in the ranking metrics the input should be the test set, not the train set. Basically, the test set represent the users that we want to applicate recommendations on. Therefore, after we fix https://github.com/Microsoft/Recommenders/issues/768, there should be no need for `prepare_metrics_train_set`. I think we should fix this issue before changing the code in prepare_metrics. Would this make sense?

reco_utils/recommender/ncf/dataset.py
miguelgfierro: changed

reco_utils/recommender/ncf/ncf_singlenode.py
miguelgfierro: about this, I´ve been looking at how they save the models in ligthgbm https://github.com/Microsoft/LightGBM/blob/master/examples/python-guide/advanced_example.py#L83 and sklearn https://machinelearningmastery.com/save-load-machine-learning-models-python-scikit-learn/. Basically, whatever they load, it has all the parameters the algo needs. Right now, whatever we use the current form or the static method we need to define the parameters

```
model = NCF (
    n_users=data.n_users, 
    n_items=data.n_items,
    model_type="NeuMF",
    n_factors=4,
    layer_sizes=[16,8,4],
    n_epochs=EPOCHS,
    batch_size=BATCH_SIZE,
    learning_rate=1e-3,
    verbose=10,
)
model.load(gmf_dir=".pretrain/GMF", mlp_dir=".pretrain/MLP", alpha=0.5)
```
or 
```
model = NCF (
    n_users=data.n_users, 
    n_items=data.n_items,
    model_type="NeuMF",
    n_factors=4,
    layer_sizes=[16,8,4],
    n_epochs=EPOCHS,
    batch_size=BATCH_SIZE,
    learning_rate=1e-3,
    verbose=10,
).load(gmf_dir=".pretrain/GMF", mlp_dir=".pretrain/MLP", alpha=0.5)
```

I think what we should do is this:

```
model = function_or_class_to_load(gmf_dir=".pretrain/GMF", mlp_dir=".pretrain/MLP", alpha=0.5)
```
which includes all the parameters.

Would this make sense?

loomlike:
I have some changes on ncf for `seed` setting. I will wait till this PR get merged and then work on that.

miguelgfierro:
> I have some changes on ncf for seed setting. I will wait till this PR get merged and then work on that.

Where you able to make it repeatable?

reco_utils/recommender/ncf/ncf_singlenode.py
gramhagen: that would be great, except for the part where there are multiple directories in the signature. Are both needed for some reason? I was suggesting there just be single directory input on where to look up the model. If we must know they type of model before we access that directory, then it should be a input parameter.

miguelgfierro: both are needed in the case of NeuMF, because it is a mix of GMF and MLP

gramhagen: ah i see, ok, that sounds good then

miguelgfierro: ok I created an issue about this: https://github.com/Microsoft/Recommenders/issues/771. Apart from this, is there anything missing on this PR?

yueguoguo:
The evaluation method takes both train and test as inputs and do the join of the two dataframes, and then calculate the metrics. So even if the input of evaluator is the training set, it still looks at the test data for the users and then do the calculation. 

Does this relate to your discussion above? @miguelgfierro @anargyri

anargyri:
> The evaluation method takes both train and test as inputs and do the join of the two dataframes, and then calculate the metrics. So even if the input of evaluator is the training set, it still looks at the test data for the users and then do the calculation.
> 
> Does this relate to your discussion above? @miguelgfierro @anargyri

No, we were talking about prepare_metrics utilities, which do only data prep (hence there can be separate methods for training and test data sets).

Issue #770, opened 2019-05-02
[BUG] Windows reco_base - tests\integration\test_notebooks_python.py:45


roalexan:
### Description
For the Windows DSVM test machine going against staging, I'm using the reco_base environment and running this test tests\integration\test_notebooks_python.py:45 and seeing this error:

E           papermill.exceptions.PapermillExecutionError: 
E           ---------------------------------------------------------------------------
E           Exception encountered at "In [16]":
E           ---------------------------------------------------------------------------
E           ValueError                                Traceback (most recent call last)
E           <ipython-input-16-55873c2ba276> in <module>
E                 3 
E                 4 ground_truth = test[test['userID']==user_id].sort_values(by='rating', ascending=False)[:TOP_K]
E           ----> 5 prediction = model.recommend_k_items(pd.DataFrame(dict(userID=[user_id])), remove_seen=True)
E                 6 pd.merge(ground_truth, prediction, on=['userID', 'itemID'], how='left')
E           
E           ~\agent\_work\1\s\reco_utils\recommender\sar\sar_singlenode.py in recommend_k_items(self, test, top_k, sort_top_k, remove_seen)
E               403         """
E               404 
E           --> 405         test_scores = self.score(test, remove_seen=remove_seen)
E               406 
E               407         top_items, top_scores = get_top_k_scored_items(
E           
E           ~\agent\_work\1\s\reco_utils\recommender\sar\sar_singlenode.py in score(self, test, remove_seen)
E               276         user_ids = test[self.col_user].drop_duplicates().map(self.user2index).values
E               277         if any(np.isnan(user_ids)):
E           --> 278             raise ValueError("SAR cannot score users that are not in the training set")
E               279 
E               280         # calculate raw scores with a matrix multiplication
E           
E           ValueError: SAR cannot score users that are not in the training set

### In which platform does it happen?
Azure Windows DSVM - this originated from the Windows testing machine against staging.

### How do we replicate the issue?
Use Azure Windows DSVM, install reco repo, create reco_base python environment, and run test tests\integration\test_notebooks_python.py. I'm not sure yet if this happens only occasionally. I suspect it's sporadic. Had a quick discussion with Scott - might be related to the seed not being as deterministic on Windows as on Linux - not sure yet.

### Expected behavior (i.e. solution)
Should pass consistently.

gramhagen:
[...]

PR #773, opened 2019-05-07
Gramhagen/windows sar split

gramhagen:
### Description
Moving sar quickstart notebook to use stratified splitting so all users will be in train and test sets. Also the example user 877 doesn't exist in the movielens 10m dataset so I switched it to user id 876.
Updated the test metrics to match values with new splitting approach.

### Related Issues
#770 

### Checklist:

- [X] I have added tests.


notebooks/00_quick_start/sar_movielens.ipynb
```diff
@@ -222,21 +223,21 @@
...
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 74,
    "metadata": {},
    "outputs": [],
    "source": [
-    "train, test = python_random_split(data)"
+    "train, test = python_stratified_split(data, ratio=0.75, col_user='userID', col_item='itemID', seed=42)"
```
miguelgfierro: I'm curious, why are we changing the splitter?

Issue #449, opened 2019-01-25
Create a new devops agent that is public so external people can see the results of the tests


miguelgfierro:
### *What* is affected by this bug?
Right now, external developers can't see why the tests are failing when they do a PR. 

### In *which* platform does it happen?
*For example:*
* *Azure Data Science Virtual Machine.*

### Expected behavior (i.e. solution)
we have to create a new agent, start the pipeline there and make the agent public

miguelgfierro:
related to https://github.com/Microsoft/Recommenders/issues/450

bethz:
Hey Folks, I'm ramping on DevOps. So far, I created a test DevOps pipeline to kick the tires and created a dummy pipeline off Recommenders.  I exported some of the existing tests, started setup and met with Jordan Edwards to discuss some of his plans .  I am doing a little more investigation to understand how to best integrate with AzureML.   Yea, and some more ramping on AzureML too....

bethz:
Update: I'm working with Aimhin and others on DevOps on a generalizable DevOps solution that we can clone and use for subsequent projects. I've also met with the Forecasting team about plans for DevOps.  The great news is that the DevOps team is developing new pipeline tasks that are advantageous for AML.  This includes tasks like use Python 3.6, Identify notebooks to test, copy scripts to artifacts, Azureml-infra-CI, AML Model Deploy, AML Model Register and more. All great additions and tasks we want to consume.  But, it's a couple of more weeks, until those are ready.  Here's what I'm working on while that's going on:
- I'm working with the DevOps team on providing inputs on their task development, I'll also be a good tester for their rough, initial releases
- Subscription - I'm sorting out how to handle the subscription as when we make devops public we want to make good choices re: subscription iimplication
- As this will be public, I'm also investigating issues around how we should think about that.

miguelgfierro:
this is fixed

PR #783, opened 2019-05-15
First PR:  new best-practices devops instance which uses AzureML Service to run a unit test

bethz:
### Description
This PR runs one unit test to check the path from DevOps to Github to AzureML etc. 
The Best-practices DevOps instance has a pipeline called unit_tests_staging which will be triggered by PRs to staging. I will update the trigger once the PR is accepted. 

As this version is for testing, I've left some print statements in the code to make it easier to validate setup.  Those prints will show up in the output files of the DevOps pipeline and will be removed once I confirm functionality.

The pipeline is temporarily private while testing this new change. Once the change works, it will be made public.

Docs for this will be added once it's publicly visible.

### Related Issues
#449 

### Checklist:

- [x ] My code follows the code style of this project, as detailed in our [contribution guidelines](../CONTRIBUTING.md).
- [ ] I have added tests.


tests/ci/config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "subscription_id": "<>",
+    "resource_group": "recommender",
```
gramhagen: I think it would be useful to populate this either from key-vault entries or variables in the build pipeline, that way we don't need a code change to adjust the parameters.

alternatively if the az ml cli extension is available and allows this to be downloaded programmatically that would be preferable, something like:

```
az ml workspace show -g recommender -w RecoWS -o json > config.json
```

tests/ci/runpytest.py
```diff
@@ -0,0 +1,27 @@
+import subprocess
+import os
+
+from azureml.core import Run
+print('before run.get_context')
+run = Run.get_context()
+print('before subprocess.run')
+
+# subprocess.run(["python", "-m", "pytest","tests/unit",
```
gramhagen: do we need these commented lines?

tests/ci/runpytest.py
```diff
@@ -0,0 +1,27 @@
+import subprocess
```
gramhagen: should add the standard copyright info at the top

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
...
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+MICROSOFT OR ITS LICENSORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
+BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THE SOFTWARE CODE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+"""
```
gramhagen: we don't need the full license info here the following is what we do everywhere else:
```
```

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
...
+try:
+    print("Trying to get ws")
+    ws = Workspace.get(
+        name=workspace_name,
+        subscription_id=subscription_id,
+        resource_group=resource_group,
+        auth=cli_auth,
+    )
+
+except Exception:
```
gramhagen: do we expect the workspace to be deleted? or is this just precautionary?

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
...
+        name=workspace_name,
+        subscription_id=subscription_id,
+        resource_group=resource_group,
+        # create_resource_group=True,
+        location=location,
+        auth=cli_auth,
+    )
+
+# print Workspace details
+print(ws.name, ws.resource_group, ws.location, ws.subscription_id, sep="\n")
```
gramhagen: some redundant printing of workspace params

tests/ci/runpytest.py
```diff
@@ -0,0 +1,27 @@
...
+#                "-m","not notebooks and not spark and not gpu",
+#                "--junitxml=reports/test-unit.xml"])
+# subprocess.run(["pytest","tests/unit/test_python_utils.py",
+#                "-m","not notebooks and not spark and gpu",
+#                "--junitxml=reports/test-unit.xml"])
+# subprocess.run(["pytest","tests/unit/test_dataset.py",
+#                "-m","not notebooks and not spark and not gpu",
+#                "--junitxml=reports/test-unit.xml"])
+
+subprocess.run(["pytest", "tests/unit/test_timer.py",
```
gramhagen: might be more generalizable to use argparse and pass in arguments for the markers / keywords to use with the test

tests/ci/unit_tests_staging.yml
```diff
@@ -0,0 +1,86 @@
...
+    echo "##vso[task.prependpath]/data/anaconda/bin"
+  displayName: Add Conda to PATH
+
+- script: 'pip install azureml-sdk'
+  displayName: 'install azureml-sdk'
+  continueOnError: true
+  
+- script: 
+     python scripts/generate_conda_file.py
+  displayName: ' generate_conda_file.py'
```
gramhagen: parameters for this will need to change based on using spark, gpu or base python

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
...
+run_config.auto_prepare_environment = True
+
+print("reco_base.yaml path")
+print(os.path.dirname('./reco_base.yaml'))
+print('reco_base.yaml exists ', os.path.exists('./reco_base.yaml'))
+
+# specify CondaDependencies obj
+run_config.environment.python.conda_dependencies = CondaDependencies(
+    conda_dependencies_file_path='./reco_base.yaml')
+
```
gramhagen: we may want this to be agnostic to the yaml file used (reco_base, reco_gpu, etc.) while we could name them all reco_base, I would suggest just changing this to reco.yaml, and in the generation script passing the name argument 'reco'.

tests/ci/unit_tests_staging.yml
```diff
@@ -0,0 +1,86 @@
...
+  displayName: 'Login to Azure'
+
+- script: |
+   sed -i 's#"subscription_id": "<>"#"subscription_id": "$(SubscriptionID)"#g' ./tests/ci/config.json
+   echo my subscription is $(SubscriptionID)
+   cat ./tests/ci/config.json
+  displayName: 'replace subscription value'
+
+- script: 
+   sed -i 's#"tests/unit_or_smoke_int"#"tests/unit"#g' ./tests/ci/runpytest.py
```
gramhagen: might be more straightforward to just pass these as arguments for submitpytest

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
...
+import json
+# uncomment if needed
+# import sys
+import azureml.core
+from azureml.core.authentication import AzureCliAuthentication
+from azureml.core import Workspace
+from azureml.core import Experiment
+from azureml.core.runconfig import RunConfiguration
+from azureml.core.conda_dependencies import CondaDependencies
+from azureml.core.runconfig import DEFAULT_CPU_IMAGE
```
miguelgfierro: here we should import both images and have a parameter somewhere to choose between them

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+MICROSOFT OR ITS LICENSORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
+BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THE SOFTWARE CODE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+"""
+
```
miguelgfierro: it would be nicer to have a main and add parameters with argparse, see an example in my codebase: https://github.com/miguelgfierro/codebase/blob/master/python/io_base/argument_io.py

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
...
+import os
+import json
+# uncomment if needed
```
miguelgfierro: We should avoid to ask the user to modify the scripts. If something is variable, it should come from arguments

tests/ci/runpytest.py
miguelgfierro: What is the function of this file? and how is this related to submitpytest?

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,177 @@
+"""
```
miguelgfierro: do you guys think we should change the name of this file to explicitly mention that we are submitting a test through azureml?

miguelgfierro:
@bethz this looks good, it would be nice to have instructions on TEST.md. Scott and I proposed some suggestions, after a couple iterations, I would love to try this

tests/ci/config.json
bethz: Yes, that's on the list of things to do.  The key vault is already set up and I'd like to get rid of the json file entirely.

tests/ci/runpytest.py
bethz: removed the comments

tests/ci/submitpytest.py
bethz: done

tests/ci/submitpytest.py
bethz: just precautionary. I copied it from the sample code.

tests/ci/submitpytest.py
bethz: - after the first run, all the prints will be removed so this is temp
- one print is the info in .json and one is after the ws is created (so *hopefully* the same but...)

tests/ci/runpytest.py
bethz: agree. But, I was asked to push this out now. So, let's get this out and then I can get additional items on the todo list.

tests/ci/unit_tests_staging.yml
bethz: agree. The pipeline will be different for spark, gpu or base python. I can pass in a par for that based on the pipelines once we get this first one tested but don't want to hold this up for that.

tests/ci/submitpytest.py
bethz: The proper yaml file is generated in each pipeline and agree that it's better to change to reco.yaml.  What if I do that in the next pass so I can test the first pass?

tests/ci/unit_tests_staging.yml
bethz: agree. again, I was asked to get something out now and then fix these items.

tests/ci/runpytest.py
bethz: done

bethz: This is the script that submitpytest submits to the AzureML service.  It's set up so that when pytest corrupts the AML service environment variables that the log files can still be returned to devops for publishing.

tests/ci/submitpytest.py
bethz: removed.

tests/ci/runpytest.py
miguelgfierro: I see. 

A general question in relation to this, how do you want to show all this pipelines? I think this is really nice. One option is to show this on the test markdown. Besides, this could be a nice history to showcase AzureML. Maybe we could do a blog post. What do you think?

bethz: Agree. I also set up the experiment name on the azure portal so that the pipeline name shows up on the azure portal as an experiment. With the initial submissions, there was no way to correllate the devops info with the azure portal and that was important for usage and for debugging.  I would like to make the info on the portal more useful at some point.

tests/ci/unit_tests_staging.yml
```diff
@@ -0,0 +1,91 @@
+# unit_tests_staging.yml
```
miguelgfierro: is the plan to have one yaml for each pipeline? 

the other day we were talking about having the same pipeline for staging and master. Can this be merged?

miguelgfierro:
unit test staging is failing, I think because the old and new pipelines have the new name. Maybe for the time being we could have different names for each

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,125 @@
...
+# uncomment if using gpu
+from azureml.core.runconfig import DEFAULT_GPU_IMAGE
+from azureml.core.script_run_config import ScriptRunConfig
+
+
+from azureml.core.compute import ComputeTarget, AmlCompute
+from azureml.core.compute_target import ComputeTargetException
+
+def submit_exp():
+    with open("tests/ci/config.json") as f:
```
miguelgfierro: it would be nice to have `"tests/ci/config.json"` as argument

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,125 @@
...
+    try:
+        print("Trying to get ws")
+        ws = Workspace.get(
+            name=workspace_name,
+            subscription_id=subscription_id,
+            resource_group=resource_group,
+            auth=cli_auth
+        )
+
+    except Exception:
```
miguelgfierro: do we know the exception type workspace gets if there is no workspace? it's normally a good practice to add that exception instead of the general `Exception` to have a better control on what we are doing

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,125 @@
...
+            name=ws,
+            subscription_id=subscription_id,
+            resource_group=resource_group,
+            # create_resource_group=True,
+            location=location,
+            auth=cli_auth
+        )
+
+    # Choose a name for your CPU cluster
+    cpu_cluster_name = "persistentcpu"
```
miguelgfierro: this could also be an argument

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,125 @@
...
+
+            print(" WS name ", workspace_name)
+            print("subscription_id ", subscription_id)
+            print("location",location)
+
+            cli_auth = AzureCliAuthentication()
+            print("cliauth")
+
+    try:
+        print("Trying to get ws")
```
miguelgfierro: in other parts of the code we are using logs instead of prints, here there is an example: https://github.com/microsoft/recommenders/blob/master/reco_utils/recommender/sar/sar_singlenode.py#L20

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,125 @@
...
+    print("cpu_cluster_name",cpu_cluster_name)
+    # Verify that cluster does not exist already
+    # https://docs.microsoft.com/en-us/azure/machine-learning/service/how-to-set-up-training-targets
+
+    try:
+        cpu_cluster = ComputeTarget(workspace=ws, name=cpu_cluster_name)
+        print('Found existing cluster, use it.')
+    except ComputeTargetException:
+        print("create cluster")
+        compute_config = AmlCompute.provisioning_configuration(vm_size='STANDARD_D2_V2',
```
miguelgfierro: vm size and all other parameters of the machine should be parameters

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,125 @@
...
+                                                            max_nodes=4)
+        cpu_cluster = ComputeTarget.create(ws, cpu_cluster_name, compute_config)
+
+    cpu_cluster.wait_for_completion(show_output=True)
+
+    from azureml.core.runconfig import RunConfiguration
```
miguelgfierro: in other python files we have the imports at the beginning of the file

tests/ci/submitpytest.py
```diff
@@ -0,0 +1,125 @@
...
+    from azureml.core.runconfig import DEFAULT_CPU_IMAGE
+
+    # Create a new runconfig object
+    run_amlcompute = RunConfiguration()
+
+    # Use the cpu_cluster you created above. 
+    run_amlcompute.target = cpu_cluster
+
+    # Enable Docker
+    run_amlcompute.environment.docker.enabled = True
```
miguelgfierro: all these could be parameters, so we have more control on the script

tests/ci/run_pytest.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: there are two files pretty similar `run_pytest.py` and `runpytest.py`, is this one duplicated?

tests/ci/runpytest.py
miguelgfierro: After talking with @bethz, we agreed that we don't want to have code in the repo that doesn't follow the contribution standards. It is important to make sure that everything that we put in the repo works correctly.

I'm going to talk with the LT to make sure they understand the hard work that @bethz is putting into this.

tests/ci/submit_AzureML_pytest.py
```diff
@@ -0,0 +1,154 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: there are several files very similar, `submit_AzureML_pytest.py` and `submitpytest.py`. 

Also, in the rest of the code we are following the pattern of using lowercase for filenames `submit_azureml_pytest.py`

tests/ci/submitpytest.py
bethz: done

tests/ci/submitpytest.py
bethz: done

tests/ci/submitpytest.py
bethz: I'll add it.

tests/ci/submit_AzureML_pytest.py
bethz: changed name and removed old.

tests/ci/run_pytest.py
bethz: I updated name based on a comment above. removed old file.

tests/ci/submitpytest.py
bethz: done.

tests/ci/unit_tests_staging.yml
bethz: done.

tests/ci/run_pytest.py
```diff
@@ -0,0 +1,96 @@
...
+                        help="Folder where tests are located")
+    # internal test purposes
+    parser.add_argument("--num",
+                        action="store",
+                        default="99",
+                        help="test num")
+    # test markers
+    parser.add_argument("--testmarkers", "-m",
+                        action="store",
+                        default="not notebooks and not spark and not gpu",
```
gramhagen: this will fail unless the user input is enclosed in quotes, another way might be to have flags for each option

```
parser.add_argument("--notebooks", "-n", action="store_true")
parser.add_argument("--spark", "-s", action="store_true")
parser.add_argument("--gpu", "-g", action="store_true")
...
test_marker = ' and '.join([name if getattr(args, name) else 'not ' + name for name in ['notebooks', 'spark', 'gpu'])
```

tests/ci/run_pytest.py
```diff
@@ -0,0 +1,96 @@
...
+    print("os.listdir reports", os.listdir("./reports"))
+    print("os.listdir outputs", os.listdir("./outputs"))
+
+    # Leveraged code from this  notebook:
+    # https://msdata.visualstudio.com/Vienna/_search?action=contents&text=upload_folder&type=code&lp=code-Project&filters=ProjectFilters%7BVienna%7DRepositoryFilters%7BAzureMlCli%7D&pageSize=25&sortOptions=%5B%7B%22field%22%3A%22relevance%22%2C%22sortOrder%22%3A%22desc%22%7D%5D&result=DefaultCollection%2FVienna%2FAzureMlCli%2FGBmaster%2F%2Fsrc%2Fazureml-core%2Fazureml%2Fcore%2Frun.py
+
+
+if __name__ == "__main__":
+
+    args = create_arg_parser()
```
gramhagen: no need to create a separate method for argument parsing

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,142 @@
...
+
+    run_amlcompute.environment.python.conda_dependencies = CondaDependencies(
+            conda_dependencies_file_path='./reco.yaml')
+
+    experiment_name = 'unit_tests_staging-d3-v2'
+
+    experiment = Experiment(workspace=ws, name=experiment_name)
+    project_folder = "."
+    # start_time = time.time
+    script_run_config = ScriptRunConfig(source_directory=project_folder,
```
gramhagen: need to add args here right?

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,142 @@
...
+from azureml.core.compute import ComputeTarget, AmlCompute
+from azureml.core.compute_target import ComputeTargetException
+
+HELP_MSG = """
+This program sets up an AzureML environment and submits a pytest script on aml.
+$ python dir/setup_AzureML_script_to_submits_pytest.py
+"""
+
+
+def submit_exp():
```
gramhagen: no need for separate function

also need argument parsing here too, in order to pass them onto run_pytest

tests/ci/submitpytest.py
bethz: addeded logging.

tests/ci/run_pytest.py
bethz: There's a difference as the args are passed to AML as part of ScriptRunConfig.

To demonstrate this works,  testmarkers are deleted from the devops pipeline where python is called and it then uses the parser.add_argument:
 https://dev.azure.com/best-practices/recommenders/_build/results?buildId=1997&view=ms.vss-test-web.build-test-results-tab

tests/ci/run_pytest.py
bethz: I'm using arg.parameter to access the args.

tests/ci/submit_azureml_pytest.py
bethz: I pushed a stale copy.

bethz: checked in stale copy

tests/ci/submit_azureml_pytest.py
bethz: same stale copy.

tests/ci/config.json
bethz: done

tests/ci/submitpytest.py
bethz: added logging.

tests/ci/runpytest.py
bethz: added argparse for parameters.

tests/ci/unit_tests_staging.yml
bethz: deleted sed. Added needed parameters.

tests/ci/submitpytest.py
bethz: done

tests/ci/submitpytest.py
bethz: done

bethz:
about this:
> @bethz this looks good, it would be nice to have instructions on TEST.md. Scott and I proposed some suggestions, after a couple iterations, I would love to try this

I think I've answered all the issues raised so far. I also started info for docs on the repo wiki. It would to talk about what else to include for docs.

tests/ci/cpu_unit_tests.yml
```diff
@@ -0,0 +1,101 @@
+# cpu_unit_tests.yml
```
gramhagen: i would remove this line and add in the standard microsoft license header

tests/ci/cpu_unit_tests.yml
```diff
@@ -0,0 +1,101 @@
...
+#  value : 'STANDARD_NC6'
+# CPU
+- name  : 'vmsize'
+  value : 'STANDARD_D3_V2'
+- name  : 'dockerproc'
+  value : 'cpu'
+- name  : 'expname'
+  value : 'cpu_unit_tests'
+
+#trigger:
```
gramhagen: do we need to keep this? should it change to the correct branch?

tests/ci/cpu_unit_tests.yml
```diff
@@ -0,0 +1,101 @@
...
+- script: 
+     python scripts/generate_conda_file.py --name reco 
+  displayName: ' generate_conda_file.py'
+
+- script: |
+     chmod +x scripts/*.py
+     chmod +x tests/ci/*.py
+  displayName: 'ls'
+
+- script: |
```
gramhagen: why not combine this with azureml-sdk pip install?

tests/ci/cpu_unit_tests.yml
```diff
@@ -0,0 +1,101 @@
...
+   
+  displayName: 'Login to Azure'
+
+- bash: |
+    echo "##vso[task.prependpath]/data/anaconda/bin"
+  displayName: Add Conda to PATH
+
+- script: 
+     pip install azureml-sdk
+  continueOnError: true
```
gramhagen: curious why we want to continue if this fails?

tests/ci/cpu_unit_tests.yml
bethz: done

tests/ci/cpu_unit_tests.yml
bethz: good point. I changed it to trigger none so it will trigger off PRs and not branch changes too. Although, we may want to talk about adding a dev branch trigger for use by Microsoft devs at some point.

bethz: done.

tests/ci/cpu_unit_tests.yml
bethz: I put the chmod in it's own task so it's easier to know what went wrong if something hiccups. If they are separate, it's trivial to know the issue. If they are together, it means reading the pipeline logs and those are not as kind.

tests/ci/cpu_unit_tests.yml
bethz: good point. Initially, I did it for debugging.
I removed it. I'll rerun tests to ensure no impact. Shouldn't be any but....

tests/ci/cpu_unit_tests.yml
```diff
@@ -0,0 +1,90 @@
...
+#  value : 'recommender'
+#- name  : 'wsname'
+#  value : 'RecoWS'
+- name  : 'reponame'
+  value : 'Recommenders'
+- name  : 'branch'
+  value : 'azure-pipelines-bz'
+- name  : 'clustername'
+  value : 'Reco-cpu-test'
+# GPU
```
miguelgfierro: if this is only for CPU, then the GPU content should be erased

tests/ci/cpu_unit_tests.yml
```diff
@@ -0,0 +1,90 @@
...
+  displayName: ' generate_conda_file.py'
+
+- script: |
+     chmod +x scripts/*.py
+     chmod +x tests/ci/*.py
+  displayName: 'ls'
+
+- script: |
+     pip install azure-cli
+     pip install --upgrade azureml-sdk[cli]
```
miguelgfierro: probably we would save some time if we mix `pip install --upgrade azureml-sdk[cli]` with the previous `pip install azureml-sdk`

tests/ci/run_pytest.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def create_arg_parser():
+    parser = argparse.ArgumentParser(description='Process inputs')
+    # test folder
+    parser.add_argument("--testfolder", "-f",
+                        action="store",
+                        default="./tests/unit",
+                        help="Folder where tests are located")
+    # internal test purposes
```
miguelgfierro: this parameter is temporal and you plan to erase it?

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,394 @@
...
+            logger.debug("Creating new workspace")
+            ws = Workspace.create(
+                name=workspace_name,
+                subscription_id=subscription_id,
+                resource_group=resource_group,
+                # create_resource_group=True,
+                location=location,
+                auth=cli_auth
+            )
+    return(ws)
```
miguelgfierro: there is no need for parentesis, this could be `return ws`

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,394 @@
...
+        resource_group  (str): Azure Resource Groups are logical collections of
+                         assets associated with a project. Resource groups
+                         make it easy to track or delete all resources
+                         associated with a project by tracking or deleting
+                         the Resource group.
+        cli_auth         Azure authentication
+        location        (str): Resource location on Azure such as EastUS
+
+    Returns:
+        ws: workspace reference
```
miguelgfierro: in docstrings we put the type of the return object. If what it is returned is an string, this should be:
```
Returns:
    str: Workspace reference

```

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,394 @@
...
+    parser.add_argument("--expname",
+                        action="store",
+                        default="persistentAML",
+                        help="experiment name on Azure")
+    # Azure datacenter location
+    parser.add_argument("--location",
+                        default="EastUS",
+                        help="Azure location")
+    # github repo, stored in AzureML experiment for info purposes
+    parser.add_argument("--reponame",
```
miguelgfierro: I really like this, it can be generalized easily to other repos

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,394 @@
...
+
+    cli_auth = AzureCliAuthentication()
+
+    workspace = setup_workspace(workspace_name=args.wsname,
+                                subscription_id=args.subid,
+                                resource_group=args.rg,
+                                cli_auth=cli_auth,
+                                location=args.location)
+
+    cpu_cluster = setup_persistent_compute_target(
```
miguelgfierro: this would be `cluster`? we are going to have 3 kinds of environments: cpu, gpu, spark, will this be correct?

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,394 @@
...
+    return(args)
+
+
+if __name__ == "__main__":
+    logger = logging.getLogger("submit_azureml_pytest.py")
+    # logger.setLevel(logging.DEBUG)
+    # logging.basicConfig(level=logging.DEBUG)
+    args = create_arg_parser()
+
+    if (args.dockerproc == "cpu"):
```
miguelgfierro: there is no need for parenthesis in the if: `if args.dockerproc == "cpu":`

tests/ci/submit_azureml_pytest.py
```diff
@@ -0,0 +1,394 @@
...
+        from azureml.core.runconfig import DEFAULT_CPU_IMAGE
+        docker_proc_type = DEFAULT_CPU_IMAGE
+    else:
```
miguelgfierro: I guess the spark option should be added right?

tests/ci/cpu_unit_tests.yml
```diff
@@ -0,0 +1,90 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro: are you planning to add the gpu and spark files in this PR or in a different one? We have 3 environments, 2 OS, 2 branches, and 3 kinds of tests (unit, notebooks and nightly). How are we going to setup all this?

bethz: Different one. My plan is to get this PR in, then I've got others waiting behind this one for:
- template - to abstract common code for the other items you mention
- job - another useful pipeline abstraction I have ready to submit
- pipelines on other branch - I've got another branch where I set up the 3 envs and 3 kinds of tests. I can show you those.  Spark has some issues that I believe are setup issues. The others have a few test fails. I want to get this PR in and then those will follow. 
- About Windows: AzureML only runs linux.  We'll need to

tests/ci/submit_azureml_pytest.py
bethz: done

tests/ci/submit_azureml_pytest.py
bethz: Yes, but this is a persistent compute definition in AzureML which is a little different than an Azure compute (aka VM).  In AML, the act of defining vmsize and num of nodes results in a persistent compute.  AML handles the spin up/down of the the compute resource. The test code does not need to handle that. It also enables fan out in pipelines. For example, due to this feature smoke and integration code start with the same pipeline set up and then fan out to run in parallel. This is very convenient and saves a bit of time.

tests/ci/submit_azureml_pytest.py
bethz: The scripts were written so they could be very easily ported to other repos.

tests/ci/submit_azureml_pytest.py
bethz: done

tests/ci/submit_azureml_pytest.py
bethz: old habits ;-)
done. Fixed all of them.

tests/ci/run_pytest.py
bethz: After this pr, I would like to finish adding the capability to take the unique pipeline number on DevOps and tag it in AML so there is a correlation between experiments on AML and pipeline runs.  That's a "nice to have" for others who clone this work. 

I removed the comment and kept the argument.

tests/ci/cpu_unit_tests.yml
bethz: In the next pr, common yml code is abstracted into a template.   I can test if time is saved and change it there.  As the template will be used in each pipeline, it's easier to make the change once.

tests/ci/cpu_unit_tests.yml
bethz: Because the intent is for other projects to clone, I included both cpu and gpu definitions and commented out the unused one to make it easier for them to switch as needed.

As requested, it's been deleted

tests/ci/cpu_unit_tests.yml
bethz: About Windows: AzureML only runs linux. We'll need to talk about what to do with Windows.

PR #785, opened 2019-05-17
DNN random seed (NCF, xDeepFM, W&D)

loomlike:
### Description
Refactor NCF (dataset and model) and W&D (input_fns) to accept random seed.
Update tests accordingly

### Related Issues
#736 

### Checklist:
- [x] I have added tests.


reco_utils/common/gpu_utils.py
```diff
@@ -23,6 +24,40 @@ def get_number_gpus():
         return 0
 
 
+def get_gpu_info():
```
miguelgfierro: I would recommend to use the numba api, it is much faster than nvidia-smi. Here you have an example of what you want: https://github.com/miguelgfierro/codebase/blob/master/python/system/memory.py#L59

reco_utils/common/tf_utils.py
```diff
@@ -1,10 +1,16 @@
...
+import collections
 import itertools
-import tensorflow as tf
+
 import numpy as np
 import pandas as pd
+import tensorflow as tf
+from tensorflow.python.estimator.inputs.queues import feeding_functions
+from tensorflow.python.estimator.inputs.numpy_io import (
+    _get_unique_target_key,
```
miguelgfierro: this hurts a little bit lol, but I guess it's not our fault. I guess there are no public functions that we can use right?

reco_utils/common/tf_utils.py
```diff
@@ -49,15 +57,122 @@ def pandas_input_fn(
...
+                shape_of_y = y.shape
+            else:
+                shape_of_y = {k: ordered_dict_data[k].shape for k in target_keys}
+
+            raise ValueError('Length of tensors in x and y is mismatched. All '
+                             'elements in x and y must have the same length.\n'
+                             'Shapes in x: {}\n'
+                             'Shapes in y: {}\n'.format(shape_dict_of_x, shape_of_y))
+
+        queue = feeding_functions._enqueue_data(  # pylint: disable=protected-access
```
miguelgfierro: is this doing some kind of prefetching?

tests/integration/test_notebooks_gpu.py
```diff
@@ -41,7 +42,7 @@ def test_ncf_integration(notebooks, size, epochs, expected_values):
         OUTPUT_NOTEBOOK,
         kernel_name=KERNEL_NAME,
         parameters=dict(
-            TOP_K=10, MOVIELENS_DATA_SIZE=size, EPOCHS=epochs, BATCH_SIZE=512
+            TOP_K=10, MOVIELENS_DATA_SIZE=size, EPOCHS=epochs, BATCH_SIZE=512, SEED=SEED,
```
miguelgfierro: I would argue that seed should be a parameters, we are getting the metrics we get because we have seed=42, right? with another seed we will get different metrics. Would this make sense?

FYI @yueguoguo @anargyri @gramhagen

After reviewing this, we should rethink if we need SEED in the constants

reco_utils/common/tf_utils.py
loomlike: ah... yeah. Not sure why they did not expose the seed variable to numpy input function.
If we really don't want to do this, alternative way is to shuffle by ourselves like:

```
def numpy_input_fn(x, ...):
    ...
    if shuffle==True and seed is not None:
        # set random seed and shuffle x
        ....
        # then call tf's numpy_input_fn with the shuffled x
        tf.estimator.inputs.numpy_input_fn(x)
```

I actually like this solution. Much cleaner.

reco_utils/common/tf_utils.py
loomlike: Yes, but with the new approach I replied to your previous comment, which is simply adding a wrapper for `tf.estimator.inputs.numpy_input_fn` to support shuffle w/ seed, we can hide all of this code. Let me know if that make sense

tests/integration/test_notebooks_gpu.py
loomlike: You mean, we should directly put 42 instead of using `constant.SEED = 42`?

reco_utils/common/tf_utils.py
loomlike: Ah... the thing is, w/ this, we cannot shuffle for every epoch... hmmm

tests/integration/test_notebooks_gpu.py
miguelgfierro: yeah

reco_utils/common/tf_utils.py
miguelgfierro: cool

reco_utils/common/tf_utils.py
loomlike: I think ‘tf.set_random_seed’ will work for this case. Let me try and will update codes if works.

loomlike: Refactored to use `tf.data.dataset` (newer API) within our `pandas_input_fn` and now much cleaner.

tests/integration/test_notebooks_gpu.py
loomlike: Changed to set `42` via test parameters in integration tests.

PR #789, opened 2019-05-17
adding normalization for SAR scoring

gramhagen:
### Description
Adding optional normalization to SAR scoring output

This lets you normalize (using L1 norm) the outputs for each user, for example prior you might get the following scores for all items and a given user:

|user|item_1|item_2|item_3|
|-|-|-|-|
|Bob|3|20|2|

now you will get the following if normalized flag is set to True

|user|item_1|item_2|item_3|
|-|-|-|-|
|Bob|0.12|0.8|0.08|

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/recommender/sar/sar_singlenode.py
+++ b/reco_utils/recommender/sar/sar_singlenode.py
@@ -5,7 +5,6 @@
 import numpy as np
 import pandas as pd
 import logging
-from sklearn.preprocessing import normalize
 from scipy import sparse
 
 from reco_utils.common.python_utils import (
@@ -39,6 +38,7 @@ class SARSingleNode:
         time_now=None,
         timedecay_formula=False,
         threshold=1,
+        normalize=False,
     ):
         """Initialize model parameters
 
@@ -53,6 +53,7 @@ class SARSingleNode:
             time_now (int | None): current time for time decay calculation
             timedecay_formula (bool): flag to apply time decay
             threshold (int): item-item co-occurrences below this threshold will be removed
+            normalize (bool): option for normalizing predictions to scale of original ratings
         """
         self.col_rating = col_rating
         self.col_item = col_item
@@ -79,6 +80,11 @@ class SARSingleNode:
         if self.threshold <= 0:
             raise ValueError("Threshold cannot be < 1")
 
+        # set flag to capture unity-rating user-affinity matrix for scaling scores
+        self.normalize = normalize
+        self.col_unity_rating = "_unity_rating"
+        self.unity_user_affinity = None
+
         # column for mapping user / item ids to internal indices
         self.col_item_id = "_indexed_items"
         self.col_user_id = "_indexed_users"
@@ -97,7 +103,7 @@ class SARSingleNode:
         # track user-item pairs seen during training
         self.seen_items = None
 
-    def compute_affinity_matrix(self, df, n_users, n_items):
+    def compute_affinity_matrix(self, df, rating_col):
         """ Affinity matrix
         The user-affinity matrix can be constructed by treating the users and items as
         indices in a sparse matrix, and the events as the data. Here, we're treating
@@ -105,28 +111,50 @@ class SARSingleNode:
         formats to de-duplicate user-item pairs, otherwise they will get added up.
         
         Args:
-            df (pd.DataFrame): Indexed df of users and items.
-            n_users (int): Number of users.
-            n_items (int): Number of items.
+            df (pd.DataFrame): Indexed df of users and items
+            rating_col (str): Name of column to use for ratings
 
         Returns:
             sparse.csr: Affinity matrix in Compressed Sparse Row (CSR) format.
         """
 
         return sparse.coo_matrix(
-            (df[self.col_rating], (df[self.col_user_id], df[self.col_item_id])),
-            shape=(n_users, n_items),
+            (df[rating_col], (df[self.col_user_id], df[self.col_item_id])),
+            shape=(self.n_users, self.n_items),
         ).tocsr()
 
-    def compute_coocurrence_matrix(self, df, n_users, n_items):
+    def compute_time_decay(self, df, decay_column):
+        """Compute time decay on provided column
+
+        Args:
+            df (pd.DataFrame): DataFrame of users and items
+            decay_column (str): column to decay
+
+        Returns:
+            DataFrame: with column decayed
+        """
+
+        # if time_now is None use the latest time
+        if not self.time_now:
+            self.time_now = df[self.col_timestamp].max()
+
+        # apply time decay to each rating
+        df[decay_column] *= exponential_decay(
+            value=df[self.col_timestamp],
+            max_val=self.time_now,
+            half_life=self.time_decay_half_life,
+        )
+
+        # group time decayed ratings by user-item and take the sum as the user-item affinity
+        return df.groupby([self.col_user, self.col_item]).sum().reset_index()
+
+    def compute_coocurrence_matrix(self, df):
         """ Co-occurrence matrix
         C = U'.transpose() * U'
         where U' is the user_affinity matrix with 1's as values (instead of ratings).
 
         Args:
-            df (pd.DataFrame): Indexed df of users and items.
-            n_users (int): Number of users.
-            n_items (int): Number of items.
+            df (pd.DataFrame): DataFrame of users and items
 
         Returns:
             np.array: Co-occurrence matrix
@@ -134,7 +162,7 @@ class SARSingleNode:
 
         user_item_hits = sparse.coo_matrix(
             (np.repeat(1, df.shape[0]), (df[self.col_user_id], df[self.col_item_id])),
-            shape=(n_users, n_items),
+            shape=(self.n_users, self.n_items),
         ).tocsr()
 
         item_cooccurrence = user_item_hits.transpose().dot(user_item_hits)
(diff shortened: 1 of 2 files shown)


anargyri:
@miguelgfierro we had a document on ways to do this, right?

anargyri:
We could also have an option to renormalize in a certain range e.g. [0, 5] instead of [0, 1].

anargyri:
I wonder, if this is the right kind of obtaining normalized scores from the outputs of SAR. This normalization depends on the number of items, which can change over time. But I think the rating of e.g. a movie should not depend on how many movies are in the database.

gramhagen:
normalization range is a good idea, although it would be fairly easy for the user to multiply them once they are in a range of [0, 1] and would be one less additional parameter to add in for this edge case

the number of items will be fixed at train
at prediction this implementation of sar does let you use unseen items, but it maps them all to a single unseen column before prediction. so the item count shouldn't be a problem

other approaches for normalization can be quite expensive, so this is the justification for a just normalizing over all items for each user.

anargyri:
> normalization range is a good idea, although it would be fairly easy for the user to multiply them once they are in a range of [0, 1] and would be one less additional parameter to add in for this edge case
> 
Agree.

gramhagen:
Capturing results of offline discussion here.

Prior work was laid out to enable SAR to produce ratings in the same range as the original ratings. This provides a more generally useful approach than L1 normalization across all items, as it independently scales each user-item score and provides a result that can be directly related to original rating values (which is useful for offline rating-based metrics like RMSE, MAE and others).

This is accomplished by computing the user-item SAR score as usual, then dividing by a normalizing factor. That factor can be calculated by first forming a unity-rating user-affinity vector (basically, computing the user-affinity vector assuming all user ratings were equal to 1). This includes applying the same time-decay as was done for the original affinity vector. Then the dot product between unity-rating user-affinity vector and the item similarity vectors can be taken to provide the normalizing constant.

Next steps here are to implement this scaling approach. The easiest path forward would be to maintain a new unity user-affinity matrix calculated during fit. This would increase memory usage of the algorithm, so perhaps there's a flag at algorithm construction that can enable/disable this.

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -282,6 +286,18 @@ def score(self, test, remove_seen=False):
         if isinstance(test_scores, sparse.spmatrix):
             test_scores = test_scores.toarray()
 
+        if normalized:
+            if remove_seen:
```
yueguoguo: `remove_seen` only applies when `normalized` is `True`?

gramhagen: this changed, now they are more clearly independent

gramhagen:
@anargyri if you can take a look at the updated implementation, I think this should be done

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -94,44 +103,66 @@ def __init__(
...
+        Args:
+            df (pd.DataFrame): DataFrame of users and items
+            decay_column (str): column to decay
+
+        Returns:
+            DataFrame: with column decayed
+        """
+
+        # if time_now is None use the latest time
+        if not self.time_now:
```
miguelgfierro: here it would be more correct to do: `if self.time_now is None`

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -250,13 +273,14 @@ def fit(self, df):
...
 
         Args:
             test (pd.DataFrame): user to test
             remove_seen (bool): flag to remove items seen in training from recommendation
-
+            normalize (bool): flag to normalize scores to be in the same scale as the original ratings
+ 1
         Returns:
             np.ndarray
         """
```
miguelgfierro: I think we can remove the todo line `# TODO: only compute scores for users in test`

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -282,6 +306,18 @@ def score(self, test, remove_seen=False):
         if isinstance(test_scores, sparse.spmatrix):
             test_scores = test_scores.toarray()
 
+        if normalize:
+            if self.unity_user_affinity is None:
```
miguelgfierro: I think this if won't trigger, because of:

```
if self.normalize:
            logger.info("Calculating normalization factors")
            temp_df[self.col_unity_rating] = 1.0
            if self.time_decay_flag:
                temp_df = self.compute_time_decay(df=temp_df, decay_column=self.col_unity_rating)
            self.unity_user_affinity = self.compute_affinity_matrix(df=temp_df, rating_col=self.col_unity_rating)
```

miguelgfierro:
> Next steps here are to implement this scaling approach. The easiest path forward would be to maintain a new unity user-affinity matrix calculated during fit. This would increase memory usage of the algorithm, so perhaps there's a flag at algorithm construction that can enable/disable this.

What is the roadmap of this? if you are going to implement RegressionSAR, do you plan to have SAR with the normalization option and implement RegressionSAR as an independent algo?

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: i wanted to ensure that self.normalize was also set, maybe it would be better to check if self.normalize was set instead of self.unity_user_affinity

gramhagen:
> > Next steps here are to implement this scaling approach. The easiest path forward would be to maintain a new unity user-affinity matrix calculated during fit. This would increase memory usage of the algorithm, so perhaps there's a flag at algorithm construction that can enable/disable this.
> 
> What is the roadmap of this? if you are going to implement RegressionSAR, do you plan to have SAR with the normalization option and implement RegressionSAR as an independent algo?

good question, I hadn't really considered additional work in the regression direction beyond what is implemented here. I think it makes sense to keep that as an independent algorithm / implementation.

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: That is still an improvement that would dramatically reduce computation cost, unfortunately there's a bit of book-keeping necessary to pull it off correctly. Maybe logging it as an issue for future enhancement would be better than inline comments?

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -378,20 +421,23 @@ def get_item_based_topk(self, items, top_k=10, sort_top_k=False):
         # drop invalid items
         return df.replace(-np.inf, np.nan).dropna()
 
-    def recommend_k_items(self, test, top_k=10, sort_top_k=False, remove_seen=False):
+    def recommend_k_items(
+        self, test, top_k=10, sort_top_k=False, remove_seen=False, normalize=False
```
anargyri: Actually, `normalize` could be omitted from the arguments; it is not needed for computing top-k recommendations.

anargyri: I would argue for sorting the top-k results by default. Usually people want to know the predicted ranking.

anargyri:
> > > Next steps here are to implement this scaling approach. The easiest path forward would be to maintain a new unity user-affinity matrix calculated during fit. This would increase memory usage of the algorithm, so perhaps there's a flag at algorithm construction that can enable/disable this.
> > 
> > 
> > What is the roadmap of this? if you are going to implement RegressionSAR, do you plan to have SAR with the normalization option and implement RegressionSAR as an independent algo?
> 
> good question, I hadn't really considered additional work in the regression direction beyond what is implemented here. I think it makes sense to keep that as an independent algorithm / implementation.

We could separate completely the rescaling and regression (i.e. `score()`) from `fit()`. In this way, if one does not want to compute RMSE etc., the overall time would be less. I don't think you need to compute `unity_user_affinity` inside `fit()` anyway. Also, someone may prefer getting ratings with a regression method rather than rescaling, so `unity_user_affinity` is unnecessary in this case.

The regression can be any method that fits the unnormalized scores to the ground truth. E.g. any regression method from sklearn. So, we don't really need to have a method for this, maybe showing how it is done is enough.

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -358,6 +395,12 @@ def get_item_based_topk(self, items, top_k=10, sort_top_k=False):
         # calculate raw scores with a matrix multiplication
         test_scores = pseudo_affinity.dot(self.item_similarity)
 
+        if normalize:
```
anargyri: I would suggest we use the rescaling only for obtaining ratings from the original SAR scores. The recommendations of items here and in `recommend_k_items()` would be based on the original matrices.

anargyri:
One thing to note is that rescaling does not preserve the ranking of the items for each user. I.e. the original SAR scores are 15, 6, 3, 8, the normalized scores could be 5, 4, 1, 3, so recommending top-2 items would differ depending on which scores you use.
So, for recommending top-k we should probably use the original scores, but for rating metrics we should use the normalized scores.
With regression you could have the same problem, but it is possible to constrain the function to be monotonic and in particular the linear function will be such.

gramhagen:
> > > > Next steps here are to implement this scaling approach. The easiest path forward would be to maintain a new unity user-affinity matrix calculated during fit. This would increase memory usage of the algorithm, so perhaps there's a flag at algorithm construction that can enable/disable this.
> > > 
> > > 
> > > What is the roadmap of this? if you are going to implement RegressionSAR, do you plan to have SAR with the normalization option and implement RegressionSAR as an independent algo?
> > 
> > 
> > good question, I hadn't really considered additional work in the regression direction beyond what is implemented here. I think it makes sense to keep that as an independent algorithm / implementation.
> 
> We could separate completely the rescaling and regression (i.e. `score()`) from `fit()`. In this way, if one does not want to compute RMSE etc., the overall time would be less. I don't think you need to compute `unity_user_affinity` inside `fit()` anyway. Also, someone may prefer getting ratings with a regression method rather than rescaling, so `unity_user_affinity` is unnecessary in this case.
> 
> The regression can be any method that fits the unnormalized scores to the ground truth. E.g. any regression method from sklearn. So, we don't really need to have a method for this, maybe showing how it is done is enough.

the issue is to properly capture the time decayed inputs the values must be captured during fit, or some other method that accesses the same information provided during fit.

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: yeah, this makes sense, i'll take it out except for explicit use during scoring

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: yep, will do

reco_utils/recommender/sar/sar_singlenode.py
gramhagen: yep, makes sense

PR #790, opened 2019-05-21
Paper contents

loomlike:
Here are my initial thoughts. The changes are just my suggestions, so feel free to give any feedbacks on them.

* Reco algorithms to Backgrounds
    We may want to spend more contents on our repo contents,
    not recommenders algorithms themselves.
    So, I put reco algos to Background section which maybe put at the
    end of the paper.

* Deep-dive and hyperparameter tuning section
    We can put those two topics under a subsection like 'Model selection
    and optimization', as the purpose of deep-dive is to help readers
    understand the model and select whichever fits to their own problems.

* We may want to have a symbolic name of the repository
    Maybe Microsoft/Recommenders or just Recommenders (with capital R)

### DIFF
--- a/publications/recommender.tex
+++ b/publications/recommender.tex
@@ -29,17 +29,20 @@
 
 \title{Microsoft/Recommenders: Best practices for building industry-grade recommender system}
 
-\author{\name Marina Meil \email mmp@stat.washington.edu \\
-       \addr Department of Statistics\\
-       University of Washington\\
-       Seattle, WA 98195-4322, USA
+\author{\name Le Zhang \email your@email.com \\
+       \addr One Microsoft Way \\
+       Redmond, WA 98052, USA
        \AND
-       \name Michael I.\ Jordan \email jordan@cs.berkeley.edu \\
-       \addr Division of Computer Science and Department of Statistics\\
-       University of California\\
-       Berkeley, CA 94720-1776, USA}
+       \name Jun-Ki Min \email jun.min@microsoft.com \\
+       \addr 1 Memorial Dr \\
+       Cambridge, MA 02142, USA
+       \AND
+       \name Authors \email your email \\
+       \addr your address line 1 \\
+       \addr your address line 2 \\
+       \addr your address line 3}
 
-\editor{Kevin Murphy and Bernhard Scholkopf}
+\editor{editors here}
 
 \maketitle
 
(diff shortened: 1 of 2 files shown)


publications/recommender.tex
```diff
@@ -29,17 +29,20 @@
 
 \title{Microsoft/Recommenders: Best practices for building industry-grade recommender system}
```
miguelgfierro: question, do we want to name this as Microsoft/Recommenders or Microsoft Recommenders?

miguelgfierro: ...for building industry-grade recommender systems (add final s)

publications/recommender.tex
```diff
@@ -52,21 +55,46 @@
 \end{keywords}
 
 \section{Introduction}
```
miguelgfierro: there is no abstract in this journal?

publications/recommender.tex
```diff
@@ -29,17 +29,20 @@
...
+       \addr 1 Memorial Dr \\
+       Cambridge, MA 02142, USA
+       \AND
+       \name Authors \email your email \\
+       \addr your address line 1 \\
+       \addr your address line 2 \\
+       \addr your address line 3}
 
-\editor{Kevin Murphy and Bernhard Scholkopf}
+\editor{editors here}
```
miguelgfierro: I don't think we need editors, do we?

publications/recommender.tex
```diff
@@ -52,21 +55,46 @@
...
 \item \emph{The repository provides an open, transparent, and collaborative platform for academic and open-source community for contributing and testing their own algorithms.} A major design principle of the repository is openness, to facilitate collaborations among researchers and developers in the field. More than 20 people contributed to the repository [4]. This is made available by the reusable and extensible utility functions. The recommenders repository became a well-recognized resource to acquire knowledge about recommendation engine beyond the basics. As of early May 2019, the repository has more than 3000 stars on Github and is forked close to 400 times. It was also covered and recognized by Hacker News, KDNuggets and O’Reilly Data Newsletter
```
miguelgfierro: About this:

> More than 20 people contributed to the repository [4]. This is made available by the reusable and extensible utility functions. The recommenders repository became a well-recognized resource to acquire knowledge about recommendation engine beyond the basics. As of early May 2019, the repository has more than 3000 stars on Github and is forked close to 400 times. It was also covered and recognized by Hacker News, KDNuggets and O’Reilly Data Newsletter

Hopefully, this will change very soon, and the numbers won't be accurate, are we sure we want to add them?

publications/recommender.tex
```diff
@@ -52,21 +55,46 @@
...
+\end{enumerate}
+
+
+\subsection{Data preparation and model evaluation}
```
miguelgfierro: maybe we could consider separating data prep and evaluation in different sections, probably add evaluation after modeling. That way we will be following the structure of the repo

publications/recommender.tex
anargyri: All the papers have an abstract. Also note that according to http://jmlr.org/mloss/mloss-info.html it has to be 4 pages long at most.

publications/recommender.tex
loomlike: This is the question I also have. We need to fix the name to call the repository throughout the paper.

publications/recommender.tex
loomlike: thought the publisher will assign editors when they publish. If not, yes we don't need.

publications/recommender.tex
loomlike: I kinda like to have this since:
1. (Number of contributors shows) the repo is not Microsoft's internal, but more like open-sourced collaboration attempt.
2. (The increase of stars shows) there exist many needs of best practices for building recommender systems. Especially, this point supports our main contribution -- people need recommenders best practice, so we did.

What do you guys think?

publications/recommender.tex
loomlike: Agreed. If we cannot meet the paper-length criteria, then we can merge some sections. For now, let's have separate sections.

publications/recommender.tex
miguelgfierro: I also like Microsoft Recommenders

miguelgfierro:
I was looking at TF, pytorch and sklearn papers and they talk about the software design. Maybe we should have a section talking abou that as well 


https://www.usenix.org/system/files/conference/osdi16/osdi16-abadi.pdf
http://www.jmlr.org/papers/volume12/pedregosa11a/pedregosa11a.pdf
https://openreview.net/pdf?id=BJJsrmfCZ

anargyri:
We could tentatively follow the skeleton of the sklearn MLOSS paper. E.g.
- Introduction
- Vision / Principles / Target Audience
- Technologies / Platforms
- Code design / guidelines / best practices
- Listing of the notebooks and algos
- Conclusion

The length should be 4 pages (without references) http://jmlr.org/mloss/mloss-info.html (bear in mind that the names of the authors will take some space). Hence the listing of algos would probably have to be very brief, with references to the literature, but not much else.

yueguoguo:
I merge this PR and take @anargyri suggestion to outline the paper.

PR #791, opened 2019-05-21
refs

miguelgfierro:
Adding some references. I'm not sure about how I cited VW, @eisber, is it correct?. Also, I have contacted the original author of SAR to check if the reference is accurate

### DIFF
--- a/publications/references.bib
+++ b/publications/references.bib
@@ -102,11 +102,16 @@
   year={2008}
 }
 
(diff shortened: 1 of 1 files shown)


eisber:
It seems folks are citing: A. Agarwal et al. A Reliable Effective Terascale Linear Learning System. JMLR, 15(1), 2014.

(also Wikipedia VW article links to this paper)

Issue #775, opened 2019-05-07
[BUG]tests are failing. 


wutaomsft:

gramhagen:
several issues going on here
- master cpu is failing due to the nni smac issue @anargyri are you ok pushing the current staging solution to this to master? or are there changes we should make to the nni notebook?

- master gpu is failing due to metrics discrepancy with xdeepfm - could be something we can fix with setting seeds, @loomlike is this something you have time to look into?

- staging failures seem to be a temporary hiccup connecting to git. i queued them to rerun, so hopefully those will be fixed shortly

anargyri:
I agree we should push staging until NNI fix the issue.

miguelgfierro:
SAR issue is fixed. We still need to fix the issue with NNI https://github.com/microsoft/recommenders/issues/746

@loomlike I think you have a fix for the xdeepfm right? is this already in staging?

gramhagen:
addressing the staging nni failure in #781

loomlike:
Will take a look at xDeepFM one.

loomlike:
Is xDeepFM test still failing? If so, please give me more context, i.e. would be nice to know specific test function names that fail. I don't see them from my machine or PRs.

miguelgfierro:
I haven't seen it failing for a while, but not sure

PR #792, opened 2019-05-21
Bug/nni

miguelgfierro:
### Description

Fixes https://github.com/microsoft/recommenders/issues/775

### Checklist:

- [x] I have added tests.

### DIFF
--- a/reco_utils/nni/nni_utils.py
+++ b/reco_utils/nni/nni_utils.py
@@ -36,7 +36,7 @@ def check_experiment_status(wait=WAITING_TIME, max_retries=MAX_RETRIES):
         if nni_status["status"] in ["DONE", "TUNER_NO_MORE_TRIAL"]:
             break
         elif nni_status["status"] not in ["RUNNING", "NO_MORE_TRIAL"]:
-            raise RuntimeError("NNI experiment failed to complete with status {} - {}".format(nni_status["status"], 
+            raise RuntimeError("NNI experiment failed to complete with status {}: {}".format(nni_status["status"], 
                                                                                               nni_status["errors"]))
         time.sleep(wait)
         i += 1
--- a/reco_utils/nni/nni_utils.py
+++ b/reco_utils/nni/nni_utils.py
@@ -114,7 +114,7 @@ def get_trials(optimize_mode):
 
 def stop_nni():
     """Stop nni experiment"""
-    proc = subprocess.run([sys.prefix + '/bin/nnictl', 'stop'])
+    proc = subprocess.run([sys.prefix + "/bin/nnictl", "stop"])
     if proc.returncode != 0:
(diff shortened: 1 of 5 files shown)


gramhagen:
looks like there are couple of test changes needed though to match the mocked return values to the expected results.

tests/integration/test_notebooks_python.py
```diff
@@ -149,19 +149,14 @@ def test_vw_deep_dive_integration(notebooks, size, expected_values):
 
 
 @pytest.mark.integration
-@pytest.mark.skipif(sys.platform == 'win32', reason="nni not installable on windows")
+@pytest.mark.skipif(sys.platform == "win32", reason="nni not installable on windows")
 def test_nni_tuning_svd(notebooks, tmp):
     notebook_path = notebooks["nni_tuning_svd"]
-    # First check if NNI is running
```
anargyri: Shouldn't we keep this? (in case a previous test has failed)

reco_utils/nni/nni_utils.py
```diff
@@ -1,31 +1,32 @@
...
+NNI_TRIAL_JOBS_URL = NNI_REST_ENDPOINT + "/trial-jobs"
 WAITING_TIME = 20
-MAX_RETRIES = 5
+MAX_RETRIES = 10
 
 
-def get_experiment_status(status_url):
-    """
-    Helper method. Gets the experiment status from the REST endpoint
+def get_experiment_status(status_url=NNI_STATUS_URL):
```
anargyri: Should we also use this function to propagate any errors in the trials? https://github.com/microsoft/nni/issues/1082#event-2345623807
Maybe add a new issue about this?

tests/integration/test_notebooks_python.py
miguelgfierro: I got an error, that's why I removed it. Also, in the notebook, we start by stoping nni. Do you still think we should keep it?

reco_utils/nni/nni_utils.py
miguelgfierro: you mean `nnictl trial ls`? can you elaborate on what is in your mind about this?

tests/integration/test_notebooks_python.py
anargyri: Ok, no problem then.

reco_utils/nni/nni_utils.py
anargyri: This, or by calling the REST API on NNI_TRIAL_JOBS_URL like in `get_trials()`. The trial status should be included in the response for all trials.

miguelgfierro: added issue https://github.com/microsoft/recommenders/issues/794

PR #793, opened 2019-05-21
Le libffm converter fix

yueguoguo:
### Description

* Fix the encoding issue of the libffm converter - the previous implementation does not encode the feature for numerical variable correctly. Its index should be incremental instead of being the same as the field index.
* Update unit test for the converter implementation
* Update docstring

### Checklist:

- [x] I have added tests.


reco_utils/dataset/pandas_df_utils.py
```diff
@@ -189,13 +189,13 @@ def transform(self, df):
...
-        field_feature_dict = {}
+        self.field_feature_dict = {}
         for field in self.field_names:
             for feature in df[field].values:
                 # Check whether (field, feature) tuple exists in the dict or not.
                 # If not, put them into the key-values of the dict and count the index.
-                if (field, feature) not in field_feature_dict:
-                    field_feature_dict[(field, feature)] = idx
+                if (field, feature) not in self.field_feature_dict:
+                    self.field_feature_dict[(field, feature)] = idx
```
gramhagen: it seems strange to modify the class during transform, this is not what I would expect based on uses of transform from spark's api anyway.

would it be possible to provide a default value or throw an error if an unexpected value is encountered?

yueguoguo: Good catch. The indexing should be moved to fit, and the class attribute of field-feature-dict should be kept unchanged in `transform`.

PR #813, opened 2019-06-05
Change url download util to use requests

AbhiramE:
### Description

The change changes the url download utility from urlretrieve library to requests. With this change I also get rid of the Tqdm utility class and move to using Tqdm inline.

Requests is a more popular library than urlretrieve.

### Checklist:

- [ x ] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project. 
There are tests already covering this use case in the repo. I believe this also requires no update to the documentation.

### DIFF
--- a/reco_utils/dataset/download_utils.py
+++ b/reco_utils/dataset/download_utils.py
@@ -31,7 +31,7 @@ def maybe_download(url, filename=None, work_directory=".", expected_bytes=None):
         r = requests.get(url, stream=True)
         total_size = int(r.headers.get("content-length", 0))
         block_size = 1024
-        num_iterables = math.ceil(total_size // block_size)
+        num_iterables = math.ceil(total_size / block_size)
 
         with open(filepath, "wb") as file:
             for data in tqdm(
--- a/reco_utils/dataset/download_utils.py
+++ b/reco_utils/dataset/download_utils.py
@@ -2,41 +2,24 @@
 # Licensed under the MIT License.
 
 import os
-from urllib.request import urlretrieve
 import logging
+import requests
+import math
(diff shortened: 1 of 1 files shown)


miguelgfierro:
Some evidences here: https://stackoverflow.com/questions/2018026/what-are-the-differences-between-the-urllib-urllib2-and-requests-module

This code had been added to the NLP repo

reco_utils/dataset/download_utils.py
```diff
@@ -2,50 +2,45 @@
...
         filename = url.split("/")[-1]
     filepath = os.path.join(work_directory, filename)
     if not os.path.exists(filepath):
-        with TqdmUpTo(unit="B", unit_scale=True) as t:
-            filepath, _ = urlretrieve(url, filepath, reporthook=t.update_to)
+
+        r = requests.get(url, stream=True)
+        total_size = int(r.headers.get("content-length", 0))
+        block_size = 1024
+        num_iterables = math.ceil(total_size // block_size)
```
gramhagen: if you want to round up you will need to avoid integer division
```
math.ceil(total_size / block_size)
```

AbhiramE:
Do you mean doing this without the math package? Like this,

(total_size+block_size-1) / block_size

gramhagen:
> Do you mean doing this without the math package? Like this, (total_size+block_size-1) / block_size

no, using math is fine, but the // operator does integer division which will round down (10 // 3 = 3), so the ceil is not actually doing anything. but if you do (10 / 3 = 3.333) then apply ceil you will get 4 which is what I'm assuming you want.

gramhagen:
awesome, thanks!

Issue #689, opened 2019-03-28
[FEATURE] add a new folder to put all the code related to tuning: reco_utils/tuning


miguelgfierro:
### Description

from a conversation with @anargyri:

>  It would be more appropriate to have a folder called tuning and, under that folder, azureml and nni and the spark tuning code. This would require testing again all the tuning notebooks, so I would leave it for a separate PR. Hyperdrive and NNI rely on several path names, so they will break as soon as you change the folder structure.

### Related
https://github.com/Microsoft/Recommenders/pull/687

loomlike:
azureml may include o16n or remote-compute training related helper scripts in the future, not necessarily limited to tuning. What else do we expect to have other than NNI under the tuning folder?

yueguoguo:
My two cents, if in the `tuning` directory, there are just training scripts instead of "utility" function scripts, would it be better we place them together with the notebooks? Most likely the training scripts are not used by other scripts/notebooks.

miguelgfierro:
interesting idea, @anargyri @loomlike are there any utility functions in the hyperparameter tuning code?

anargyri:
> interesting idea, @anargyri @loomlike are there any utility functions in the hyperparameter tuning code?

I think putting the .py training scripts in the same directory as the notebooks does not look nice and may confuse the reader. I would prefer it as it is now, or maybe have a directory for SVD_hyperdrive, one for wide_and_deep_hyperdrive etc.

There are not any common tuning utils currently.

loomlike:
@anargyri what about put under a child folder (e.g. `scripts`) of the notebooks then? like
`../notebooks/04_model_select_and_optimize/scripts/` ?

anargyri:
TBH most of these suggestions sound good to me. Just in the interest of housekeeping, having .py and .ipynb files together is a bit messy and you need to take care that it is clear which .py corresponds to each .ipynb.

anargyri:
We should probably revisit this issue when / if the hyperparameter tuning repo is open-sourced, since there are some utilities + Kubeflow notebooks there.

PR #821, opened 2019-06-07
move reco_utils/nni to reco_utils/tuning/nni

miguelgfierro:
### Description

move reco_utils/nni to reco_utils/tuning/nni to have only general concepts in reco_utils

### Related Issues

https://github.com/microsoft/recommenders/issues/689

### Checklist:


anargyri:
Since we do not have unit tests for the tuning notebooks, could you perform the integration tests before merging this PR?

anargyri:
These are not the only changes needed, because svd_training.py has now also moved. The locations of this script inside the NNI notebook should also be modified.

miguelgfierro:
mmm I know what you mean, let me take a deeper look

anargyri:
Another note is that we have the tuning scripts for AzureML under the azureml folder. Should we move those under the tuning folder too or keep them where they are?

miguelgfierro:
> Another note is that we have the tuning scripts for AzureML under the azureml folder. Should we move those under the tuning folder too or keep them where they are?

We can do it. Maybe we should do it in a different PR

gramhagen:
there's also the evaluation/parameter_sweep module, which seems like it would fit better under tuning

reco_utils/tuning/nni/svd_training.py
```diff
@@ -10,7 +10,7 @@
 import nni
 import surprise
 import sys
-sys.path.append("../../")
+sys.path.append("../../../")
```
gramhagen: is this needed?

yueguoguo:
LGTM

reco_utils/tuning/nni/svd_training.py
miguelgfierro: yeah it is needed, it won't work otherwise

PR #826, opened 2019-06-10
adding marker to skip vw tests if necessary

gramhagen:
### Description
new devops setup has the tests run on machines without vw installed. until that can be sorted out we are adding an optional marker which will allow all vw related tests to be skipped by using -k "not vw" in the pytest command.

### Checklist:

- [X] I have added tests covering my contributions.


yueguoguo:
Same ask as @bethz maybe `-m` with marker would be good?

bethz:
@yueguoguo The way it's added to the test files will work as-is with -m.

Issue #842, opened 2019-06-15
azureml_hyperdrive_surprise_svd notebook error


atimesastudios:
### Description
Notebook throws an error when submitting the experiment.

### In which platform does it happen?
System version: 3.7.3 | packaged by conda-forge | (default, Mar 27 2019, 23:18:50) [MSC v.1900 64 bit (AMD64)]
Surprise version: 1.0.6
Azure ML SDK Version: 1.0.39

### How do we replicate the issue?
pull from origin/master
run the python notebook

Problem occurs here:
# Create an experiment to track the runs in the workspace
exp = aml.core.Experiment(workspace=ws, name=EXP_NAME)
run = exp.submit(config=hd_config)

azureml.widgets.RunDetails(run).show()
run.wait_for_completion(show_output=True)

Output is:
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-14-5a56dcc735be> in <module>
      1 # Create an experiment to track the runs in the workspace
      2 exp = aml.core.Experiment(workspace=ws, name=EXP_NAME)
----> 3 run = exp.submit(config=hd_config)
      4 
      5 azureml.widgets.RunDetails(run).show()

~\AppData\Local\Continuum\anaconda3\envs\myenv\lib\site-packages\azureml\core\experiment.py in submit(self, config, tags, **kwargs)
    142         submit_func = get_experiment_submit(config)
    143         with self._log_context("submit config {}".format(config.__class__.__name__)):
--> 144             run = submit_func(config, self.workspace, self.name, **kwargs)
    145         if tags is not None:
    146             run.set_tags(tags)

~\AppData\Local\Continuum\anaconda3\envs\myenv\lib\site-packages\azureml\train\hyperdrive\_search.py in search(hyperdrive_run_config, workspace, experiment_name)
    100 
    101         experiment_dto = _create_experiment_dto(hyperdrive_run_config, workspace, experiment_name,
--> 102                                                 telemetry_values, activity_logger)
    103 
    104         try:

~\AppData\Local\Continuum\anaconda3\envs\myenv\lib\site-packages\azureml\train\hyperdrive\_search.py in _create_experiment_dto(hyperdrive_run_config, workspace, experiment_name, telemetry_values, activity_logger)
     30         activity_logger.info("Creating snapshot...")
     31 
---> 32     platform_config = hyperdrive_run_config._get_platform_config(workspace, experiment_name)
[...]

PR #847, opened 2019-06-20
Wide-deep hyperdrive notebook Azureml API update

loomlike:
### Description
AzureML API update fix

### Related Issues
#842 

### Checklist:

### DIFF
--- a/notebooks/04_model_select_and_optimize/azureml_hyperdrive_wide_and_deep.ipynb
+++ b/notebooks/04_model_select_and_optimize/azureml_hyperdrive_wide_and_deep.ipynb
@@ -7,7 +7,7 @@
     "<i>Copyright (c) Microsoft Corporation. All rights reserved.<br>\n",
     "Licensed under the MIT License.</i>\n",
     "<br>\n",
-    "# Recommender Hyperparameter Tuning w/ AzureML"
+    "# Wide-and-Deep Model Hyperparameter Tuning with AzureML"
    ]
   },
   {
@@ -24,10 +24,10 @@
     "5. Setup and Run Hyperdrive Experiment\n",
     "6. Model Import, Re-train and Test\n",
     "\n",
-    "In this notebook, we use [**Wide-and-Deep model**](https://ai.googleblog.com/2016/06/wide-deep-learning-better-together-with.html) from **TensorFlow high-level Estimator API (v1.12)** on the movie recommendation scenario. Wide-and-Deep learning jointly trains wide linear model and deep neural networks (DNN) to combine the benefits of memorization and generalization for recommender systems.\n",
+    "In this notebook, we use [**Wide-and-Deep model**](https://ai.googleblog.com/2016/06/wide-deep-learning-better-together-with.html) from **TensorFlow high-level Estimator API (v1.12 or higher)** on the movie recommendation scenario. Wide-and-Deep learning jointly trains wide linear model and deep neural networks (DNN) to combine the benefits of memorization and generalization for recommender systems.\n",
     "\n",
     "For more details about the **Wide-and-Deep** model:\n",
-    "* [Wide-Deep Quickstart notebook](../00_quick_start/wide_deep_movielens.ipynb)\n",
+    "* [Wide-and-Deep Quickstart notebook](../00_quick_start/wide_deep_movielens.ipynb)\n",
     "* [Original paper](https://arxiv.org/abs/1606.07792)\n",
     "* [TensorFlow API doc](https://www.tensorflow.org/api_docs/python/tf/estimator/DNNLinearCombinedRegressor)\n",
     "  \n",
@@ -43,14 +43,14 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Azure ML SDK Version: 1.0.10\n",
+      "Azure ML SDK Version: 1.0.18\n",
       "Tensorflow Version: 1.12.0\n"
      ]
     }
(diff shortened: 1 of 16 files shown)


yueguoguo:
@loomlike seems it's got a stackoverlfow error in Spark workload.

notebooks/04_model_select_and_optimize/azureml_hyperdrive_wide_and_deep.ipynb
```diff
@@ -7,7 +7,7 @@
     "<i>Copyright (c) Microsoft Corporation. All rights reserved.<br>\n",
```
gramhagen: <p>any thoughts on why these values dropped so much?</p><p><br></p><p>also, i think using getattr is a bit obscure, it would be clearer to just call the metrics code directly</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/847/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

gramhagen: <p>also what changed here to impact the metrics? are we not setting a random seed when training the model?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/847/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

loomlike: The w&d notebook does use the default seed unless we explicitly set `None`.
Will test again and see if it reports the same number again (should be).

tests/integration/test_notebooks_gpu.py
```diff
@@ -175,31 +175,31 @@ def test_xdeepfm_integration(notebooks, syn_epochs, criteo_epochs, expected_valu
 @pytest.mark.integration
 @pytest.mark.gpu
 @pytest.mark.parametrize(
-    "size, epochs, expected_values, seed",
+    "size, steps, expected_values, seed",
     [
         (
             "1m",
-            10,
+            50000,
```
gramhagen: how does this relate to epochs? will this be faster than what it was previously?

loomlike: @gramhagen Thanks for the review! Basically, `steps = num_train_samples * epochs / batch_size` and I'm using steps here just because the modern TF api accepts it. Benefit of using steps over epochs is you have finer control over how many batches to use for training. Useful when your training set is very big.

Previously, I converted epochs into steps manually from the notebook and passed it to model functions. Now I'm just accepting the steps directly.

loomlike: Btw, I'm having two tests failed. Are they known issue of the repo?

miguelgfierro: mmm this is a weird error:
```
tests\unit\test_tf_utils.py ...E.E                                       [ 94%]
tests\unit\test_wide_deep_utils.py ..                                    [100%]

=================================== ERRORS ====================================
________________ ERROR at teardown of test_evaluation_log_hook ________________

tp = <class 'OSError'>, value = None, tb = None

    def reraise(tp, value, tb=None):
        try:
            if value is None:
                value = tp()
            if value.__traceback__ is not tb:
                raise value.with_traceback(tb)
>           raise value

C:\Anaconda\envs\reco_gpu\lib\site-packages\six.py:693: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\Anaconda\envs\reco_gpu\lib\site-packages\six.py:693: in reraise
    raise value
C:\Anaconda\envs\reco_gpu\lib\site-packages\six.py:693: in reraise
    raise value
tests\conftest.py:25: in tmp
    yield td
C:\Anaconda\envs\reco_gpu\lib\tempfile.py:809: in __exit__
    self.cleanup()
C:\Anaconda\envs\reco_gpu\lib\tempfile.py:813: in cleanup
    _shutil.rmtree(self.name)
C:\Anaconda\envs\reco_gpu\lib\shutil.py:500: in rmtree
    return _rmtree_unsafe(path, onerror)
C:\Anaconda\envs\reco_gpu\lib\shutil.py:399: in _rmtree_unsafe
    onerror(os.rmdir, path, sys.exc_info())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

path = 'C:\\Users\\recocat\\AppData\\Local\\Temp\\pytest-of-recocat\\pytest-9\\tmpccwa3tpo'
onerror = <function rmtree.<locals>.onerror at 0x00000221524BA2F0>

    def _rmtree_unsafe(path, onerror):
        try:
            if os.path.islink(path):
                # symlinks to directories are forbidden, see bug #1669
                raise OSError("Cannot call rmtree on a symbolic link")
        except OSError:
            onerror(os.path.islink, path, sys.exc_info())
            # can't continue even if onerror hook returns
            return
        names = []
        try:
            names = os.listdir(path)
        except OSError:
            onerror(os.listdir, path, sys.exc_info())
        for name in names:
            fullname = os.path.join(path, name)
            try:
                mode = os.lstat(fullname).st_mode
            except OSError:
                mode = 0
            if stat.S_ISDIR(mode):
                _rmtree_unsafe(fullname, onerror)
            else:
                try:
                    os.unlink(fullname)
                except OSError:
                    onerror(os.unlink, fullname, sys.exc_info())
        try:
>           os.rmdir(path)
E           OSError: [WinError 145] The directory is not empty: 'C:\\Users\\recocat\\AppData\\Local\\Temp\\pytest-of-recocat\\pytest-9\\tmpccwa3tpo'

```
looks like an error with shutil?

loomlike: Somehow only the Windows DSVM complains about TemporaryDirectory cleanup with pytest yield-fixture... while that works well on my local Windows machines (both laptop and workstation) and Linux DSVM.
Anyway I made a change for that and has been fixed now.

tests/unit/test_wide_deep_utils.py
```diff
@@ -40,60 +44,73 @@ def pd_df():
...
+    # Test if wide column has two original features and one crossed feature
+    wide_columns, _ = build_feature_columns(
+        users, items, model_type='wide', crossed_feat_dim=10
+    )
+    assert len(wide_columns) == 3
+    # Check crossed feature dimension
+    assert wide_columns[2].hash_bucket_size == 10
+    # Check model type
+    model = build_model(
+        str(tmp_path / ('wide_' + MODEL_DIR)), wide_columns=wide_columns
```
gramhagen: I don't understand what this does?
```str(tmp_path / ('wide_' + MODEL_DIR))```

why is there a division operator here? should it be os.path.join(str(tmp_path), 'wide_' + MODEL_DIR), would this be causing the problem with windows path cleanup?

loomlike: @gramhagen when I use our fixture (tmp) for temporary directory, it causes error (only) on Windows DSVM. So I end up using pytest's fixture tmp_path witch returns `Path` object. And the codes use Path object's syntax `/`.

loomlike: Btw, the error is -- somehow Windows' DSVM holds TF model checkpoints folder and thus TemporaryDirectory cannot cleanup the folder by the time the test has done.

pytest's fixture tmp_path keeps the folder for last three test sessions instead of removing it right away. That's how it avoids the cleanup error. Not a clean solution but at least it is working for now for Windows testing.

gramhagen: Ok, I see, I wasn't familiar with that syntax. Do you think there is something that needs to be done to close the tf session or something in order to release the tmp path for cleanup?

gramhagen: Maybe the best thing to do is keep it as is and file an issue for using the normal yield fixture with TF

loomlike: I tried `del model` and no luck. Kind of hard to debug because that is happening only from Win-DSVM. @roalexan Any idea about this? FYI no issues when use local Windows machines (both laptop and workstation). Does Win-DSVM handle processes and folder read/write permission differently from local Windows'?

roalexan: Should just be Windows Server 2016 - don't see how it would be different than local. [Link](https://docs.microsoft.com/en-us/azure/machine-learning/data-science-virtual-machine/provision-vm) to product page. Maybe we could use one of these email aliases to ask a question about this: dsvm-product-team@ or Dsvm-interest@.

loomlike: Somewhat related issue found from Tensorflow github issues: https://github.com/tensorflow/tensorflow/issues/20606

I will try to force close tf's FileWriter and see if that solves this issue on Windows DSVM

loomlike: Ok it does solve. I'm working on adding those changes.

gramhagen: great, feel free to merge if everything passes

PR #863, opened 2019-07-12
Python evaluator module fix

loomlike:
### Description
Python evaluation module' ranking metric functions have redundant and unnecessary sorting codes.
E.g. 
```
df_hit["rank"] = df_hit.groupby(col_user)[col_prediction].rank(
        method="first", ascending=False
)
```
doesn't need to use `rank()` since `df_hit` is already sorted by user and ratings as it is generated by groupby user (pandas groupby's `sort` argument is by default True) and nlargest ratings.

This change removes those redundant and unnecessary sorts and also refactor `get_top_k_items` to return DataFrame with 'rank' column to make its behavior the same as our pyspark evaluation module.

### Checklist:

### DIFF
--- a/reco_utils/evaluation/python_evaluation.py
+++ b/reco_utils/evaluation/python_evaluation.py
@@ -363,7 +363,7 @@ def merge_ranking_true_pred(
 
     Returns:
         pd.DataFrame, pd.DataFrame, int:
-            DataFrame of recommendation hits
+            DataFrame of recommendation hits, sorted by `col_user` and `"rank"`
             DataFrmae of hit counts vs actual relevant items per user
             number of unique user ids
     """
@@ -389,9 +389,6 @@ def merge_ranking_true_pred(
         col_user=col_user,
         col_rating=col_prediction,
         k=top_k,
-    )
-    df_hit["rank"] = df_hit.groupby(col_user)[col_prediction].rank(
-        method="first", ascending=False
     )
     df_hit = pd.merge(df_hit, rating_true_common, on=[col_user, col_item])[
         [col_user, col_item, "rank"]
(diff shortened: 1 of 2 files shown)


reco_utils/evaluation/python_evaluation.py
```diff
@@ -651,14 +648,16 @@ def get_top_k_items(
...
+        pd.DataFrame: DataFrame of top k items for each user, sorted by `col_user` and `"rank"`
     """
 
-    return (
+    top_k_items = (
         dataframe.groupby(col_user, as_index=False)
         .apply(lambda x: x.nlargest(k, col_rating))
         .reset_index(drop=True)
     )
+    top_k_items["rank"] = top_k_items.groupby(col_user).cumcount() + 1
```
gramhagen: you can avoid the repeated groupby too

```
groups = dataframe.groupby(col_user, as_index=False)
top_k_items = groups.apply(lambda x: x.nlargest(k, col_rating)).reset_index(drop=True)
top_k_items["rank"] = groups.cumcount() + 1
```

reco_utils/evaluation/python_evaluation.py
```diff
@@ -651,14 +648,16 @@ def get_top_k_items(
         k (int): number of items for each user
 
     Returns:
-        pd.DataFrame: DataFrame of top k items for each user
+        pd.DataFrame: DataFrame of top k items for each user, sorted by `col_user` and `"rank"`
```
gramhagen: i would remove the double quotes from rank to match just the backticks like col_user

gramhagen: also, in the returns section =)

loomlike: good catch!

loomlike:
@gramhagen Few changes since the last review:
1. changed `"rank"` to `rank`
2. caching `groups` turns out does not work, since `nlargest` sorts the ratings while the cached group object still contains unsorted ratings. I changed it back to use groupby again, but added `sort=False` so that groupby can be performed efficiently (groupby-without-sorting still keeps the inter-group orders and we already sorted previously by 'nlargest')
3. found the above issue from spark's unit-tests which matches spark-evaluation-fn results to python's. Python evaluation tests couldn't catch the error because the test case users and items were already sorted. I made a simple tweak to the test case so that can catch such errors in the future.

gramhagen:
oh interesting, i didn't realize we use the python evaluation to validate test results for spark, we should remove that linkage, I'll add a separate feature request

gramhagen:
oh, i take it back, I guess that's an additional check just to ensure they match. i guess it helped in this case.

miguelgfierro:
@loomlike feel free to merge when you think it is convenient

PR #867, opened 2019-07-18
Staging to master after documentation and other changes

miguelgfierro:
### Checklist:


yueguoguo:
@miguelgfierro you have indeed updated the **documentation** accordingly 😄 
image

reco_utils/dataset/spark_splitters.py
```diff
@@ -3,8 +3,18 @@
...
+    from pyspark.sql.functions import (
+        col,
+        row_number,
+        broadcast,
+        rand,
+        collect_list,
+        size,
+    )
+except ImportError:
+    pass  # skip this import if we are in pure python environment
```
gramhagen: when is this imported in a non-spark environment? i would think it would be better to raise the error at that time?

reco_utils/evaluation/spark_evaluation.py
```diff
@@ -1,10 +1,14 @@
...
-from pyspark.sql.functions import col, row_number, expr
-import pyspark.sql.functions as F
+
+try:
+    from pyspark.mllib.evaluation import RegressionMetrics, RankingMetrics
+    from pyspark.sql import Window, DataFrame
+    from pyspark.sql.functions import col, row_number, expr
+    import pyspark.sql.functions as F
+except ImportError:
+    pass  # skip this import if we are in pure python environment
```
gramhagen: same comment as above, I can't recall why this change was made?

reco_utils/dataset/spark_splitters.py
miguelgfierro: if we don't add this clause, we can't run the cpu or gpu tests, because it will raise an error when trying to import spark

PR #874, opened 2019-07-23
DATA: update sum check

yueguoguo:
### Description
`sum` of floating numbers is approximate and there is loss of precision. This causes issues in the "ratio sum to 1" check in the Python random splitter. 

Using [`math.fsum`](https://docs.python.org/3/library/math.html#math.fsum) to resolve the issue.

### Related Issues
#866 

### Checklist:

- [x] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions.
- [ ] I have updated the documentation accordingly.

Issue #866, opened 2019-07-15
[BUG] Python random split sum to 1 error


gramhagen:
### Description
Python random split with multiple ratio options fails even when ratios sum to 1

### In which platform does it happen?
Python all OSs

### How do we replicate the issue?
```
from reco_utils.dataset.movielens import load_pandas_df
from reco_utils.dataset.python_splitters import python_random_split
df = load_pandas_df()
x, y, z = python_random_split(df, ratio=[.7, .2, .1])
```
yields
```
Traceback (most recent call last):
  File "C:\Users\scgraham\AppData\Local\Continuum\anaconda3\envs\reco_base\lib\site-packages\IPython\core\interactiveshell.py", line 3325, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-5-49ca4e7e6410>", line 1, in <module>
    x, y, z = python_random_split(df, ratio=[.7, .2, .1])
  File "C:\Users\scgraham\repos\Recommenders\reco_utils\dataset\python_splitters.py", line 38, in python_random_split
    splits = split_pandas_data_with_ratios(data, ratio, shuffle=True, seed=seed)
  File "C:\Users\scgraham\repos\Recommenders\reco_utils\dataset\split_utils.py", line 155, in split_pandas_data_with_ratios
    raise ValueError("The ratios have to sum to 1")
ValueError: The ratios have to sum to 1
```

this occurs because sum([.7, .2, .1]) == .999999.... and there is a check in split_utils.py:39 which fails and renormalizes, subsequently split_utils:154 has the same check which also fails which leads to this error state

### Expected behavior (i.e. solution)
data is split into train, test, and validate sets with expected fraction.

yueguoguo:
Good catch. Will fix it.

yueguoguo:
Resolved in #874 and validated in #876

PR #876, opened 2019-07-24
adding unit test for sum to 1 issue

gramhagen:
### Description
Test to address code change in #874 

### Related Issues
#866 

### Checklist:

- [ X I have added tests covering my contributions.

### DIFF
--- a/tests/unit/test_python_splitter.py
+++ b/tests/unit/test_python_splitter.py
@@ -196,6 +196,12 @@ def test_random_splitter(test_specs, python_dataset):
     for split in splits:
         assert set(split.columns) == set(python_dataset.columns)
 
+    # check values sum to 1
+    splits = python_random_split(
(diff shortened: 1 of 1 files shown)


tests/unit/test_python_splitter.py
```diff
@@ -196,6 +196,12 @@ def test_random_splitter(test_specs, python_dataset):
     for split in splits:
         assert set(split.columns) == set(python_dataset.columns)
 
+    # check values sum to 1
```
miguelgfierro: one question, you mentioned `# check values sum to 1` but then you are not checking this. Did you want to check that the first split is 70%, etc?

gramhagen: without the changes Le made this raises an error, so as long as it gets through this I'm happy. the other checks confirm the ratio is as expected.

Issue #907, opened 2019-08-29
[FEATURE] Improve performance of score() in SAR


zegerius:
### Description
Every time the SARSingleNode.score(**) is called for a list of users, it actually calculates the scores for all users. In our training-set of nearly .5M rows the dot product of the 'User Affinity'-matrix and 'Item Similarity'-matrix takes roughly 5+ seconds. The sizes of these matrices are respectively (23009, 17262) and (17262, 17262). This is too slow when calculating recommendations for one user.

### Expected behavior with the suggested feature
Optimize calculations by only calculating the scores for the specified users.

### Other Comments
This is already a TODO: https://github.com/microsoft/recommenders/blob/71a38d422c00329f8f8226ea24ad6260b1b7b4e9/reco_utils/recommender/sar/sar_singlenode.py#L304

yueguoguo:
This is indeed something useful to have. Thanks @zegerius

Add @maxkazmsft who is the original author for visibility and awareness.

zegerius:
Thanks @yueguoguo.

@Overv and I started working on a fix. Our first version still fails some tests for removing seen items, but the performance without removing seen items went to 33ms (99.4% improvement).

yueguoguo:
@zegerius great to know. What is the detailed failure for tests in removing seen items?

Overv:
@yueguoguo We've changed the code to select just the slice of the user affinities matrix that corresponds with the users that we actually want to predict. However, this new "incomplete" matrix makes it harder to use the existing logic/maths for removing seen items from it.

gramhagen:
that's great, yeah this is why it's remained a todo for so long =)

the approach that was in my head was to construct a sparse matrix (of -np.infs) for the seen items of the same size as user_affinity, then you could slice it the same way you are slicing user_affinity for relevant users and that to the score matrix.

PR #914, opened 2019-09-04
Improve SAR performance

zegerius:
Co-authored-by: @Overv

### Description

During production implementation of the SAR recommender algorithm, we were facing severe performance issues. This was the result of calculating recommendations for all users in the entire set, despite only requiring recommendations for one user.

The approach that was taken during the initial implementation allowed for easy removal of previously 'seen items'.

We changed the size of the affinity matrix that is used for the calculations, by only keeping the rows for the users in the test set, as follows:

https://github.com/microsoft/recommenders/blob/4c875df2d0b6b3ddad70a49a4353c5de19c28e61/reco_utils/recommender/sar/sar_singlenode.py#L307

To be able to do the same thing with the seen-items matrix, we needed to reshape it to reflect the affinity matrix, as follows:

https://github.com/microsoft/recommenders/blob/4c875df2d0b6b3ddad70a49a4353c5de19c28e61/reco_utils/recommender/sar/sar_singlenode.py#L252-L254

To follow the conventions set in the repo, we also clean up the `seen`-variable during model fitting.

https://github.com/microsoft/recommenders/blob/4c875df2d0b6b3ddad70a49a4353c5de19c28e61/reco_utils/recommender/sar/sar_singlenode.py#L260-L261

This then allows us to do the following, when the `remove_seen` arg is `True`:

https://github.com/microsoft/recommenders/blob/4c875df2d0b6b3ddad70a49a4353c5de19c28e61/reco_utils/recommender/sar/sar_singlenode.py#L314-L316

The resulting matrix then has all previously seen items set to `-np.inf`. The result gets cleaned up, as follows (not part of our PR):

https://github.com/microsoft/recommenders/blob/4c875df2d0b6b3ddad70a49a4353c5de19c28e61/reco_utils/recommender/sar/sar_singlenode.py#L457-L458

#### Performance
The performance improvement seen is significant, as summarized in the following table:

| remove_seen 	|     Original    	|     Optimized     	| Improvement 	|
|-----------:	|---------------:	|-----------------:	|--------------------:	|
|        **True** 	| 3.83 s ± 459 ms 	| 17.5 ms ± 1.06 ms 	|                 219x 	|
|       **False** 	| 3.42 s ± 468 ms 	|  17.9 ms ± 1.7 ms 	|                 191x 	|

_This analysis was done on our production dataset of approx. 500K rows._

We have not investigated the difference in training time.

### Related Issues

https://github.com/microsoft/recommenders/issues/907

### Checklist:

### DIFF
--- a/reco_utils/recommender/sar/sar_singlenode.py
+++ b/reco_utils/recommender/sar/sar_singlenode.py
@@ -244,18 +244,21 @@ class SARSingleNode:
                 temp_df = self.compute_time_decay(df=temp_df, decay_column=self.col_unity_rating)
             self.unity_user_affinity = self.compute_affinity_matrix(df=temp_df, rating_col=self.col_unity_rating)
 
-        # retain seen items for removal at prediction time
-        self.seen_items = temp_df[[self.col_user_id, self.col_item_id]].values
-
         # affinity matrix
         logger.info("Building user affinity sparse matrix")
         self.user_affinity = self.compute_affinity_matrix(df=temp_df, rating_col=self.col_rating)
+
+        # retain seen items for removal at prediction time
+        seen = temp_df[[self.col_user_id, self.col_item_id]].values
+        self.seen_items = np.ones(self.user_affinity.shape)
+        self.seen_items[seen[:, 0], seen[:, 1]] = -np.inf
 
         # calculate item co-occurrence
         logger.info("Calculating item co-occurrence")
         item_cooccurrence = self.compute_coocurrence_matrix(df=temp_df)
 
         # free up some space
+        del seen
         del temp_df
 
         self.item_frequencies = item_cooccurrence.diagonal()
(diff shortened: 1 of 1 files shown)


reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -244,18 +244,21 @@ def fit(self, df):
...
-
         # affinity matrix
         logger.info("Building user affinity sparse matrix")
         self.user_affinity = self.compute_affinity_matrix(df=temp_df, rating_col=self.col_rating)
 
+        # retain seen items for removal at prediction time
+        seen = temp_df[[self.col_user_id, self.col_item_id]].values
+        self.seen_items = np.ones(self.user_affinity.shape)
+        self.seen_items[seen[:, 0], seen[:, 1]] = -np.inf
+
```
gramhagen: what if we remove seen and self.seen_items altogether and leverage the user_affinity since that is already captured, see code change below

reco_utils/recommender/sar/sar_singlenode.py
```diff
@@ -301,20 +304,17 @@ def score(self, test, remove_seen=False, normalize=False):
...
+        test_scores = self.user_affinity[user_ids, :].dot(self.item_similarity)
 
         # ensure we're working with a dense ndarray
         if isinstance(test_scores, sparse.spmatrix):
             test_scores = test_scores.toarray()
 
+        # remove items in the train set so recommended items are always novel
+        if remove_seen:
+            logger.info("Removing seen items")
+            test_scores = test_scores * self.seen_items[user_ids, :]
```
gramhagen: here we can add the reduced user_affinity back in after multiplying with -inf, zeros will stay zero and anything we consider the user has seen will have a non-zero entry which will turn to -inf.
```
test_scores += self.user_affinity[user_ids, :] * -np.inf
```

zegerius:
Thank you @gramhagen for that very elegant improvement. I have incorporate your suggested changes. All tests pass. 

Also ran a quick performance comparison and I don't seen any difference compared to our suggested solution (at least not more than one std dev):
`remove_seen = True`: 18.2 ms ± 1.95 ms (from 17.5 ms ± 1.06 ms)
`remove_seen = False`: 20.4 ms ± 2.35 ms (from 17.9 ms ± 1.7 ms)

PR #930, opened 2019-09-15
Sar renaming

miguelgfierro:
### Description

According to Vesselin, this is the correct name for SAR: https://www.linkedin.com/feed/update/urn:li:activity:6578020680435056640?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A6578020680435056640%2C6578026247052374016%29&replyUrn=urn%3Ali%3Acomment%3A%28activity%3A6578020680435056640%2C6578059273253773312%29

### Checklist:


gramhagen:
There was some debate about this. The consensus was that everyone liked the original name so we changed it back here #761

yueguoguo:
To add more on top of @gramhage comments, the SAR was decided to be "Simple Algorithm for Recommendation", because we thought using the term of "Smart" does not make much sense as all kinds of algorithms are smart 😂 but the simplicity of the algorithm is in fact the key uniqueness that differentiates from the others. Maybe @wutaomsft can provide more details about the rationality.

Not sure but we may want to hear opinions from Vesselin who is the original author as well.

miguelgfierro:
ok got it

Issue #939, opened 2019-09-24
[BUG] Smoke test failure because of run_cell module in papermill


yueguoguo:
### Description
Running of smoke test failed because `run_cell` in `nbconvert` has been added with a new argument that breaks `papermill`. Discussion from other developers of this issue can be also found [here](https://github.com/nteract/papermill/issues/410). 

The problem can be fixed if upgrading `papermill` to version >1.0. However, this will lead to another issue where the method `record` is not available in `papermill` (it is deprecated). 
 
### In which platform does it happen?
All platforms where `papermill` 0.18 is used. 

### How do we replicate the issue?
Run the smoke testing by 
```
python -m pytest tests/smoke/test_notebooks_python.py -k test_sar_single_node_smoke
```

The following error will be generated

![image](https://user-images.githubusercontent.com/1726448/65485377-50b54200-ded4-11e9-91ba-5832b3a8d00b.png)

### Expected behavior (i.e. solution)
* The easiest way to resolve the issue is to fix the versions of `nbconvert` and `papermill` to the old one (`nbconvert` 5.5.0 and `papermill` 0.18.2) with which we passed the tests
* Alternative may be using [`scrapbook`](https://github.com/nteract/scrapbook) for recording and reading the data in executing the notebooks.

yueguoguo:
@miguelgfierro @gramhagen do we want to use `scrapbook` instead - it is used in NLP repo.

miguelgfierro:
+1 to using scrapbook

miguelgfierro:
related to https://github.com/microsoft/recommenders/issues/641

yueguoguo:
Fixed version of `nbconvert` and `papermill` to resolve the issue. 

Reopen the issue when `scrapbook` is being developed.

PR #940, opened 2019-09-24
SCRIPT: fix nbconvert version

yueguoguo:
### Description
Fix nbconvert version to resolve issue in #939

### Related Issues
#939 

### Checklist:

### DIFF
--- a/scripts/generate_conda_file.py
+++ b/scripts/generate_conda_file.py
@@ -81,6 +81,7 @@ PIP_BASE = {
     "memory-profiler": "memory-profiler>=0.54.0",
     "nvidia-ml-py3": "nvidia-ml-py3>=7.352.0",
     "papermill": "papermill==0.18.2",
+    "nbconvert": "nbconvert==5.5.0",
     "pydocumentdb": "pydocumentdb>=2.3.3",
     "pymanopt": "pymanopt==0.2.3",
     "tqdm": "tqdm==4.31.1",


miguelgfierro:
here you are fixing nvconvert, but this won't be needed if we upgrade papermill to the newest version right?

Second question, is nvconvert really needed? I don't see it in the nlp repo https://github.com/microsoft/nlp/blob/master/tools/generate_conda_file.py

yueguoguo:
@miguelgfierro `nvconvert` comes together with conda. Here we fix `nbcovert` to make it compatible with `papermill` (0.18.2) used in our conda environment. If we upgrade `papermill` to version > 1.0, we don't need to use `papermill` for recording the execution results - we need to use `scrapbook` instead, like nlp repo. Then we need to modify all of the notebooks where `papermill.record` is used.

So this is just a temporary solution to make the testing pass. In the long-term run we probably want to use `scrapbook`. Given that we are now in the process of migrating build pipelines, maybe we can postpone the replacement of `papermill.record` by using `scrapbook` in the next PR. 

What do you guys think? 

cc @bethz @gramhagen

miguelgfierro:
> So this is just a temporary solution to make the testing pass. In the long-term run we probably want to use scrapbook. Given that we are now in the process of migrating build pipelines, maybe we can postpone the replacement of papermill.record by using scrapbook in the next PR.

Makes sense

PR #909, opened 2019-09-01
FM deep dive notebook with xlearn examples

yueguoguo:
### Description
Add Factorization Machine / Field-Aware Factorization Machine deep dive notebook with `xlearn` example

### Related Issues
#908 

### Checklist:

- [x] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions.
- [x] I have updated the documentation accordingly.

PR #941, opened 2019-09-25
TEST: update aml test with custom docker image

yueguoguo:
### Description
@bethz Two updates in the script

1. I used a custom docker image for running the AzureML testing in the `submit_azureml_pytest.py` based on the old one. 
2. I used my service principal set up for the AzureML workspace for authentication. You can replace that with your old one as I noticed that you have it for the entire Azure subscription. 

The credentials used in the Azure Container Registry can be found in the email I sent you earlier. I exported them as environment variables in my own development machine. So you probably need to update them (line 413 to line 416) 

cc: @gramhagen @miguelgfierro 

### Related Issues
Testing failure of PR #909 because of lack of CMAKE in the AzureML base Docker image

### Checklist:

### DIFF
--- a/tests/ci/submit_azureml_pytest.py
+++ b/tests/ci/submit_azureml_pytest.py
@@ -22,7 +22,7 @@ Args:
     Optional but suggested, this info will be stored on Azure as
     text information as part of the experiment:
     --pr          (str): the Github PR number
-    --reponame    (str): the Github repository name
+    --reponame    (str): the Github registry name
     --branch      (str): the branch being run
                     It is also possible to put any text string in these.
 Example:
@@ -37,7 +37,7 @@ Example:
 import argparse
 import logging
 
-from azureml.core.authentication import AzureCliAuthentication
+from azureml.core.authentication import AzureCliAuthentication, ServicePrincipalAuthentication
 from azureml.core import Workspace
 from azureml.core import Experiment
 from azureml.core.runconfig import RunConfiguration
@@ -46,6 +46,8 @@ from azureml.core.script_run_config import ScriptRunConfig
 from azureml.core.compute import ComputeTarget, AmlCompute
 from azureml.core.compute_target import ComputeTargetException
 from azureml.core.workspace import WorkspaceException
+
+import os
 
 
 def setup_workspace(workspace_name, subscription_id, resource_group, cli_auth,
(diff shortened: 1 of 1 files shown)


tests/ci/submit_azureml_pytest.py
```diff
@@ -37,7 +37,7 @@
 import argparse
 import logging
 
-from azureml.core.authentication import AzureCliAuthentication
+from azureml.core.authentication import AzureCliAuthentication, ServicePrincipalAuthentication
```
bethz: SPA looks interesting.

tests/ci/submit_azureml_pytest.py
```diff
@@ -165,7 +181,15 @@ def create_run_config(cpu_cluster, docker_proc_type, conda_env_file):
     run_amlcompute = RunConfiguration()
     run_amlcompute.target = cpu_cluster
     run_amlcompute.environment.docker.enabled = True
-    run_amlcompute.environment.docker.base_image = docker_proc_type
+
+    # Allow use of a custom Docker image
+    if use_custom_docker:
+        run_amlcompute.environment.docker.base_image = custom_docker_image
+        run_amlcompute.environment.docker.base_image_registry.address = custom_docker_image_repo
+        run_amlcompute.environment.docker.base_image_registry.username = custom_docker_image_repo_username
```
bethz: not sure it's secure to pass username+pw as it may end up in log files?

tests/ci/submit_azureml_pytest.py
yueguoguo: NOTE this is for the AML workspace and I noticed in the current pipeline there is an SPA authentication at higher level - subscription. Is this something done by you and Richin?

tests/ci/submit_azureml_pytest.py
yueguoguo: Yea I am not sure about this, either. To make it secure probably we need to follow the approach with Azure Keyvault.

gramhagen:
is it necessary to have these separate functions inside submit_azureml_pytest? Are they re-used in any way elsewhere?

I would suggest simplifying this script by removing wrapper functions, this will also remove the need to pass secret variables extracted from the environment.

yueguoguo:
> is it necessary to have these separate functions inside submit_azureml_pytest? Are they re-used in any way elsewhere?

> I would suggest simplifying this script by removing wrapper functions, this will also remove the need to pass secret variables extracted from the environment.

Not sure eventually we want to reuse some of these functions for other BP repos (in the long-term run), but agree that they are currently not used anywhere else. Alternatively, these env variables can be made input arguments of the script, and then their values can be obtained from the group variables in the Azure DevOps pipeline.

gramhagen:
sure, i think there's room to keep simplifying this, just want to make sure we can work towards that direction

miguelgfierro:
do we want to merge this or are we waiting for more changes?

yueguoguo:
@bethz should I merge the codes into your branch so that you can test it for any necessary change?

PR #955, opened 2019-10-14
sarplus

eisber:
### Description
Spark/C++ based implementation of the SAR algorithm. Main goal is to scale further in terms of number of items/ratings.

(I'll add publishing of the pip and maven jar in a separate PR)

### Checklist:


contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+# Usage
+
+```python
+import pandas as pd
+from pysarplus import SARPlus
+
+# spark dataframe with user/item/rating/optional timestamp tuples
+train_df = spark.createDataFrame(
+      pd.DataFrame({
```
gramhagen: is pandas only used to generate the dataframe?
if so you could remove the import and create the dataframes by using a list of tuples
```
train_df = spark.createDataFrame([(1, 1, 1), (1, 2, 1), (2, 1, 1), (3, 1, 1), (3, 3, 1)], ['user_id', 'item_id', 'rating'])
```

contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
...
+os.environ["PYSPARK_SUBMIT_ARGS"] = SUBMIT_ARGS
+
+from pyspark.sql import SparkSession
+
+spark = (
+    SparkSession.builder.appName("sample")
+    .master("local[*]")
+    .config("memory", "1G")
+    .config("spark.sql.shuffle.partitions", "1")
+    .config("spark.sql.crossJoin.enabled", True)
```
gramhagen: i think we should remove crossJoin.enabled config in favor of calling df.crossJoin explicitly in the code

contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+Disable annoying logging
+
+```pyspark
+import logging
+logging.getLogger("py4j").setLevel(logging.ERROR)
+```
+
+
+# Packaging
```
gramhagen: might be better to put this into a separate document to avoid the perception that this is needed in order to use the package

contrib/sarplus/python/setup.py
```diff
@@ -0,0 +1,51 @@
...
+    version="0.2.6",
+    description="SAR prediction for use with PySpark",
+    url="https://github.com/Microsoft/Recommenders/contrib/sarplus",
+    author="Markus Cozowicz",
+    author_email="marcozo@microsoft.com",
+    license="MIT",
+    classifiers=[
+        "Development Status :: 4 - Beta",
+        "License :: OSI Approved :: MIT License",
+        "Programming Language :: Python :: 3.4",
```
gramhagen: we're testing 3.5-3.7?

contrib/sarplus/scala/python/setup.py
```diff
@@ -0,0 +1,10 @@
+from distutils.core import setup
```
gramhagen: not sure i understand why scala/python directory and this setup.py exist?

contrib/sarplus/python/pysarplus/SARPlus.py
```diff
@@ -0,0 +1,362 @@
...
+from pysarplus import SARModel
+
+SIM_COOCCUR = "cooccurrence"
+SIM_JACCARD = "jaccard"
+SIM_LIFT = "lift"
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger('sarplus')
+
+class SARPlus:
```
gramhagen: would be preferable to match the method signatures used in the python implementation:
reco_utils/recommender/sar/sar_singlenode.py, aside from unique variables for sarplus

for instance many of the fit variables were moved into the init method to make it more consistent with spark ml lib format

contrib/sarplus/README.md
gramhagen: hmm, maybe this is not possible when using sql queries to do the join?

AUTHORS.md
```diff
@@ -64,7 +64,7 @@ To contributors: please add your name to the list when you submit a patch to the
 * **[Jianxun Lian](https://github.com/Leavingseason)**
    * xDeepFM algorithm
    * DKN algorithm
-* **[Markus Cosowicz](https://github.com/eisber)**
+* **[Markus Cozowicz](https://github.com/eisber)**
```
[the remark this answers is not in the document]
eisber: let me revert

contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
+# sarplus (preview)
```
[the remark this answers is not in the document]
eisber: done

contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
...
+** make sure to always just keep top-k items in-memory
+** use standard join using binary search between users past seen items and the related items
+
+![Image of sarplus top-k recommendation optimization](images/sarplus_udf.svg) 
+
+# Usage
+
+```python
+import pandas as pd
+from pysarplus import SARPlus
```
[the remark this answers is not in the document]
eisber: I think due to conflict with the dummy python package for scala... I vaguely remember that I had to create the dummy package to trigger building...

contrib/sarplus/README.md
eisber: I wasn't aware one can do... we can rewrite that part using pyspark functions...

contrib/sarplus/README.md
eisber: +1

contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
+# sarplus (preview)
+pronounced sUrplus as it's simply better if not best!
+
+[![Build Status](https://dev.azure.com/best-practices/recommenders/_apis/build/status/contrib%20sarplus?branchName=master)](https://dev.azure.com/best-practices/recommenders/_build/latest?definitionId=107&branchName=master)
```
yueguoguo: Understanding that we are having a different pipeline for SAR+, do we want to consolidate the build status with other pipelines? 

Maybe also good to add it, and perhaps the other contrib in the future if there are more to come, to the [testing strategy](https://github.com/Microsoft/Recommenders/wiki/Test-Strategy) to make it clear.

contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
...
+5. Hit 'Create Library'
+6. Attach to your cluster
+7. Create 2nd library
+8. Under 'Source' select 'Upload Python Egg or PyPI'
+9. Enter 'pysarplus'
+10. Hit 'Create Library'
+
+This will install C++, Python and Scala code on your cluster.
+
+You'll also have to mount shared storage
```
yueguoguo: This may be optional - we can always use a local storage on ADB for caching the results, thought I am not sure if there is any downside.

contrib/sarplus/azure-pipelines.yml
```diff
@@ -0,0 +1,93 @@
...
+    include:
+      - contrib/sarplus/*
+
+# no CI trigger
+trigger: none
+
+jobs:
+- job: 'Test'
+  pool:
+    vmImage: 'Ubuntu 16.04'
```
yueguoguo: Do we want to reuse the agent pool for other builds?

contrib/sarplus/azure-pipelines.yml
```diff
@@ -0,0 +1,93 @@
...
+      alertWarningLevel: 'High'
+      sourceScanPath: contrib/sarplus
+
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '$(python.version)'
+      architecture: 'x64'
+
+  # pyarrow version: https://issues.apache.org/jira/projects/SPARK/issues/SPARK-29367
+  - script: python -m pip install --upgrade pip && pip install pyspark==$(spark.version) pytest pandas pybind11 pyarrow==0.14.1 sklearn
```
yueguoguo: Do we want to reuse the env setup with the repo, like what we do for other testing?

contrib/sarplus/python/pysarplus/SARPlus.py
```diff
@@ -0,0 +1,362 @@
...
+    """SAR implementation for PySpark"""
+
+    def __init__(
+        self,
+        spark,
+        col_user="userID",
+        col_item="itemID",
+        col_rating="rating",
+        col_timestamp="timestamp",
+        table_prefix="",
```
yueguoguo: `table_prefix` cannot pre-exist in the caching point, so do we want to add any detection scheme in the codes for that? So far, in my production system I have something manual for this. The problem comes from the use case when I want to have multiple SAR+ models created on the same cluster, hence the same caching point, at the same time, and I need to make sure that the prefixes are unique for the models.

contrib/sarplus/python/tests/sample-input.txt
```diff
@@ -0,0 +1,8 @@
+i1,i2,value
```
yueguoguo: Do we need this?

contrib/sarplus/python/tests/test_pyspark_sar.py
```diff
@@ -0,0 +1,390 @@
...
+from pyspark.sql import SparkSession
+
+from pysarplus import SARPlus, SARModel
+
+def assert_compare(expected_id, expected_score, actual_prediction):
+    assert expected_id == actual_prediction.id
+    assert math.isclose(expected_score, actual_prediction.score, rel_tol=1e-3, abs_tol=1e-3)
+
+@pytest.fixture(scope="module")
+def spark(app_name="Sample", url="local[*]", memory="1G"):
```
yueguoguo: Maybe reuse the initializer fixture function from the repo?

contrib/sarplus/python/tests/test_pyspark_sar.py
```diff
@@ -0,0 +1,390 @@
...
+    header = {
+        "col_user": "UserId",
+        "col_item": "MovieId",
+        "col_rating": "Rating",
+        "col_timestamp": "Timestamp",
+    }
+    return header
+
+@pytest.fixture(scope="module")
+def pandas_dummy_dataset(header):
```
yueguoguo: Maybe reuse some datasets we have for testing? https://github.com/microsoft/recommenders/blob/master/tests/sar_common.py

contrib/sarplus/README.md
```diff
@@ -0,0 +1,191 @@
...
+* convert similarity matrix to exactly the representation the C++ component needs, thus enabling simple shared, memory mapping of the cache file and avoid parsing. This requires a customer formatter, written in Scala
+* shared read-only memory mapping allows us to re-use the same memory from multiple python executors on the same worker node
+* partition the input test users and past seen items by users, allowing for scale out
+* perform as much of the work as possible in PySpark (way simpler)
+* top-k computation
+** reverse the join by summing reverse joining the users past seen items with any related items
+** make sure to always just keep top-k items in-memory
+** use standard join using binary search between users past seen items and the related items
+
+![Image of sarplus top-k recommendation optimization](images/sarplus_udf.svg)
```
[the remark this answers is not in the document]
eisber: not sure how to create vanity links, but uploaded the image

contrib/sarplus/README.md
eisber: I'd love to get rid of the setting as it's kind of annoying, though I'm not sure how to do it, as crossJoin() doesn't let me specify conditions...

contrib/sarplus/README.md
eisber: moved out

contrib/sarplus/python/setup.py
eisber: see azure-pipelines.yml (also changed from 3.4 to 3.5-3.7). I can reduce the number of python environments.

contrib/sarplus/scala/python/setup.py
eisber: this is not just a scala package, but a Spark package. That way I can automatically pull in the python dependency. Notice in the Jupyter notebook that there is no need to do a "pip install pysarplus" ?

contrib/sarplus/python/pysarplus/SARPlus.py
eisber: done

contrib/sarplus/README.md
eisber: not sure what you mean. do you want to consolidate and build this as part of the other pipelines? Since sarplus doesn't have dependencies to other parts of the repo, building it independently would save time and would allow us to release binaries from there.

contrib/sarplus/README.md
eisber: I don't think local storage will work as each node memory maps the distributed file..

contrib/sarplus/azure-pipelines.yml
eisber: aren't those the multi-tenant agents anyway?

contrib/sarplus/azure-pipelines.yml
eisber: are you referring to scripts/tests/ci/env-setup.yml?

contrib/sarplus/python/pysarplus/SARPlus.py
eisber: I'd suggest to use table_prefix=str(uuid.uuid4())? Does that work?

contrib/sarplus/python/tests/sample-input.txt
eisber: using the file as part of the unit tests. referring to the headers?

contrib/sarplus/python/tests/test_pyspark_sar.py
eisber: not sure it's worth the dependency?

contrib/sarplus/python/tests/test_pyspark_sar.py
eisber: not sure it's worth the dependency?

contrib/sarplus/README.md
yueguoguo: Ah no I mean do we want to put the build status on the same README and show them on the front page of the repo.

contrib/sarplus/README.md
yueguoguo: By local I mean `dbfs`

contrib/sarplus/azure-pipelines.yml
yueguoguo: Not sure but this will create new agent pool for VM to run the build or it will use the existing one? I am thinking of cost, but this may be cheaper because it does not have to use the GPU based VM agent we have in place.

contrib/sarplus/azure-pipelines.yml
yueguoguo: Something like that - `scripts/testes/ci/env-setup.yaml` is used for AML based testing. We put something into each of the yaml files on the DSVM based testing. Not sure reusing of them will be helpful...

contrib/sarplus/python/tests/test_pyspark_sar.py
yueguoguo: It's a part of reco_utils. So here we don't want to have much reliance on the original reco repo, right?

contrib/sarplus/python/tests/test_pyspark_sar.py
yueguoguo: Same as above.

contrib/sarplus/README.md
eisber: understood. good idea!

added a section to the end of the README under contribution builds

contrib/sarplus/README.md
eisber: if you refer to /dbfs/mnt/foo it has the big issue that it has issues with files larger than 2GB (I think it's backed by fuse

contrib/sarplus/python/tests/test_pyspark_sar.py
eisber: I think it makes the build processes easier (e.g. we'd have to check if other directories are changed as well), the yaml file only triggers on contrib/sarplus

contrib/sarplus/README.md
gramhagen: let's leave it as is

PR #971, opened 2019-11-05
TEST: revert to the old aml test pipeline yeaml

yueguoguo:
### Description
Remove the added arguments that take custom Docker images from private registy

### Checklist:

### DIFF
--- a/tests/ci/cpu_unit_tests.yml
+++ b/tests/ci/cpu_unit_tests.yml
@@ -54,7 +54,7 @@ steps:
   displayName: ' generate_conda_file.py'
 
 - script: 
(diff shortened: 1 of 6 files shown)


gramhagen:
Why are these being reverted?

yueguoguo:
@gramhagen ah I should say this is something temporary for now because we are using the DSVM-based testing pipeline. Later on we can get them back when revisiting the AzureML based testing pipeline. Does it make sense?

gramhagen:
I thought we were just disabling them?

yueguoguo:
Ah the reverted part is the unfinished ones that incorporate customized Docker images into the AzureML testing pipeline which are currently not working. I am thinking keeping the original AzureML testing pipeline that works, regardless where the resources are located, and later on when necessary, adding them back.

yueguoguo:
One thing I am not sure about is that even if I disabled the triggers of some pipelines in Azure Pipelines, they are still detected here and run when a new PR is submitted... 😢

gramhagen:
I see. Well if you need to remove them altogether we can always get them back from the history.

PR #972, opened 2019-11-07
O16n notebook update 

loomlike:
O16n (ALS) notebook update 

### Description
This changes includes:
* Revise databricks setup to zip only necessary codes (`./reco_utils`)
* Change cosmosdb writing codes for better performance (previously, took more than 10 minutes to write into the db. Now it takes 3-5 secs. It is a known [issue](https://github.com/Azure/azure-cosmosdb-spark/issues/318)).
* Cleanup unused imports
* Re-format the notebook cells

### Checklist:

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -248,7 +248,7 @@ To install the repo manually onto Databricks, follow the steps:
 
     ```{shell}
     cd Recommenders
-    zip -r Recommenders.egg .
+    zip -r Recommenders.egg ./reco_utils
     ```
 
 3. Once your cluster has started, go to the Databricks workspace, and select the `Home` button.
@@ -258,6 +258,12 @@ To install the repo manually onto Databricks, follow the steps:
 7. Next, click on the box that contains the text `Drop library egg here to upload` and use the file selector to choose the `Recommenders.egg` file you just created, and select `Open`.
 8. Click on the `Create library`. This will upload the egg and make it available in your workspace.
 9. Finally, in the next menu, attach the library to your cluster.
+
+In addition to the repo, you must install the following package as a library from PyPI:
+
+* `tqdm==4.31.1`
+
+You can follow instructions [here](https://docs.azuredatabricks.net/user-guide/libraries.html#install-a-library-on-a-cluster) for details on how to install packages from PyPI.
 
 </details>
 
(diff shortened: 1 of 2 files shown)


miguelgfierro:
is there a relationship between this PR and what we have been doing in the branch ignite?

loomlike:
@miguelgfierro Yes and no... 
* yes part:
   - found cosmosDB writing issue while working on `ignite` branch, so fix that here
   - used more explicit try / except in `ignite` branch, so applied them to here
* no part:
   - `ignite` branch uses notebook vm. This notebook assumes to be run on Databricks
   - `ignite` branch uses AzureML api for setup image environment. I haven't touched that part in this notebook (this notebook uses "written env file")

miguelgfierro:
I have been looking at both notebooks, the one you changed https://github.com/microsoft/recommenders/blob/765f51ee75b37e183817472534d53fa543c0ef90/notebooks/05_operationalize/als_movie_o16n.ipynb and the one that we have in ignite branch https://github.com/microsoft/recommenders/blob/ignite/notebooks/05_operationalize/als_movie_o16n.ipynb with the contributions of @Chris113113 and a modification I made https://github.com/microsoft/recommenders/blob/61728d869caef8c2a9b1cdbad1fec3e30b2e81cb/notebooks/05_operationalize/als_movie_o16n.ipynb that should run in DB and in a notebook.

The content is maybe 95% the same. I think we should have all together. If we use the branch ignite we just need to revert the hardcoded variables and leave it like this notebook and then add again the `if not is_jupyter:` couple of instances that we originally have.

Does this make sense?

loomlike:
@miguelgfierro I like the idea of using `is_jupyter`. The only thing we have to deal with is the beginning section's explanation and diagram about Databricks. Do we want to change that to be more general term, like "remote compute" or something like that?

miguelgfierro:
> @miguelgfierro I like the idea of using is_jupyter. The only thing we have to deal with is the beginning section's explanation and diagram about Databricks. Do we want to change that to be more general term, like "remote compute" or something like that?

yeah that's a good point. I think remote compute or compute target or whatever, and then explain that this notebook has been tested on Databricks and on AzureML Notebook VM

loomlike:
@miguelgfierro I will make another commit to this PR to merge ignite branch together.

loomlike:
I'm closing this PR for #1003 as promised

PR #983, opened 2019-11-25
Update generate_conda_file.py

dciborow:
### Description

I hit a strange error when running the CPU Unit Tests.

"ERROR: fastai 1.0.46 requires nvidia-ml-py3, which is not installed."

It looks like "fastai" is in the conda base. But "nvidia-ml-py3", is only included in the the GPU image. I think fastai should be moved into CONDA_GPU.

Here is a PR that makes the change.

### Checklist:

### DIFF
--- a/scripts/generate_conda_file.py
+++ b/scripts/generate_conda_file.py
@@ -57,13 +57,13 @@ CONDA_BASE = {
     "lightgbm": "lightgbm==2.2.1",
     "cmake": "cmake==3.14.0",
     "cornac": "cornac>=1.1.2",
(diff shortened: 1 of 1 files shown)


dciborow:
I hit a strange error when running the CPU Unit Tests.

"ERROR: fastai 1.0.46 requires nvidia-ml-py3, which is not installed."

It looks like "fastai" is in the conda base. But "nvidia-ml-py3", is only included in the the GPU image. I think fastai should be moved into CONDA_GPU.

Here is a PR that makes the change.

Issue #1013, opened 2019-12-16
[FEATURE] Create 4 deepdive notebooks for new MSRA sequential models


miguelgfierro:
### Other Comments
related to https://github.com/microsoft/recommenders/pull/1010

PR #1015, opened 2019-12-18
Miguel/docs sequential

miguelgfierro:
### Description

Added description and papers in docstrings

### Checklist:

### DIFF
--- a/reco_utils/recommender/deeprec/models/sequential/asvd.py
+++ b/reco_utils/recommender/deeprec/models/sequential/asvd.py
@@ -15,13 +15,13 @@ class ASVDModel(SequentialBaseModel):
     It extends ASVD with an attention module.
     
     ASVD: Y. Koren, "Factorization Meets the Neighborhood: a Multifaceted Collaborative 
-    Filtering Model", In Proceedings of the 14th ACM SIGKDD international conference on
-    Knowledge discovery and data mining, pages 426–434. ACM, 2008.
+    Filtering Model", in Proceedings of the 14th ACM SIGKDD international conference on
+    Knowledge discovery and data mining, pages 426–434, ACM, 2008.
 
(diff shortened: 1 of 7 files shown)


reco_utils/recommender/deeprec/models/sequential/asvd.py
```diff
@@ -10,6 +10,19 @@
...
+    
+    ASVD: Y. Koren, "Factorization Meets the Neighborhood: a Multifaceted Collaborative 
+    Filtering Model", in Proceedings of the 14th ACM SIGKDD international conference on
+    Knowledge discovery and data mining, pages 426–434, ACM, 2008.
+
+    A2SVD:  Z. Yu, J. Lian, A. Mahmoody, G. Liu and X. Xie, "Adaptive User Modeling with
+    Long and Short-Term Preferences for Personailzed Recommendation", in Proceedings of 
+    the 28th International Joint Conferences on Artificial Intelligence, IJCAI’19, 
+    Pages 4213-4219, AAAI Press, 2019.
+    """
```
miguelgfierro: @leavingseason, some questions:
1) do you think this is a good way to cite this?
2) the `ASVModel` class is actually using the attention module, right? is that correct? 
3) if 2) is true, I think we should rename `ASVDModel` to `A2SVDModel`. Will this be ok? I can do the refactoring while I'm working on #1013

Leavingseason: 1.  Yes, I think it is good
2.  Correct, It is indeed A2SVD.  Because  'A2SVD' is named by us and it still belongs to the group of ASVD, to avoid some confusion we just reused the name of `ASDV`' .   So it looks both find for me.  You can change it to 'A2SVD' .

miguelgfierro: perfect, I'll fix it on a different PR. I'm tracking it in https://github.com/microsoft/recommenders/issues/1020

PR #1011, opened 2019-12-16
Fix for issue #970

elogicaadith:
### Description

existing code, in base_model.py, under the fit function -

if self.hparams.save_model:

            if epoch % self.hparams.save_epoch == 0:
                checkpoint_path = self.saver.save(
                    sess=train_sess,
                    save_path= self.hparams.MODEL_DIR + "epoch_" + str(epoch),
                )

Modified code:
from os.path import join
.
.
.

if self.hparams.save_model:
                if epoch % self.hparams.save_epoch == 0:
                    save_path_str = join(self.hparams.MODEL_DIR, "epoch_" + str(epoch))
                    checkpoint_path = self.saver.save(
                        sess=train_sess,
                        save_path= save_path_str,
                    )

Azure Machine Learning service tests :- 

Code in AML -
image

AML outputs folder -
image

Local machine tests (using xdeepfm_criteo notebook in 00_quick_start) :-

prepare_hparams() -

image

local directory -

image

When running the xDeepFM script, the model was not getting saved to the ./outputs directory in the Azure Machine Learning Service. This was due to the way in which the string concatenation for save directory was set up - the model directory & the file names were concatenated without a "/" in between them. As a result, Azure Machine Learning Service was not able to save models.

### Related Issues

This is a fix for Issue #970 

### Checklist:

- [x] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [x] I have added tests covering my contributions.
- [ ] I have updated the documentation accordingly.

PR #1017, opened 2019-12-19
correct misspells

Leavingseason:
### Description

To correct some misspells in the last merge.

### Checklist:


reco_utils/recommender/deeprec/config/asvd.yaml
```diff
@@ -41,12 +41,12 @@ train:
...
 info:
     show_step : 100   #print training information after a certain number of mini-batch 
-    save_model: True   #whether to save modl
+    save_model: True   #whether to save models
     save_epoch : 1    #is save_model is set to True, save the model every save_epoch.
     metrics : ['auc','logloss']  #metrics for evaluation.
     pairwise_metrics : ['mean_mrr', 'ndcg', "group_auc"]
-    MODEL_DIR : ./tests/resources/deeprec/asvd/model/din_model/
-    SUMMARIES_DIR : ./tests/resources/deeprec/asvd/summary/din_summary/
+    MODEL_DIR : ./tests/resources/deeprec/asvd/model/asvd_model/
```
miguelgfierro: can you check this line? if at the top of the file you changed to slirec, shouldn't we change here as well slirec? Not sure if you prefer to have a more general name (like `sequential` or similar)

reco_utils/recommender/deeprec/models/base_model.py
```diff
@@ -437,10 +432,9 @@ def fit(self, train_file, valid_file, test_file=None):
 
             if self.hparams.save_model:
                 if epoch % self.hparams.save_epoch == 0:
-                    save_path_str = join(self.hparams.MODEL_DIR, "epoch_" + str(epoch))
                     checkpoint_path = self.saver.save(
                         sess=train_sess,
-                        save_path = save_path_str,
+                        save_path=self.hparams.MODEL_DIR + "epoch_" + str(epoch),
```
miguelgfierro: this was recently changed by @elogicaadith on this PR https://github.com/microsoft/recommenders/pull/1011, he said it was not working. Can you check?

reco_utils/recommender/deeprec/config/asvd.yaml
Leavingseason: The data is shared for different models, but the log and model path still need to be separated.

reco_utils/recommender/deeprec/models/base_model.py
Leavingseason: It's caused by the slash '/' char in path str. I haven't tried in in AML.  But since it is a recently changed, I will update it. Thanks!

PR #1027, opened 2020-01-06
small update: fix a bug in ndcg metrics

Leavingseason:
### Description

Small updates:  (1) fix a bug in NDCG calculation in deeprec; (2) add hit metrics (3) add more comments in yaml file

### Checklist:

### DIFF
--- a/reco_utils/recommender/deeprec/config/gru4rec.yaml
+++ b/reco_utils/recommender/deeprec/config/gru4rec.yaml
@@ -13,9 +13,9 @@ model:
     activation : [relu, relu] # activation function for DNN
     user_dropout: True 
     dropout : [0.3, 0.3]  #drop out values for DNN layer
-    item_embedding_dim : 32
-    cate_embedding_dim : 8
-    user_embedding_dim : 16
+    item_embedding_dim : 32 # the embedding dimension of items
+    cate_embedding_dim : 8  # the embedding dimension of categories
+    user_embedding_dim : 16 # the embedding dimension of users
 
 #train
 #init_method: normal,tnormal,uniform,he_normal,he_uniform,xavier_normal,xavier_uniform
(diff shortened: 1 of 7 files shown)


Leavingseason:
@miguel-ferreira  @yueguoguo  can you help to review it, thanks!

reco_utils/recommender/deeprec/config/gru4rec.yaml
```diff
@@ -29,26 +29,28 @@ train:
...
-    pairwise_metrics : ['mean_mrr', 'ndcg', "group_auc"]
-    MODEL_DIR : ./tests/resources/deeprec/gru4rec/model/gru4rec_model/
-    SUMMARIES_DIR : ./tests/resources/deeprec/gru4rec/summary/gru4rec_summary/
-    write_tfevents : True
+    show_step : 100   # print training information after a certain number of mini-batch 
+    save_model: True   # whether to save models
+    save_epoch : 1    # if save_model is set to True, save the model every save_epoch.
+    metrics : ['auc','logloss']  # metrics for evaluation.
+    pairwise_metrics : ['mean_mrr', 'ndcg@2;4;6', "group_auc"]  # pairwise metrics for evaluation, available when pairwise comparisons are needed
+    MODEL_DIR : ./tests/resources/deeprec/gru4rec/model/gru4rec_model/  # directory of saved models.
```
yueguoguo: Just a minor thing, do we want to use a temp dir for this, and after the running of a test they can be cleaned?

reco_utils/recommender/deeprec/deeprec_utils.py
```diff
@@ -539,11 +555,27 @@ def cal_metric(labels, preds, metrics):
         elif metric == "mean_mrr":
             mean_mrr = np.mean(mrr_score(labels, preds))
             res["mean_mrr"] = round(mean_mrr, 4)
-        elif metric == "ndcg":
-            ndcg_list = [2, 4, 6, 8, 10]
-            for r in ndcg_list:
-                ndcg_temp = np.mean(ndcg_score(labels, preds))
-                res["ndcg{0}".format(r)] = round(ndcg_temp, 4)
+        elif metric.startswith("ndcg"):  # format like:  ndcg@2;4;6;8
+            ndcg_list = [1, 2]
```
yueguoguo: If the metric is `ndcg` it will be equivalent to `ndcg@1;2`. Is that correct? Is this what we expect?

Leavingseason: yes, in case of user didn't assign the k value, we use [1,2] as the defualt

reco_utils/recommender/deeprec/config/gru4rec.yaml
Leavingseason: It is a good point. But actually it's not all about test cases...   the normal experiments also write model/logs into the tests/resources directory. So if the user want to persist the models, we should not delete the files.

PR #1039, opened 2020-01-22
Standardize predict interface using SAR standard

miguelgfierro:
### Description

for rating: predict
for ranking: recommend_top_k
also applied black

### Checklist:


reco_utils/recommender/surprise/surprise_utils.py
```diff
@@ -74,7 +74,7 @@ def compute_rating_predictions(
     return predictions.drop(["details", "r_ui"], axis="columns")
 
 
-def compute_ranking_predictions(
+def recommend_k_items(
```
anargyri: This function does not recommend k items. It computes the predictions for all users and items.

miguelgfierro: ohh good catch, the cut of k is done in fact when the metric is computed https://github.com/microsoft/recommenders/blob/master/notebooks/02_model/surprise_svd_deep_dive.ipynb

miguelgfierro: also check with @yueguoguo whether or not we are using the threshold cut https://github.com/microsoft/recommenders/issues/1041

gramhagen: if we don't want to work with the scores as a matrix like sar does we could use a heap

```
from heapq import heappush, heappushpop

users = data[usercol].unique()
items = data[itemcol].unique()
preds_lst = np.zeros([len(users), k])

for user_idx, user in enumerate(users):
    heap = []
    for item in items:
        score = algo.predict(user, item).est
        if len(heap) < k:
            heappush(heap, score)
        elif score > heap[k - 1]:
            heappushpop(heap, score);
    preds_lst[user_idx] = heap[::-1]
```

anargyri: `heapq` also has an `nlargest()` method. Not sure which way would be faster (depends on which algorithm it uses).

gramhagen: true, but it wasn't clear to me if that method tries to sort the items (which would be unnecessary) or leverages the heap?

anargyri: https://github.com/python/cpython/blob/61b3484cdf27ceca1c1069a351487d2db4b2b48c/Lib/heapq.py#L395
It looks similar to your for loop.

miguelgfierro:
hey guys @anargyri @gramhagen, I reverted the interface of ranking in surprise to do the work on a different PR. The related issue is here: https://github.com/microsoft/recommenders/issues/1042, please take a look once the tests have passed

PR #1159, opened 2020-07-23
Staging to master to add news reco and DKN algos with MIND

miguelgfierro:
### Checklist:


miguelgfierro:
error in DKN notebook:

```
model = DKN(hparams, input_creator)

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-607bbc4e411e> in <module>
----> 1 model = DKN(hparams, input_creator)

~/repos/Recommenders/reco_utils/recommender/deeprec/models/dkn.py in __init__(self, hparams, iterator_creator)
     80                     )
     81 
---> 82         super().__init__(hparams, iterator_creator, graph=self.graph)
     83 
     84     def _init_embedding(self, file_path):

~/repos/Recommenders/reco_utils/recommender/deeprec/models/base_model.py in __init__(self, hparams, iterator_creator, graph, seed)
     31 
     32         self.graph = graph if graph is not None else tf.Graph()
---> 33         self.iterator = iterator_creator(hparams, self.graph)
     34         self.train_num_ngs = (
     35             hparams.train_num_ngs if "train_num_ngs" in hparams else None

~/repos/Recommenders/reco_utils/recommender/deeprec/io/dkn_iterator.py in __init__(self, hparams, graph, col_spliter, ID_spliter)
     73                 click_news_index = []
     74                 click_news_entity_index = []
---> 75                 if len(user_history) > self.history_size:
     76                     user_history = user_history[-self.history_size:]
     77                 for newsid in user_history:

TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

reco_utils/recommender/deeprec/models/base_model.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, hparams, iterator_creator, graph=None, seed=None):
             seed (int): Random seed.
         """
         self.seed = seed
-        tf.set_random_seed(seed)
+        tf.compat.v1.set_random_seed(seed)
```
loomlike, line 29: I we don't use tf v2 in this file,
we can simply do:
```
import tensorflow.compat.v1 as tf
tf.disable_v2_behavior()
```
and use `tf.`

See [tf official doc](https://www.tensorflow.org/guide/migrate)

reco_utils/recommender/newsrec/models/base_model.py
```diff
@@ -36,13 +38,17 @@ def __init__(self, hparams, iterator_creator, seed=None):
             seed (int): Random seed.
         """
         self.seed = seed
-        tf.set_random_seed(seed)
+        tf.compat.v1.set_random_seed(seed)
```
loomlike, line 41: Same here. See above comment.

tests/integration/test_notebooks_gpu.py
```diff
@@ -264,14 +264,14 @@ def test_slirec_quickstart_integration(
     "epochs, seed, expected_values",
     [
         (
-            10,
+            8,
```
loomlike, line 267: Any reason we changed test epochs in these tests?

tests/smoke/test_newsrec_model.py
```diff
@@ -11,97 +11,215 @@
...
+    train_news_file = os.path.join(tmp, "train", r"news.tsv")
+    train_behaviors_file = os.path.join(tmp, "train", r"behaviors.tsv")
+    valid_news_file = os.path.join(tmp, "valid", r"news.tsv")
+    valid_behaviors_file = os.path.join(tmp, "valid", r"behaviors.tsv")
+    wordEmb_file = os.path.join(tmp, "utils", "embedding.npy")
+    userDict_file = os.path.join(tmp, "utils", "uid2index.pkl")
+    wordDict_file = os.path.join(tmp, "utils", "word_dict.pkl")
+    yaml_file = os.path.join(tmp, "utils", r"nrms.yaml")
+
+    if not os.path.exists(train_news_file):
```
loomlike, line 30: I believe `tmp` is the function level, meaning a temp folder will be generated for each test function and thus the data will be downloaded anyway (as path.exists will return False). We have couple of tests that use the same dataset, and this dataset will be downloaded multiple times.

We probably can create module-level tmp folder and pass that so that a suite of tests in the same module can share the same file if necessary.

reco_utils/recommender/deeprec/models/base_model.py
miguelgfierro: this is to avoid the warnings, also in the future, if we decide to use tf2, this will work out of the box

tests/integration/test_notebooks_gpu.py
miguelgfierro: it finished earlier, not special reason

reco_utils/recommender/newsrec/models/base_model.py
miguelgfierro: this is to avoid the warnings, also in the future, if we decide to use tf2, this will work out of the box

tests/smoke/test_newsrec_model.py
miguelgfierro: in this case, in fact, maybe we should add a fixture that downloaded all the data once and then that to be used by all the other functions. I agree that we should have another tmp for the context module, and maybe another one for system

anargyri:
I don't understand some things about GeoIMC. There are two Matlab files related to user / item features checked in. They seem to be used by the data set loader for Movielens100K. Any ideas what happens with these files?

miguelgfierro:
> I don't understand some things about GeoIMC. There are two Matlab files related to user / item features checked in. They seem to be used by the data set loader for Movielens100K. Any ideas what happens with these files? https://github.com/microsoft/recommenders/blob/221feb50cfa0680adb4bc4f62208595627e16281/reco_utils/recommender/geoimc/geoimc_data.py

where are the matlab files?

@SatyadevNtv could you please take a look at Andreas' comment?

SatyadevNtv:
> I don't understand some things about GeoIMC. There are two Matlab files related to user / item features checked in. They seem to be used by the data set loader for Movielens100K. Any ideas what happens with these files? https://github.com/microsoft/recommenders/blob/221feb50cfa0680adb4bc4f62208595627e16281/reco_utils/recommender/geoimc/geoimc_data.py

Those are the files with side information (features) of users/items derived in the similar way as [this](https://aaai.org/ocs/index.php/AAAI/AAAI17/paper/view/14676/13916) work.

anargyri:
> > I don't understand some things about GeoIMC. There are two Matlab files related to user / item features checked in. They seem to be used by the data set loader for Movielens100K. Any ideas what happens with these files? https://github.com/microsoft/recommenders/blob/221feb50cfa0680adb4bc4f62208595627e16281/reco_utils/recommender/geoimc/geoimc_data.py
> 
> Those are the files with side information (features) of users/items derived in the similar way as [this](https://aaai.org/ocs/index.php/AAAI/AAAI17/paper/view/14676/13916) work.

Is there code that generates those Matlab files? Are they available somewhere?

SatyadevNtv:
> > > I don't understand some things about GeoIMC. There are two Matlab files related to user / item features checked in. They seem to be used by the data set loader for Movielens100K. Any ideas what happens with these files? https://github.com/microsoft/recommenders/blob/221feb50cfa0680adb4bc4f62208595627e16281/reco_utils/recommender/geoimc/geoimc_data.py
> > 
> > 
> > Those are the files with side information (features) of users/items derived in the similar way as [this](https://aaai.org/ocs/index.php/AAAI/AAAI17/paper/view/14676/13916) work.
> 
> Is there code that generates those Matlab files? Are they available somewhere? I mean, you wrote some Matlab code, didn't you? Is it hard to convert this to Python?

I used python itself to generate those files. I dumped in `mat` since I was experimenting with sparse representation `smat` and all during that time. 

> In general, if one wants to apply GeoIMC to a data set other than Movielens, what does one need to do to generate the features?
> Looking at the paper, it seems what they are doing is one-hot encoding of the user and item features from Movielens, am I right?

Yes. True, it's a one-hot encoding. 
So, essentially, GeoIMC just needs side information, one of the way to generate them would be as the mentioned paper does (there can be other ways also). Hence, in the view of keeping the `ML100K` data class generic and transparent to these, I didn't include the specific code.

anargyri:
@SatyadevNtv So, since this is just one-hot encoding of the features, there is no need for matlab files and the notebook code can become more transparent. That is, instead of `download_geoimc_features` in `geoimc_movielens.ipynb` you could put your python code that reads the features data and converts to one-hot encoding inside `_load_features()` in `geoimc_data.py`.

SatyadevNtv:
> @SatyadevNtv So, since this is just one-hot encoding of the features, there is no need for matlab files and the notebook code can become more transparent. That is, instead of `download_geoimc_features` in `geoimc_movielens.ipynb` you could put your python code that reads the features data and converts to one-hot encoding inside `_load_features()` in `geoimc_data.py`.

Sure. Definitely. I will update it accordingly.

Thanks.

SatyadevNtv:
> @SatyadevNtv So, since this is just one-hot encoding of the features, there is no need for matlab files and the notebook code can become more transparent. That is, instead of `download_geoimc_features` in `geoimc_movielens.ipynb` you could put your python code that reads the features data and converts to one-hot encoding inside `_load_features()` in `geoimc_data.py`.

Included at 63dadd78fa08bd7291a62f0443ee5824455b640d

tests/smoke/test_newsrec_model.py
loomlike: do you want to create an issue for that and handle separately from this PR?

miguelgfierro: yes, let me create it

miguelgfierro: https://github.com/microsoft/recommenders/issues/1169

loomlike:
qq about the folder names in examples:

02_model_collaborative_filtering
02_model_content_based_filtering
02_model_hybrid
03_evaluate
04_model_select_and_optimize

These are our current folder names. Seems prefix of 02 is "02_model". Then, I think 04 should be changed to 04_select_and_optimize_models

PR #1186, opened 2020-08-21
spark installation issue

miguelgfierro:
### Description

I created a new dsvm and had some problems when installing the spark environment. This fixes this problem

### Checklist:

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -85,7 +85,7 @@ To install the PySpark environment:
 
 > Additionally, if you want to test a particular version of spark, you may pass the --pyspark-version argument:
 >
->     python tools/generate_conda_file.py --pyspark-version 2.4.0
+>     python tools/generate_conda_file.py --pyspark-version 2.4.5
 
 Then, we need to set the environment variables `PYSPARK_PYTHON` and `PYSPARK_DRIVER_PYTHON` to point to the conda python executable.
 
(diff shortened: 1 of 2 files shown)


SETUP.md
```diff
@@ -107,7 +107,7 @@ Then, create the file `$RECO_ENV/etc/conda/activate.d/env_vars.sh` and add:
     export PYSPARK_PYTHON=$RECO_ENV/bin/python
     export PYSPARK_DRIVER_PYTHON=$RECO_ENV/bin/python
     export SPARK_HOME_BACKUP=$SPARK_HOME
-    unset SPARK_HOME
+    export SPARK_HOME=/dsvm/tools/spark/current
```
anargyri, line 110: This works only for DSVM, not in general for Linux and MacOS.

miguelgfierro: yes, that's right. We need to change this because unset creates a problem. So we can have this path, or add a dummy one, or maybe add a comment here saying that please change the path to your current SPARK_HOME...

anargyri: Well, this path is the standard SPARK_HOME that DSVM is deployed with. So why do you need to do this line, unless someone modified SPARK_HOME after deploying the DSVM?

miguelgfierro: In the dsvm `SPARK_HOME` is not defined, at least in two VMs I saw

anargyri: I see. It works on the NC6_Promo though.

miguelgfierro:
TODO: review this with a new DSVM

miguelgfierro:
On a new DSVM, SPARK_HOME is set correctly
```
$ echo $SPARK_HOME
/dsvm/tools/spark/current
```

PR #1297, opened 2021-02-10
Update pandas_df_utils.py

yueguoguo:
Fixed the issues of the negative sampler

### Description
Fix the issue in the negative sampling function

### Related Issues
N/A

### Checklist:

### DIFF
--- a/reco_utils/dataset/pandas_df_utils.py
+++ b/reco_utils/dataset/pandas_df_utils.py
@@ -266,21 +266,18 @@ class LibffmConverter:
 
 def negative_feedback_sampler(
     df,
-    col_user=DEFAULT_USER_COL,
-    col_item=DEFAULT_ITEM_COL,
-    col_label=DEFAULT_LABEL_COL,
+    col_user="CustomerId",
+    col_item="ProductId",
+    col_label="Label",
     ratio_neg_per_user=1,
     seed=42,
 ):
     """Utility function to sample negative feedback from user-item interaction dataset.
-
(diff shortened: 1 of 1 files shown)


reco_utils/dataset/pandas_df_utils.py
```diff
@@ -293,10 +290,8 @@ def negative_feedback_sampler(
         If the samples exceed the number of total possible negative feedback samples, it will be reduced to the number
         of all the possible samples.
         seed (int): seed for the random state of the sampling function.
-
```
miguelgfierro, line 296: minor thing here, if we remove these intermediate lines, the sphinx documentation doesn't parse the docstring correctly

miguelgfierro (approved):

anargyri (approved):
I think we can make negative sampling more efficient. Currently we create all user item pairs, which  is time and memory consuming i.e. O(n_users x n_items). Instead, we could sample indexes iteratively, which would be O(n_sample).

reco_utils/dataset/pandas_df_utils.py
miguelgfierro: hey @yueguoguo, since this is a small fix and you are on leave, I'll go ahead and fix it

miguelgfierro:
``` I think we can make negative sampling more efficient. Currently we create all user item pairs, which is time and memory consuming i.e. O(n_users x n_items). Instead, we could sample indexes in O(n_sample) time e.g. using numpy sampling functions.

+ 1 to using numpy

PR #1300, opened 2021-02-12
Dciborow/pandas cg update

dciborow:
### Description

Updating Deps to recent version, 

CG has alerted that these deps have high or critical issues.
image

### Checklist:

- [x ] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [x ] I have added tests covering my contributions.
- [x ] I have updated the documentation accordingly.
- [x ] This PR is being made to `staging branch` and not to `main branch`.

### DIFF
--- a/tools/generate_conda_file.py
+++ b/tools/generate_conda_file.py
@@ -71,7 +71,7 @@ CONDA_GPU = {
 }
 
 PIP_BASE = {
-    "azureml-sdk[notebooks,fboard]": "azureml-sdk[notebooks,tensorboard]==1.0.69",
+    "azureml-sdk[notebooks,tensorboard]": "azureml-sdk[notebooks,tensorboard]==1.0.69",
     "azure-storage-blob": "azure-storage-blob<=2.1.0",
     "azure-cli-core": "azure-cli-core==2.0.75",
     "azure-mgmt-cosmosdb": "azure-mgmt-cosmosdb==0.8.0",
--- a/tools/generate_conda_file.py
+++ b/tools/generate_conda_file.py
@@ -71,7 +71,7 @@ CONDA_GPU = {
 }
 
 PIP_BASE = {
-    "azureml-sdk[notebooks,tensorboard]": "azureml-sdk[notebooks,tensorboard]==1.0.69",
+    "azureml-sdk[notebooks,fboard]": "azureml-sdk[notebooks,tensorboard]==1.0.69",
     "azure-storage-blob": "azure-storage-blob<=2.1.0",
     "azure-cli-core": "azure-cli-core==2.0.75",
     "azure-mgmt-cosmosdb": "azure-mgmt-cosmosdb==0.8.0",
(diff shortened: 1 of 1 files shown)


tools/generate_conda_file.py
```diff
@@ -87,13 +87,13 @@
     "pymanopt": "pymanopt==0.2.5",
     "xlearn": "xlearn==0.40a1",
     "transformers": "transformers==2.5.0",
-    "tensorflow": "tensorflow==1.15.2",
+    "tensorflow": "tensorflow>=2.2.2,<2.4",
```
miguelgfierro, line 90: ufff, right now we can't upgrade to TF 2, a large part of the code will fail. @dciborow do you now if newer versions of 1.15 are whitelisted https://pypi.org/project/tensorflow/#history?

tools/generate_conda_file.py
```diff
@@ -46,7 +46,7 @@
     "mock": "mock==2.0.0",
     "nltk": "nltk>=3.4",
     "numpy": "numpy>=1.13.3",
-    "pandas": "pandas>=0.23.4,<1.0.0",
```
miguelgfierro, line 49: we had problems in the past with pandas >1 https://github.com/microsoft/recommenders/pull/1052. There are multiple versions from 0.23.4 to <1 https://pypi.org/project/pandas/#history, is there an easy easy way to know which of them are ok?

dciborow: CG says we need to be above 1.0.3

tools/generate_conda_file.py
dciborow: no... Jeremy just looked into this 2.2 is the lowest we can use.

miguelgfierro: Bufff that's really bad news, we need to review a lot of the code.

FYI @Leavingseason

gramhagen: re: tensorflow this is my suggestion
1. move functions from reco-utils/common/tf_utils -> reco-utils/recommender/wide_deep/wide_deep_utils (they don't seem to be used outside wide_deep examples)
2. refactor tests so to handle (1)
3. remove use of tensorflow in azureml/wide_deep_training.py (it's only used for printing the version)
4. remove tensorflow as a dependency altogether from core package (which would just be reco-utils/* - recommender module)

tools/generate_conda_file.py
gramhagen: can we set to >1.0.3,<2.0

tools/generate_conda_file.py
dciborow: but.... the tests..... they pass...... can we start off with getting some tests that fail?

tools/generate_conda_file.py
dciborow: Ill should have added upper bound with the latest release. We don't want to up this pin until the latest version is released and verified... 
im not worrying about this for the rest of the packages in this PR, at some point I will.

tools/generate_conda_file.py
dciborow: its not in the PR anymore. Nothing to talk about here....

miguelgfierro (approved):

PR #1305, opened 2021-02-12
More efficient negative sampling

anargyri:
### Description
Improves the negative feedback sampling, by drawing a sample of indexes of items instead of creating the full user - item data frame.
The user_item_pairs() function is not needed and removed. I also added a unit test for the negative sampling.

### Checklist:

- [ x] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ x] I have added tests covering my contributions.
- [ x] I have updated the documentation accordingly.
- [ x] This PR is being made to `staging branch` and not to `main branch`.


reco_utils/dataset/pandas_df_utils.py
```diff
@@ -316,54 +316,40 @@ def negative_feedback_sampler(
...
-                        replace=False,
-                    )
-                ],
-                ignore_index=True,
-                sort=True,
+    rng = np.random.default_rng(seed=seed)
+
+    def sample_items(user_df):
+        # Sample negative items for the data frame restricted to a specific user
+        unseen_items = np.setdiff1d(items, user_df[col_item])
```
anargyri, line 324: This can be more efficient (you don't need to compute the set difference). I will modify it on Monday.

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -316,53 +316,38 @@ def negative_feedback_sampler(
...
-                        random_state=seed,
-                        replace=False,
-                    )
-                ],
-                ignore_index=True,
-                sort=True,
-            )
+    rng = np.random.default_rng(seed=seed)
+
+    def sample_items(user_df):
```
miguelgfierro, line 322: super nice function!
a minor thing, maybe we want to parametrize the positive and negative values? instead of hardcoding 0 and 1

miguelgfierro, line 322: could the `sample_items` function be used for other purposes? in that case it might be worthwhile to consider taking it out of the `negative_feedback_sampler`

@yueguoguo @simonzhaoms for input

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -316,53 +316,38 @@ def negative_feedback_sampler(
...
+        new_df = pd.DataFrame(
+            data={
+                col_user: user_df[col_user][0],
+                col_item: new_items,
+                col_label: 0,
+            }
         )
+        return pd.concat([user_df, new_df], ignore_index=True)
+
+    res_df = df.copy()
```
miguelgfierro, line 341: I think we should consider to add an `inplace` parameter, if it is True we don't copy the full df, otherwise, we copy

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -316,53 +316,38 @@ def negative_feedback_sampler(
...
+    res_df[col_label] = 1
+    return (
+        res_df.groupby(col_user)
+        .apply(sample_items)
         .reset_index(drop=True)
         .sort_values(col_user)
+        .rename(columns={col_label: "feedback"})
```
miguelgfierro, line 348: maybe parametrize the column name

reco_utils/dataset/pandas_df_utils.py
```diff
@@ -316,53 +316,38 @@ def negative_feedback_sampler(
...
         )
+        return pd.concat([user_df, new_df], ignore_index=True)
+
+    res_df = df.copy()
+    res_df[col_label] = 1
+    return (
+        res_df.groupby(col_user)
+        .apply(sample_items)
         .reset_index(drop=True)
         .sort_values(col_user)
```
miguelgfierro, line 347: `.sort_values(col_user)` -> do we need this? we will save time if we remove it

FYI @yueguoguo

miguelgfierro (approved):
super nice function @anargyri! 

I added some questions, please review

anargyri:
Thanks, these are good suggestions. I will look into them. Btw I could not find a call to `negative_feedback_sampler()` anywhere in the code (except the unit test). So, maybe this function was used once or twice only; maybe @yueguoguo you know more, when you are back.

reco_utils/dataset/pandas_df_utils.py
anargyri: Not really. This function assumes that there is only a single user in `user_df`.

reco_utils/dataset/pandas_df_utils.py
anargyri: I think this cannot be really an inplace function because it adds new rows to the data frame. So it involves concatenations of data frames, which inevitably create new copies.

miguelgfierro: makes sense

miguelgfierro:
hey @anargyri, is the PR finished? feel free to merge if so, or I can do it if you prefer

anargyri:
> hey @anargyri, is the PR finished? feel free to merge if so, or I can do it if you prefer

There is a failure (pyspark). This must be some unrelated bug. Should I merge regardless?

miguelgfierro:
> There is a failure (pyspark). This must be some unrelated bug. Should I merge regardless?

yeah, this error was fixed by @gramhagen here https://github.com/microsoft/recommenders/pull/1341

PR #1343, opened 2021-03-12
Fixed seed

miguelgfierro:
### Checklist:


anargyri (approved):
Numpy's more recent approach is to use a random generator object, see https://numpy.org/doc/stable/reference/random/index.html#random-quick-start
So, we could define a single generator object in order to avoid our problems with arbitrary seeds (since the seed will be specified in a single place in the code).

PR #1372, opened 2021-04-13
Create pypi-publish-action.yml

dciborow:
### Checklist:

### DIFF
--- a/.github/workflows/pypi-publish-action.yml
+++ b/.github/workflows/pypi-publish-action.yml
@@ -1,0 +1,26 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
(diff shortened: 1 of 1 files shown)


anargyri:
Could you move this from main to staging?

miguelgfierro:
> Could you move this from main to staging?

@anargyri since there is only one new file, if we just change it in the dashboard, I think we shouldn't get a lot of errors. I can help if you want

anargyri:
> > Could you move this from main to staging?
> 
> @anargyri since there is only one new file, if we just change it in the dashboard, I think we shouldn't get a lot of errors. I can help if you want

So if we merge it to main this action will be active, right? This needs to be the last change we do after we have finalized everything else (package name, setup.py etc.). I'd also rather test things with manual publishing before we automate. I suggest we keep it open until we have the rest in place.

miguelgfierro:
@dciborow this PR is the same thing as https://github.com/microsoft/recommenders/pull/1481?

laserprec:
Should we close this PR since it looks like a duplicate of  #1481?

PR #1412, opened 2021-05-21
Pipeline to generate the release and pass all the tests

miguelgfierro:
### Related Issues

1408

### Checklist:


anargyri (commented):
Maybe we want to keep dev-requirements.txt ? Or change CONTRIBUTING.md where we reference it.

tests/ci/azure_pipeline_test/release_pipeline.yml
```diff
@@ -0,0 +1,89 @@
...
+    pip_opts: "[spark]"
+    pytest_markers: "not notebooks and spark and not gpu"
+
+- template: dsvm_linux_template.yml
+  parameters:
+    test_types:
+    - unit
+    task_name: "Test - Unit Notebook Linux Spark"
+    conda_env: "relase_unit_notebook_linux_spark"
+    conda_opts: "python=3.6"
```
anargyri, line 75: I think we can start using 3.7, is there any obstacle to that?

miguelgfierro: yeah I can create another  set with 3.7, I'll try to create a matrix

miguelgfierro:
> Maybe we want to keep dev-requirements.txt ? Or change CONTRIBUTING.md where we reference it.

I removed this because it was only installing pytest and black, and we were only installing pytest, so it is added directly into the pipeline and we removed an extra file we don't really need

anargyri:
> > Maybe we want to keep dev-requirements.txt ? Or change CONTRIBUTING.md where we reference it.
> 
> I removed this because it was only installing pytest and black, and we were only installing pytest, so it is added directly into the pipeline and we removed an extra file we don't really need

Yes, but besides the pipeline, it can be used by people who contribute to the repo. This is why we reference it in CONTRIBUTING.md.

PR #1422, opened 2021-06-02
simpler version in setup

miguelgfierro:
### Checklist:


anargyri (approved):

gramhagen:
the only issue with this is that it will only work if you are in the same directory as setup.py. that's probably the most likely spot anyone would run this, so I think it's fine to make this simplification, but just wanted to point that out.

Issue #1408, opened 2021-05-18
[FEATURE] Create ADO pipeline for generating pypi package


miguelgfierro:
### Description

Andreas:
- [x] Remove the sys path include
- [x] Change the [current pipeline](https://github.com/microsoft/recommenders/tree/main/tests/ci/azure_pipeline_test) so it installs the package locally, instead of using the path
- [x] Update setup.py
- [x] Fix issue with spark tests
- [x] Fix the issue with GPU version to match TF 1.15
- [x] Update documentation to reflect the new installation process
- [x] Review docs evaluation and datasets
- [x] test if the wheel package works on Databricks

Miguel:
- [x] Add a new yaml file that when [there is a new tag](https://github.com/microsoft/LightGBM/blob/master/.vsts-ci.yml#L5), builds a package with bdist, installs it, executes all tests (unit, smoke and integration) 
- [x] Create the github release draft
- [x] Upload artifacts to the github release draft: wheel and compressed code
- [x] Publish the wheel to the package limbo only when we are executing a release
- [x] Add fix to the spark tests in the pipeline
- [x] BUG: remove the wildcard in the installation and use the full name programmatically [see comment](https://github.com/microsoft/recommenders/issues/1408#issuecomment-848898749)
- [x] Check why we are installing all the deps in CPU, instead of only the CPU ones [see run](https://dev.azure.com/best-practices/recommenders/_build/results?buildId=47347&view=logs&j=ea6d793f-c9bd-55e6-b1b2-e25bd3674a1c&t=58ebb881-49a2-5c8f-c3f1-1abae14a2a25)
- [x] Make sure we are using the package when running the tests, [see comment here](https://github.com/microsoft/recommenders/issues/1408#issuecomment-850351697)
- [x] Fix issue with xlearn, it was using gpu deps that were not needed
- [x] See if we can simplify the code when forcing exit on error by removing exit -1 on each line and add other instructions. See [comment](https://github.com/microsoft/recommenders/issues/1408#issuecomment-850391727)
- [x] if smoke tests fail, we don't continue with the integration tests (nice to have feature). [Working now](https://dev.azure.com/best-practices/recommenders/_build/results?buildId=47475&view=logs&j=b097bf04-c10b-513d-2c77-b9ddcc54001a&t=2a7f1c25-fa3f-589c-0e5d-3acbe763a1a5)
- [ ] Review docs common, reco, tuning:
    - [x] common
    - [ ] recommender
    - [ ] tunning
[...]

PR #1427, opened 2021-06-04
Pipeline for release and updated tests

miguelgfierro:
### Description

Details https://github.com/microsoft/recommenders/issues/1408

### Checklist:


setup.py
```diff
@@ -64,22 +61,27 @@
...
+        "torch==1.2.0",  # last os-common version with CUDA 10.0 support
         "fastai>=1.0.46,<2",
-        "torch==1.2.0",  # last os-common version with cuda 10.0 support
     ],
     "spark": [
         "databricks_cli>=0.8.6,<1",
         "pyarrow>=0.8.0,<1.0.0",
         "pyspark>=2.4.5,<3.0.0",
     ],
+    "xlearn": [
```
miguelgfierro, line 73: @anargyri can you please confirm, xlearn goes here or does it go inside extras_requre?

miguelgfierro:
there is a very weird behavior that is happening. If I commit to the branch and the test are triggered automatically, then there is an error: https://dev.azure.com/best-practices/recommenders/_build/results?buildId=47825&view=results. However, if I run the pipeline manually it works: https://dev.azure.com/best-practices/recommenders/_build/results?buildId=47811&view=results I wonder if somehow, the automatic trigger doesn't clone the repo and branch every time?

gramhagen:
> there is a very weird behavior that is happening. If I commit to the branch and the test are triggered automatically, then there is an error: https://dev.azure.com/best-practices/recommenders/_build/results?buildId=47825&view=results. However, if I run the pipeline manually it works: https://dev.azure.com/best-practices/recommenders/_build/results?buildId=47811&view=results I wonder if somehow, the automatic trigger doesn't clone the repo and branch every time?

yeah, that is defined in the build pipeline, it looks like clean is not set to true (at least for this pipeline). You will need to edit the pipeline -> triggers -> yaml -> get sources -> clean -> set to true
https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/pipeline-options-for-git?view=azure-devops&tabs=yaml

probably should check the other pipelines too

miguelgfierro:
> yeah, that is defined in the build pipeline, it looks like clean is not set to true (at least for this pipeline). You will need to edit the pipeline -> triggers -> yaml -> get sources -> clean -> set to true
https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/pipeline-options-for-git?view=azure-devops&tabs=yaml

do you know if this is something that we can add in the yaml file directly?

gramhagen:
> > yeah, that is defined in the build pipeline, it looks like clean is not set to true (at least for this pipeline). You will need to edit the pipeline -> triggers -> yaml -> get sources -> clean -> set to true
> > https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/pipeline-options-for-git?view=azure-devops&tabs=yaml
> 
> do you know if this is something that we can add in the yaml file directly?

yes, in the documentation listed you can either add a checkout step, or use the workspace setting for any job

tests/integration/examples/test_notebooks_gpu.py
```diff
@@ -206,24 +206,6 @@ def test_xdeepfm_integration(
         )
 
 
-# TODO: remove tf dependency in this notebook and drop gpu marker
```
anargyri, line 209: Why remove the test? We could use an "xlearn" marker to disable it from the pipeline, and maybe add a separate pipeline in the future.

anargyri: On a DSVM, xlearn should be pip installable, because cmake is already installed at `/usr/local/bin/cmake`. So the test would run to completion, wouldn't it?

tests/integration/examples/test_notebooks_python.py
```diff
@@ -246,3 +247,20 @@ def test_geoimc_integration(notebooks, output_notebook, kernel_name, expected_va
 
     for key, value in expected_values.items():
         assert results[key] == pytest.approx(value, rel=TOL, abs=ABS_TOL)
+
+
+@pytest.mark.integration
+@pytest.mark.skip(reason="Tests removed due to installation incompatibilities")
```
anargyri, line 253: We could add a separate pipeline for this test in the future, since xlearn can be pip installed on a DSVM.

miguelgfierro: yeah, good suggestion, maybe consider it for a new PR?

anargyri: Yes, I created an issue about this https://github.com/microsoft/recommenders/issues/1437

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
```diff
@@ -21,51 +21,103 @@ parameters:
...
-        export PYSPARK_DRIVER_PYTHON=`which python`
+        pip install pytest>=3.6.4 || exit -1
+
+        if [[ "${{ parameters.conda_env }}" == *"spark"* ]]; then
+          export PYSPARK_PYTHON=`which python`
+          export PYSPARK_DRIVER_PYTHON=`which python`
+          export SPARK_DIR=$PWD/build/spark
+          rm -rf $SPARK_DIR || exit -1
+          mkdir -p $SPARK_DIR || exit -1
+          wget https://mirrors.ukfast.co.uk/sites/ftp.apache.org/spark/spark-2.4.8/spark-2.4.8-bin-hadoop2.7.tgz -O $SPARK_DIR/spark.tgz || exit -1
```
gramhagen, line 94: i thought installing pip installing pyspark was sufficient, is that not the case? do we need the spark binaries? if that's the case is the version of spark important to synchronize?

anargyri (approved):
Excellent work @miguelgfierro, you are now officially the master of build pipelines!

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
```diff
@@ -21,51 +21,103 @@ parameters:
...
+    displayName: 'Install Dependencies'
+
   - ${{ each test in parameters.test_types }}:
     - script: |
         eval "$(conda shell.bash hook)"
+
         conda activate ${{ parameters.conda_env }}
-        export PYSPARK_PYTHON=`which python`
-        export PYSPARK_DRIVER_PYTHON=`which python`
+        pip install pytest>=3.6.4 || exit -1
```
gramhagen, line 86: curious why this was moved out of dev-requirements.txt and into the pipeline?

tests/ci/azure_pipeline_test/release_pipeline.yml
```diff
@@ -0,0 +1,209 @@
...
+  - release_unit_linux_gpu
+  - release_unit_notebook_linux_gpu
+  - release_unit_linux_spark
+  - release_unit_notebook_linux_spark
+  - release_nightly_linux_cpu
+  - release_nightly_linux_gpu
+  - release_nightly_linux_spark
+  condition: succeeded()
+  steps:
+    # Create archives with complete source code included
```
gramhagen, line 142: what are the archives for and do we need separate files for reco_utils and examples? also, do we need both tarballs and zips?

gramhagen (approved):
most of my comments are questions so I can better understand how things work, maybe we should set up some time to talk this through, i don't think any of them should block this.

it would be nice though to add a readme that explains how the release process works, it looks like we need to tag a release starting with 0. or 1. and then some artifacts are pushed to github, then we manually upload them to pypi?

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
miguelgfierro: same answer here: https://github.com/microsoft/recommenders/pull/1427#discussion_r649911784, closing this one so we follow the conversation in the other message

dev-requirements.txt
```diff
@@ -1,6 +0,0 @@
-# to run tests ensure you have installed reco-utils with necessary optional dependencies
```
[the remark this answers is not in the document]
gramhagen: it would be nice to have something for external contributors to use to setup an environment to run tests and format the code based on contribution guidelines. if we want to reduce files we can push that into setup.py.

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
miguelgfierro: yeah we need the binaries, in the old test VM we have set up the right SPARK_HOME, but what I wanted to do here is a more portable approach that downloads the binary, sets SPARK_HOME, etc. If in the future, we want to move from DSVM we have now to a hosted machine, it will be just a copy pase. 

About the question of syncronizing the version, that's a good point. Right now we have in setup.py `"pyspark>=2.4.5,<3.0.0"`, that resolves to 2.4.8, which is the same version we have in the binaries. Here I see two options:
1) fix the spark version of the setup.py to 2.4.8
2) get the pyspark version that is installed dynamically (maybe `pip list | grep pyspark | awk "get me the value"` and then input this to the url when we download the binaries

tests/ci/azure_pipeline_test/release_pipeline.yml
miguelgfierro: being super honest, the only reason is that I had the code of both, but happy to kill the tar one if you guys think it will simplify

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
```diff
@@ -21,51 +21,103 @@ parameters:
...
+      conda env remove -n ${{ parameters.conda_env }} || exit -1
+
+      # Create a clean conda environment
+      conda create -fyn ${{ parameters.conda_env }} ${{ parameters.conda_opts }}
       eval "$(conda shell.bash hook)"
       conda activate ${{ parameters.conda_env }}
-      pip install .${{ parameters.pip_opts }}
-      pip install -r dev-requirements.txt
+
+      # install reco-utils and needed dependencies
```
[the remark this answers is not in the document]
miguelgfierro: yeah that sounds good

dev-requirements.txt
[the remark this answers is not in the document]
miguelgfierro: sounds good I'll make the changes

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
anargyri: Even now, a new VM has newer Spark versions than 2.4.8. So it won't work without downloading the binaries. See [this way of setting up the VM](https://github.com/microsoft/recommenders/issues/1408#issuecomment-848078262).

tests/ci/azure_pipeline_test/release_pipeline.yml
anargyri: This is the recommended practice. See https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives: "The tar.gz file is a source archive whereas the .whl file is a built distribution. Newer pip versions preferentially install built distributions, but will fall back to source archives if needed. You should always upload a source archive and provide built archives for the platforms your project is compatible with."

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
gramhagen: i pretty sure these tests work without requiring binaries, we should test it out before worrying too much about managing the version

pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[build-system]
+requires = [
+    "setuptools>=52",
+        "wheel>=0.36"
+	]
+	build-backend = "setuptools.build_meta"
```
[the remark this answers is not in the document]
anargyri: Fixed it.

tests/ci/azure_pipeline_test/release_pipeline.yml
gramhagen: the tarball mentioned in that doc is already created by the `python setup.py sdist bdist_wheel` command and resides in the dist/ folder next to the wheel. Creating additional archives of the source and example code is unnecessary since we are tagging the release

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
anargyri: We have already tested this, they don't work out of the box on a VM.

gramhagen: I just spun up a new Ubuntu 18.04 DSVM and it works fine with out installing any binaries.
Here were the steps I took (the key is to unset SPARK_HOME so pip installed spark files are used instead of the pre-existing files on the dsvm):
```
git clone https://github.com/microsoft/recommenders
cd recommenders
conda create -n reco python=3.6 -y
conda activate reco
pip install -e .[spark]
pip install pytest
sudo apt install -y openjdk-8-jdk
export JAVA_HOME="/usr/lib/jvm/java-1.8.0-openjdk-amd64"
export PYSPARK_PYTHON=`which python`
export PYSPARK_DRIVER_PYTHON=`which python`
unset SPARK_HOME
pytest tests/unit/reco_utils -m "spark and not notebooks and not gpu"
```

miguelgfierro: good catch, I'll review this on another PR

tests/ci/azure_pipeline_test/release_pipeline.yml
miguelgfierro: I think we are talking here about two different things, the tar.gz generated by `python setup.py ...` is the package, as Andreas mentioned. It has the same content as the wheel. The tar.gz is a copy of the code at the time of the release, the same as the zip, but with better compression. 

Since the good practice is to upload the wheel and the package tar.gz, we should do that. Also with the code, we can have just the zip for simplicity.

> Creating additional archives of the source and example code is unnecessary since we are tagging the release

I would argue that this improves the user experience, it seems easier for the user to download just the utilities and examples from the zip directly than to the tag, download all the repository and then grab the part it is needed. All this is inspired in LightGBM.

Does it make sense?

gramhagen: I don't really see the use case for this. Pulling the auto generated archive from the tag seems like it would be easy enough and pretty standard. It's not that much extra effort on our end but without a specific scenario driving it to me it seems like overkill. Up to you if you want to keep it.

miguelgfierro: ok, maybe I'm wrong and it is easy to download the tag. @anargyri what's your view on this?

anargyri: So, if I understand the discussion, we all agree that the tarball and wheel should be in the archives, So the question is about the zip archive with the source code. I guess we could skip it because people can click on the tagged commit and download it from there as Scott said. It's one extra click more or less. I am a bit curious why LightGBM have it, maybe we could ask them.

miguelgfierro: ok sounds good, I'll remove that part then

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
anargyri: You are right, unsetting SPARK_HOME seems to work.

tests/ci/azure_pipeline_test/release_pipeline.yml
miguelgfierro: PR here: https://github.com/microsoft/recommenders/pull/1442

tests/ci/azure_pipeline_test/dsvm_linux_template.yml
miguelgfierro: PR here: https://github.com/microsoft/recommenders/pull/1444

dev-requirements.txt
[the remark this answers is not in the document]
miguelgfierro: Pr here: https://github.com/microsoft/recommenders/pull/1445

PR #1429, opened 2021-06-04
Update the README files for the pip packaging

anargyri:
### Description

Replace conda installation with pip installation in the README. Edit the README shown on pypi.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ For a more detailed overview of the repository, please see the documents on the 
 
 Please see the [setup guide](SETUP.md) for more details on setting up your machine locally, on a [data science virtual machine (DSVM)](https://azure.microsoft.com/en-gb/services/virtual-machines/data-science-virtual-machines/) or on [Azure Databricks](SETUP.md#setup-guide-for-azure-databricks).
 
-To setup on your local machine:
+The installation of the recommenders package has been tested with Python versions 3.6 and 3.7. It is recommended to install the package and its dependencies inside a clean environment (such as [conda](https://docs.conda.io/projects/conda/en/latest/glossary.html?highlight=environment#conda-environment) or [venv](https://docs.python.org/3/library/venv.html)).
+
+To set up on your local machine:
 
 To install core utilities, CPU-based algorithms, and dependencies
 
--- a/README.md
+++ b/README.md
@@ -32,27 +32,26 @@ Please see the [setup guide](SETUP.md) for more details on setting up your machi
 
 To setup on your local machine:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -32,27 +32,26 @@ Please see the [setup guide](SETUP.md) for more details on setting up your machi
 
 To setup on your local machine:
 
-1. Install Anaconda with Python >= 3.6. [Miniconda](https://conda.io/miniconda.html) is a quick way to get started.
-
-2. Clone the repository
+To install core utilities, CPU-based algorithms, and dependencies
```
gramhagen, line 35: we should mention here which versions of python can be used, and ideally suggest using Anaconda or virtual env to provide isolation of dependencies, maybe point to some external documentation on how to set either of those up.

gramhagen (changes requested):
looks good, just one suggestion on pre-reqs.

README.md
```diff
@@ -32,27 +32,26 @@ Please see the [setup guide](SETUP.md) for more details on setting up your machi
...
 
+1. Ensure software required for compilation is installed. On Linux this can be supported by adding build-essential dependencies:
 ```bash
-git clone https://github.com/Microsoft/Recommenders
+sudo apt-get install -y build-essential
 ```
+On Windows you will need [Microsoft C++ Build Tools](https://visualstudio.microsoft.com/visual-cpp-build-tools/)
+
```
miguelgfierro, line 42: A question about this, build-essential is only needed in xlearn or is it needed also for the core deps?

miguelgfierro (approved):

README.md
anargyri: What I did here is I just copied the text from the setup guide. It was introduced here https://github.com/microsoft/recommenders/blob/b170c8459d91480d11ec450eb5e99e4f321af182/SETUP.md 
@gramhagen could confirm but this is not only about xlearn but it's more general. These are the build tools and libraries (like gcc, g++) that are needed for some pip packages to install. Normally they should be there (they are in the DSVM for example), but it's good to be up to date just in case. 
xlearn has the additional dpeendency on cmake. Cmake is not included in build-essential; it depends on it though.

README.md
anargyri: Good points! It would be nice if we had a complete setup for the virtual env in a future PR but this is going to be a bit tricky.

README.md
gramhagen: yeah, this is required by many of the libraries, not just xlearn. if you install pip through `apt-get install python3-pip` it will install `build-essential` by default since it is so common. but other methods of installing pip don't do that, so it's important to provide this step.

PR #1431, opened 2021-06-07
Rearrange README and SETUP files; add info about pip install 

anargyri:
### Description

- Separated the part about conda and yaml files from the setup requirements into a new Markdown file.
- Replaced references to conda envs with more general envs.
- Added references to pip installation in the setup guide.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ sudo apt-get install -y build-essential
 On Windows you will need [Microsoft C++ Build Tools](https://visualstudio.microsoft.com/visual-cpp-build-tools/)
   
 
-2. Install the package from pypi.org:
+2. Install the package from [PyPI](https://pypi.org):
 
 ```bash
 pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


miguelgfierro:
We have right now 3 places with dependencies, the old conda file, the setup.py and docs/requirements.txt. And also, two ways of installing, with pip and with conda. 

Maybe we can consider the advantages/disadvantages of having this versus having 1 single place where we have all deps and only installation with pip. Do you guys think this is possible?

gramhagen:
yeah, i agree it will be best if we can consolidate around using pip install for dependencies and remove the conda setup, should we capture this in a new issue?

anargyri:
> yeah, i agree it will be best if we can consolidate around using pip install for dependencies and remove the conda setup, should we capture this in a new issue?

Yes, feel free to create an issue around this. Let's leave it though for after publication on PyPI.

miguelgfierro (approved):
merging this, we'll review in the future if needed

PR #1449, opened 2021-06-15
Update description

miguelgfierro:
### Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,7 @@ extras_require["experimental"] = [
 setup(
     name=name,
     version=version,
-    description="Microsoft Recommenders, a library of best practices and utilities for building recommendation systems",
+    description="Microsoft Recommenders - Python utilities for building recommender systems",
     long_description=LONG_DESCRIPTION,
     long_description_content_type="text/markdown",
     url="https://github.com/microsoft/recommenders",


anargyri (approved):

wutaomsft:
We should add Summary which is a mandatory field based on https://www.python.org/dev/peps/pep-0314

anargyri:
> We should take a look at https://packaging.python.org/specifications/core-metadata/ too, which include "Summary".

I don't think these specs map exactly to the `setup()` keywords, for example "summary" is missing, see https://setuptools.readthedocs.io/en/latest/references/keywords.html

miguelgfierro:
I was looking at the summary field, and according to [this](https://github.com/microsoft/malmo/blob/master/scripts/python-wheel/package/setup.py#L47), the [summary field in the metadata](https://packaging.python.org/specifications/core-metadata/#summary) corresponds to the description in setup.py. And the [description field in the metadata](https://packaging.python.org/specifications/core-metadata/#description) corresponds with long_description in setup.py. 

Right now we have the description in setup.py, however, we don't have a long summary. Not sure if people want to add that

anargyri:
> I was looking at the summary field, and according to [this](https://github.com/microsoft/malmo/blob/master/scripts/python-wheel/package/setup.py#L47), the [summary field in the metadata](https://packaging.python.org/specifications/core-metadata/#summary) corresponds to the description in setup.py. And the [description field in the metadata](https://packaging.python.org/specifications/core-metadata/#description) corresponds with long_description in setup.py.
> 
> Right now we have the description in setup.py, however, we don't have a long summary. Not sure if people want to add that

No, we have both description and long_description in `setup.py`.

miguelgfierro:
that's right, so then I think we are ok with this

Issue #1390, opened 2021-04-30
[FEATURE] Rename the folder names under reco_utils 


anargyri:
### Description

It would be better to have more descriptive names for the folders under "reco_utils". Specifically replace
- common --> utils
- dataset --> datasets
- recommender --> models

PR #1470, opened 2021-07-07
Improvements on diversity, novelty etc. metrics

anargyri:
### Description

Made improvements on diversity metrics for Spark in GitHub repo:
 - Moved diversity metrics into the same file as other Spark metrics
 - Updated READMEs
 - Added info in the notebook explaining which definitions are used, referring to the relevant citations
 - Added unit test for the notebook
 - Improved the code e.g. replaced cross join, changed novelty definitions to conform with Castells et al. 

### Checklist:


examples/03_evaluate/README.md
```diff
@@ -20,6 +21,11 @@ Two approaches for evaluating model performance are demonstrated along with thei
 3. Classification metrics: These are used to evaluate binary labels
     * Arear Under Curver (AUC) - integral area under the receiver operating characteristic curve
     * Logistic loss (Logloss) - the negative log-likelihood of the true labels given the predictions of a classifier
+4. Non accuracy based metrics: These do not require any ground truth but instead evaluate the following properties of the recommendations
```
gramhagen, line 24: doesn't serendipity still use ground truth, i guess it's optional but alternatively you could just say:

`4. Non accuracy based metrics: These do not compare predictions against ground truth but instead evaluate the following properties of the recommendations`

reco_utils/evaluation/spark_evaluation.py
```diff
@@ -470,3 +479,457 @@ def _get_relevant_items_by_timestamp(
     )
 
     return items_for_user
+
+
+class SparkDiversityEvaluation:
```
gramhagen, line 484: why is this class in the same module as other classes? imo we should be moving towards modules that have a single class in them instead of making larger modules with multiple classes.

examples/03_evaluate/README.md
anargyri: Good point. I made this change.

reco_utils/evaluation/spark_evaluation.py
anargyri: I think that whichever way we choose it has to be consistent i.e. either there are separate modules for rating, ranking and diversity, or one module containing all of them. We currently follow the latter approach, so in this PR I am making the code consistent with this. We could discuss whether to change approach, but that would mean reorganizing all the evaluation code and would go beyond the scope of this PR.

gramhagen: ok fair enough, they're all related so it's not that bad, i think my gut reaction is just that the file is very long now. I'll put in a feature to capture discussion around this.

gramhagen (approved):

examples/03_evaluate/README.md
```diff
@@ -20,6 +21,11 @@ Two approaches for evaluating model performance are demonstrated along with thei
 3. Classification metrics: These are used to evaluate binary labels
     * Arear Under Curver (AUC) - integral area under the receiver operating characteristic curve
     * Logistic loss (Logloss) - the negative log-likelihood of the true labels given the predictions of a classifier
+4. Non accuracy based metrics: These do not compare predictions against ground truth but instead evaluate the following properties of the recommendations
+    * Novelty - measures of how different an item is compared to the items that users have already known or interacted with
```
YanZhangADS, line 25: * Novelty - measures of how novel recommendation items are by calculating their recommendation frequency among users who have not interacted with these items  
 * Diversity - measures of how different items in a recommendation set are with respect to each other
 * Serendipity - measures of how surprising recommendations are to a specific user by comparing them to the items that the user have already interacted with 
 * Coverage - measures related to the distribution of items recommended by the system

YanZhangADS: Novelty is calculated for each item respect to all users; Serendipity is calculated for each item respect to a specific user. These two concepts tend to get confused.

reco_utils/evaluation/spark_evaluation.py
YanZhangADS: I agrees with Scott. If we want to make the code consistent at this point, I would suggest to break down spark_evaluation.py into spark_rating_evaluation.py (SparkRatingEvaluation class) and spark_ranking_evaluation.py (SparkRankingEvaluation class).

reco_utils/evaluation/spark_evaluation.py
```diff
@@ -470,3 +479,457 @@ def _get_relevant_items_by_timestamp(
...
+        :Citation:
+
+            P. Castells, S. Vargas, and J. Wang, Novelty and diversity metrics for recommender systems:
+            choice, discovery and relevance, ECIR 2011
+
+        Returns:
+            pyspark.sql.dataframe.DataFrame: A dataframe with the following columns: col_item, item_novelty.
+        """
+        if self.df_item_novelty is None:
+            n_records = self.train_df.count()
```
gramhagen, line 709: p(i) as defined here is (# times item i is recommended across all users) / (# all interactions in training data)
i would expect p(i) to be more like (# times item i is recommended over all users) / (# all recommendations)

reco_utils/evaluation/spark_evaluation.py
```diff
@@ -470,3 +479,457 @@ def _get_relevant_items_by_timestamp(
...
+            self.df_item_novelty = (
+                self.train_df.groupBy(self.col_item)
+                .count()
+                .withColumn("item_novelty", -F.log2(F.col("count") / n_records))
+                .select(self.col_item, "item_novelty")
+                .orderBy(self.col_item)
+            )
+        return self.df_item_novelty
+
+    def user_novelty(self):
```
gramhagen, line 719: i'm not sure using the ground model definition in this paper is appropriate here, my interpretation is that is useful to estimate what a diversity-focused model should provide, whereas here we already have a model that has generated recommendations so there's no need to estimate probabilities as they can be calculated explicitly from the set of predictions.

examples/03_evaluate/README.md
anargyri: Good suggestions, I changed these.

reco_utils/evaluation/spark_evaluation.py
anargyri: No, actually novelty of each item (`item_novelty`) is calculated on the historical data (`train_df`). Novelty of the recommendations (`novelty`) is restricted to those items in `reco_df`. 
I follow Castells et al. Item novelty is based on p(i) from eq. (1). As they say,p(i) represents the probability that ݅i is observed. It is basically the popularity of the item, so it has to come from data about user choice, not from what you recommend to them (after all you may recommend items that are not popular to them).

reco_utils/evaluation/spark_evaluation.py
anargyri: I am not sure I understand your point, but like item novelty above, user-relative novelty is calculated on historical interactions. It is basically item novelty conditioned on a specific user. So everything they have in section 4 is measured on historical data. In section 5 however they introduce the recommendations, because they define the overall novelty on the recommendations (which corresponds to `novelty` in the code).

reco_utils/evaluation/spark_evaluation.py
gramhagen: do we have separate metrics for novelty assessed on historical data versus recommendations from a model? We should make this distinction clear to users, because I would expect evaluation metrics to apply to the model not the training data, just like the rating and ranking metrics.

reco_utils/evaluation/spark_evaluation.py
gramhagen: see comment above, so far we've been looking at the problem from the perspective of evaluating the recommendations from a model not the historical interactions. if we use this approach comparing results of ALS vs a random recommendation algorithm will be equal, which seems unintuitive to me.

reco_utils/evaluation/spark_evaluation.py
anargyri: So, the novelty metric depends on the recommendations and hence the model. See https://github.com/microsoft/recommenders/blob/a66d452531e503cef9a5d495cb941058b5654ead/reco_utils/evaluation/spark_evaluation.py#L789
At the notebook, the novelties of ALS and random are not the same.

reco_utils/evaluation/spark_evaluation.py
anargyri: They are not equal, see above. What happens that may be a bit confusing is that there is 
- novelty per item (whether plain or user-relative)
- aggregate novelty over the recommendations. 
The former depends only on the historical data, the latter on both historical data and recommendations.

reco_utils/evaluation/spark_evaluation.py
gramhagen: ok, i get it now. thanks

reco_utils/evaluation/spark_evaluation.py
gramhagen: where is this used? should there be a flag in novelty to use item or user-relative novelties?

anargyri: There are no flags. There are three class methods:
- `item_novelty` (i.e. novelty per item, across all users)
- `user_novelty` (novelty per item, user-relative)
- `novelty` (aggregate novelty over a set of recommendations).

See the docstrings about which definitions in the paper they implement. The "headline" metric shown in the notebook is `novelty` (but its computation requires `item_novelty`).

gramhagen: I get that internally that item novelty (assessed on historical data) is used in calculating novelty. But do we expect users to access the raw item_novelty or user_novelty results? Is there an aggregate form of user_novelty based on recommendations that is useful?

we might want to make it clear they do not use the recommendations, by naming them historical_item_novelty or something similar.

anargyri: `user_novelty` is not referenced anywhere, but Yan had already written a version, so I made it consistent with the paper.

anargyri: Yes, good suggestion, I will rename them. It may be useful to have access to the novelty of a certain item outside recommendations e.g. for data pre or post processing or analytics, although it's probably not a frequent scenario.
There is an aggregate version of user novelty in the definition of `m(R|u)` in section 5 of the paper. Maybe it is good to add it too.

gramhagen: I agree that would be nice to add

gramhagen (approved):

tests/unit/reco_utils/evaluation/test_spark_evaluation.py
```diff
@@ -308,3 +373,119 @@ def test_spark_python_match(python_data, spark):
...
+@pytest.mark.spark
+def test_user_novelty(spark_diversity_data, target_metrics):
+    train_df, reco_df = spark_diversity_data
+    evaluator = SparkDiversityEvaluation(
+        train_df=train_df, reco_df=reco_df, col_user="UserId", col_item="ItemId"
+    )
+    actual = evaluator.historical_user_novelty().toPandas()
+    assert_frame_equal(
+        target_metrics["user_novelty"], actual, check_exact=False, check_less_precise=4
+    )
```
miguelgfierro, line 419: I think it would be good to check the limits of all these metrics. For example in [mae](https://github.com/microsoft/recommenders/blob/main/tests/unit/reco_utils/evaluation/test_python_evaluation.py#L161) we have that if the prediction and true vectors are the same, then mae is 0. I guess that if all the recommendations are the same, there is no novelty.

reco_utils/evaluation/spark_evaluation.py
```diff
@@ -470,3 +479,457 @@ def _get_relevant_items_by_timestamp(
...
+                    -F.log2(F.col("item_prob") / F.col("sum(item_prob)")),
+                )
+                .select(self.col_user, F.col("i1").alias(self.col_item), "user_novelty")
+                .orderBy([self.col_user, self.col_item])
+            )
+
+        return self.df_user_novelty
+
+    def novelty(self):
+        """Calculate average novelty for recommended items. Follows section 5 from
```
miguelgfierro, line 775: Maybe clarify here that it is the novelty in a list

```
"""Calculate the average novelty in a list of recommended items (this assumes that the recommendation list is already computed). Follows section 5 from ....
```

miguelgfierro (approved):

tests/unit/reco_utils/evaluation/test_spark_evaluation.py
anargyri: Good point. For some of these there are no upper limits e.g. novelty can approach infinity.

reco_utils/evaluation/spark_evaluation.py
anargyri: What do you mean? It says it's the novelty of a list of recommendations.

reco_utils/evaluation/spark_evaluation.py
```diff
@@ -470,3 +479,457 @@ def _get_relevant_items_by_timestamp(
...
+            self.df_diversity = self.df_user_diversity.agg(
+                {"user_diversity": "mean"}
+            ).first()[0]
+        return self.df_diversity
+
+    # Novelty metrics
+    def historical_item_novelty(self):
+        """Calculate novelty for each item. Novelty is computed as the minus logarithm of
+        (number of interactions with item / total number of interactions). The definition of the metric
+        is based on the following reference using the choice model (eqs. 1 and 6):
```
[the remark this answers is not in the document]
anargyri: \frac doesn't seem to work in Sphinx. Maybe some change in the Sphinx configuration is needed.

YanZhangADS (approved):
Below metrics are based on paper Castells et. al. Novelty and diversity metrics for recommender systems: choice, discovery and relevance, ECIR 2011.

historical-item-novelty: an item is less frequently being interacted with in the training data (Eq. 1, Eq. 6)
novelty: "novelty" is built on top of "historical-item-novelty". It is weighted average historical-item-novelty for all recommended items in the whole reco system, not respect to one particular user (Eq. 10)

reco_utils/evaluation/spark_evaluation.py
```diff
@@ -470,3 +479,457 @@ def _get_relevant_items_by_timestamp(
...
+        """
+        if self.df_diversity is None:
+            self.df_user_diversity = self.user_diversity()
+            self.df_diversity = self.df_user_diversity.agg(
+                {"user_diversity": "mean"}
+            ).first()[0]
+        return self.df_diversity
+
+    # Novelty metrics
+    def item_novelty(self):
```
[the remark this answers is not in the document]
anargyri: Now I have removed user novelty, equation (7) is not used anywhere in the code.

The definition of item novelty you had came from the blog post above. I changed this definition to conform with the Castells et al. paper. I suggest we stick to the paper instead of a blog post for several reasons, including that papers in the literature follow the same approach which is based on a probabilistic model. Moreover, the papers have been peer reviewed whereas a blog post reflects just one person's ideas. 

What happened was that the blogger changed the standard definition because he misunderstands these models in the literature. He proposes his own definition but it is incorrect and doesn't correspond to any actual probability. It is basically justified on rough intuition.

miguelgfierro:
@YanZhangADS @anargyri feel free to merge when you consider the work is finished. After this is merged, I'll work on https://github.com/microsoft/recommenders/issues/1390

anargyri:
It looks like GitHub does not render the formulas properly on the branch (this is true for other notebooks too). You can see them all right on VSCode though.

miguelgfierro:
@anargyri I'm able to render the formulas in reviewnb, however, this one is not rendering:

```
\textrm{IL} = \frac{1}{|M|} \sum_{u \in M} \frac{1}{\binom{N_r(u)}{2}} \sum_{i,j \in N_r (u),\, i
```
at first sight, there is a } missing

anargyri:
> @anargyri I'm able to render the formulas in reviewnb, however, this one is not rendering:
> 
> ```
> \textrm{IL} = \frac{1}{|M|} \sum_{u \in M} \frac{1}{\binom{N_r(u)}{2}} \sum_{i,j \in N_r (u),\, i
> ```
> 
> at first sight, there is a } missing

This looks like a ReviewNB bug. It works for me on VSCode.

PR #1481, opened 2021-07-17
Create pypi-test-publish.yml

dciborow:
### Checklist:

### DIFF
--- a/.github/workflows/pypi-test-publish.yml
+++ b/.github/workflows/pypi-test-publish.yml
@@ -1,0 +1,25 @@
+name: Test Upload Python Package
+
+on:
+  workflow_dispatch
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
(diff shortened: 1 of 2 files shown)


gramhagen:
how can we run the test suite before kicking this off, or better yet, using the test-pypi package?

miguelgfierro:
Right now we are using this pipeline to generate the wheels: https://github.com/microsoft/recommenders/blob/main/tests/ci/azure_pipeline_test/release_pipeline.yml. The package wheel is tested and then manually uploaded to pypi. The reason why we are doing it manually is by following the advice on how Guolin does it in LightGBM, because he says it is a good idea to have manual control on this process, since we can't erase a version once it is uploaded, and also he mentioned that they don't make this action frequently.

Is this pipeline trying to replace the manual process? @dciborow can you clarify its use?

dciborow:
@miguelgfierro , i made an update so that neither pipeline will trigger automatically, and that both would only run on manual run.
But yes, this will automate the publishing to pypi. The first build tests the release to the test pypi. The 2nd targets production pypi. 

Bringing this back up to minimize the manual steps required to get an update to date release out. We want to be able to move quickly when their is a security or dependancy update that needs to be released.

miguelgfierro (approved):

Issue #1498, opened 2021-08-16
Add gpu build step to Github Action CI


laserprec:
### Description
This is one of the sequence of changes to migrate CI infrastructures onto GitHub Actions:

Adding self-hosted agent to run the GPU tests.

PR #1503, opened 2021-08-19
Add nightly build

laserprec:
### Description
To improve developer experience working with this repo, we will begin migrating our DevOps pipelines into Github Action, leveraging popular DevOps tools like `tox` and `flake8`. This will be a sequence of updates to our DevOps infrastructures:

1. ~~Propose and draft the CI pipeline in Github Action~~
2. ~~Setup self-hosted machines to run our GPU workloads~~
3. Create feature parity with the existing CI pipelines (pr-gates & nightly-build)  **<---- (Current PR)**
4. Run tests on the appropriate dependency subsets
5. Optimize build time for unit tests
6. Enforce flake8 (coding style checks) on the build and clean up coding styles to pass flake8 checks
7. Deprecate CI pipelines in ADO and switch to Github Actions

#### In this PR:

Add a nightly build pipeline that runs the full test suites: unit, integration and smoke tests:
image

### Related Issues
#1498 

### Checklist:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -19,7 +19,7 @@ on:
   push:
     # because we can't schedule run for non-main branches,
     # to ensure we are running the build on the staging branch, we can add push policy for it
-    branches: [staging]
+    branches: [staging, laserprec/nightly-ci]
 
   # enable manual trigger
   workflow_dispatch:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,6 +2,31 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # ---------------------------------------------------------
+
+# NOTE:
+# This file defines following CI workflow:
+#  ┌──────────┐   ┌─────────┐   ┌────────┐
+#  │  static  ├─┬─►  build* ├─┬─►  test  │
+#  │ analysis │ │ │  (cpu)  │ │ │ report │
+#  └──────────┘ │ └─────────┘ │ └────────┘         
+#               │ ┌─────────┐ │
+#               ├─►  build* ├─┤
+#               │ │ (spark) │ │
+#               │ └─────────┘ │
+#               │ ┌─────────┐ │
+#               └─►  build* ├─┘
+#                 │  (gpu)  │
+#                 └─────────┘
+#                    ....
+# *each runs in PARALLEL the different combinations
+#  of python version, OS, test subsets, etc
+#
+# There are 3 compute environments our library is supporting: CPU, GPU and Spark
+# This pipeline's goal is to ensure we run and pass our tests in these supported compute
+# For all of the jobs in GPU build, we are using self-host VMs to run them.
+# For the rest of the jobs, with the exception of integration tests (#1507), we will use default agents provided by GitHub 
+#
+# ASCII chart created via https://asciiflow.com/
 name: nightly
 
 on:
@@ -28,26 +53,6 @@ on:
         description: 'Test scenario tags'
         default: 'Anything to describe this manual run (optional)'
     
-
-# This file defines following CI workflow:
-#
-#  ┌──────────┐   ┌─────────┐   ┌────────┐
-#  │  static  ├─┬─►  build* ├─┬─►  test  │
-#  │ analysis │ │ │  (cpu)  │ │ │ report │
-#  └──────────┘ │ └─────────┘ │ └────────┘         
-#               │ ┌─────────┐ │
-#               ├─►  build* ├─┤
-#               │ │ (spark) │ │
-#               │ └─────────┘ │
-#               │ ┌─────────┐ │
-#               └─►  build* ├─┘
-#                 │  (gpu)  │
-#                 └─────────┘
-#                    ....
-# *each runs in PARALLEL the different combinations
-#  of python version, OS, test subsets, etc
-#
-#  ASCII chart created via https://asciiflow.com/
 
 jobs:
 ###############################################
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,210 @@
+# ---------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# ---------------------------------------------------------
+name: nightly
+
+on:
+  #            ┌───────────── minute (0 - 59)
+  #            │ ┌───────────── hour (0 - 23)
+  #            │ │ ┌───────────── day of the month (1 - 31)
+  #            │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+  #            │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)                                  
+  #            │ │ │ │ │
+  #            │ │ │ │ │
+  # schedule: "0 0 * * *"  # basically running everyday at 12AM
+  #   branches: [ staging, main ]
+
+  # development triggers, can be removed afterwards.
+  push:
+    branches: [ laserprec/nightly-ci ]
+
+# This file defines following CI workflow:
+#
+#  ┌──────────┐   ┌─────────┐   ┌────────┐
+#  │  static  ├─┬─►  build* ├─┬─►  test  │
+#  │ analysis │ │ │  (cpu)  │ │ │ report │
+#  └──────────┘ │ └─────────┘ │ └────────┘         
+#               │ ┌─────────┐ │
+#               ├─►  build* ├─┤
+#               │ │ (spark) │ │
+#               │ └─────────┘ │
+#               │ ┌─────────┐ │
+#               └─►  build* ├─┘
+#                 │  (gpu)  │
+#                 └─────────┘
+#                    ....
+# *each runs in PARALLEL the different combinations
+#  of python version, OS, test subsets, etc
+#
+#  ASCII chart created via https://asciiflow.com/
+
+jobs:
+###############################################
+############### STATIC-ANALYSIS ###############
+###############################################
+  static-analysis:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Python 3.6
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6
+    
+    - name: Install dependencies (tox)
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install tox
+
+    - name: Run flake8
+      # TODO: re-enable this flake8 block (turned off to get a draft of the pipeline infrastructure)
+      continue-on-error: true
+      run: |
+       tox -e flake8
+
+###############################################
+################## CPU-BUILD ##################
+###############################################
+  build-cpu:
+    runs-on: ${{ matrix.os }}
+    needs: static-analysis
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python: [3.6]
+        # different kind of tests are located in tests/<unit|integration|smoke> folders
+        test-kind: ['unit', 'smoke', 'integration'] 
+        # pytest markers configured in tox.ini. See https://docs.pytest.org/en/6.2.x/example/markers.html
+        test-marker: ['not spark and not gpu']
+
+    steps:
+    - uses: actions/checkout@v2
+    ################# Run Python tests #################
+    - name: Use Python ${{ matrix.python }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    # There are little documentation on how to call **local** actions
+    # https://docs.github.com/en/actions/creating-actions/about-actions#choosing-a-location-for-your-action
+    # but there is some working insights from this discussion:
+    # https://github.community/t/path-to-action-in-the-same-repository-as-workflow/16952/2
+    - name: Run ${{ matrix.test-kind }} tests ('${{ matrix.test-marker }}')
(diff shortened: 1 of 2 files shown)


gramhagen (approved):

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,229 @@
...
+        python: [3.6]
+        # different kind of tests are located in tests/<unit|integration|smoke> folders
+        test-kind: ['unit', 'smoke']
+        # pytest markers configured in tox.ini. See https://docs.pytest.org/en/6.2.x/example/markers.html
+        test-marker: ['not spark and not gpu']
+        include:
+          # Integration tests needs a more powermachine with more memory. GitHub-hosted agents only have 7GB memory.
+          - os: self-hosted
+            python: 3.6
+            test-kind: 'integration'
```
miguelgfierro, line 89: one question about this, is this pipeline running unit, smoke and integration test?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,229 @@
...
+###############################################
+################## CPU-BUILD ##################
+###############################################
+  build-cpu:
+    runs-on: ${{ matrix.os }}
+    needs: static-analysis
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python: [3.6]
```
miguelgfierro, line 80: let's say we wanted to add python 3.7 in parallel, how will that look like?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,229 @@
...
+################# GPU-BUILD #################
+###############################################
+  build-gpu:
+    runs-on: [self-hosted, Linux, gpu] # this is a union of labels to select specific self-hosted machine
+    needs: static-analysis
+    strategy:
+      matrix:
+        python: [3.6]
+        # different kind of tests are located in tests/<unit|integration|smoke> folders
+        test-kind: ['unit', 'smoke', 'integration']
```
miguelgfierro, line 178: in the other builds we have only `test-kind: ['unit', 'smoke']`, what is the difference?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,229 @@
...
+    - cron: '0 0 * * *'  # basically running everyday at 12AM
+    # cron works with default branch (main) only: # https://github.community/t/on-schedule-per-branch/17525/2
+  
+  push:
+    branches:
+      # Because we can't schedule run for non-main branches,
+      # to ensure we are running the build on the staging branch, we can add push policy for it
+      - staging
+      # Development triggers, can be removed afterwards
+      - laserprec/nightly-ci
```
miguelgfierro, line 25: do we need to keep this until it is merged? or can you execute the pipeline manually?

.github/workflows/nightly.yml
laserprec: we can just expand the python array to:

```
matrix:
   os: [ubuntu-latest]
   python: [3,6,, 3.7]
```

.github/workflows/nightly.yml
laserprec: Yup! I know this extra `include` clause is looking unusual. At the moment, the integration tests cannot run on the smaller SKUs (7GB) provided by GitHub because there is enough memory to run the training on the full dataset. 

Ideally, all our tests shouldn't be that compute intensive and they can be ran on the default SKUs provided by the CI (unless we need to test for compatibility with GPU enabled machines). Thus the matrix should ideally look like:

```yaml
   test-kind: ['unit', 'smoke', 'integration']
   test-marker: [....]
   ....
```
I think we can have a discussion over the objective of the different unit/smoke/integration tests and whether doing training on the full dataset achieve those goals. Added a new issue for the discussion: #1507.

.github/workflows/nightly.yml
laserprec: Nope. In fact I can add a manual trigger.

.github/workflows/nightly.yml
laserprec: This difference is due to how we need to make an exception for running integrations in self-hosted machine. 

Here in the GPU-build, there aren't problem running all of the 3 different test-kinds in the self-hosted GPU machine, so there's no `include` clause in the matrix. In the SPARK-build, some of the the integration & unit tests will fail running on the github-hosted machines (i.e. `os: [ubuntu-latest]`). So I added a `include` clause to make rooms for such exception.

.github/workflows/nightly.yml
laserprec: In fact, it turns out that the manual trigger button wouldn't show up in the UI until these changes are made into the repo's default branch (master) https://github.community/t/cant-trigger-workflow-manually/121740/27. 

I will re-activate the push triggers for testing purpose and delete before merge.

.github/workflows/nightly.yml
miguelgfierro: sorry I don't understand, you mentioned that this pipeline runs unit, smoke and integration but later that the integration cannot run on the smaller SKUs. Some questions:
1) Where are the integration tests running? 
2) Is there a UI where I can see an example on how the tests are running?
3) there are 3 tests running for CPU, unit, smoke and integration, are they running in parallel or sequentially?
4) Do we know if GitHub has plans to support bigger SKUs?
5) Related to 4, Azure Pipelines also has SKUs for testing, are they the same as GitHub, or are they bigger? (if they are bigger, then GitHub might catch up eventually)

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,229 @@
...
+        # different kind of tests are located in tests/<unit|integration|smoke> folders
+        test-kind: ['unit', 'smoke'] 
+        # pytest markers configured in tox.ini. See https://docs.pytest.org/en/6.2.x/example/markers.html
+        test-marker: ['spark and not gpu']
+        include:
+          # Integration tests needs a more powermachine with more memory. GitHub-hosted agents only have 7GB memory.
+          - os: self-hosted
+            python: 3.6
+            test-kind: 'integration'
+            test-marker: 'not spark and not gpu'
```
miguelgfierro, line 144: Not sure if this should be `spark and not gpu`? This is the spark build

.github/workflows/nightly.yml
miguelgfierro: do we know when GitHub has plans to add GPU hosted machines?

.github/workflows/nightly.yml
miguelgfierro: can we try to add both 3.6 and 3.7? My guess is that the current code would work with both. I'm assuming that each pipeline will run in parallel, right?

miguelgfierro:
@laserprec do you have the time saving between doing the old pipeline vs this one?

.github/workflows/nightly.yml
anargyri: There is an issue I have observed with python 3.7. When pip installing with the gpu option (on a DSVM) and importing numba I get
numpy.core.multiarray failed to import

.github/workflows/nightly.yml
laserprec: No worries! 

1. The integration tests are currently running in the [self-hosted gpu agent](https://github.com/microsoft/recommenders/settings/actions/runners). This is configured with the `self-hosted` keyword.
  ```
  include:
          - os: self-hosted
  ```
2. Please find all of the GitHub Action runs [here](https://github.com/microsoft/recommenders/actions), especially the `pr-gate` and `nightly` pipeline.
3. Current all of them are running in parallel. However because we currently set up only 2 self-hosted but there are about 5 nightly step needing to run on self-hosted machine, we haven't scale up the resource to run them in true parallel.
4. As far as I know, I don't think there's a plan. Maybe the SKU offerings are different with GitHub Enterprise, but I haven't gotten too much info on that.
5. Azure Pipeline and GitHub Action both uses the same SKU: Standard_DS2_v2.

I hope that helps.

.github/workflows/nightly.yml
laserprec: good catch :D

.github/workflows/nightly.yml
laserprec: Unfortunately, I don't have much information on that. For folks I asked, they don't have visibility in these feature planning cycles. My assumption is that CI workload requiring GPU resources is quite rare and those who do can be refer to self-hosting their GPU resources. There might not be enough incentives in providing managed GPU resources. Also, recently hackers have been [exploiting these small GitHub Action SKUs to mine bitcoins](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/), so security could be big concern if they offer GPU machine.

.github/workflows/nightly.yml
miguelgfierro: Looking at SOF, it seems that this is solved by numpy 1.8 https://stackoverflow.com/a/21946788 , but we have a higher numpy. If this is the only issue we have in python 3.7, it might be worthwhile explore if just a change in the version of numpy or numba fixes the problem

.github/workflows/pr-gate.yml
```diff
@@ -7,9 +7,13 @@ name: pr-gate
 on:
   pull_request:
     branches: [ staging, main ]
```
miguelgfierro, line 9: if this pr-gate.yml is the equivalent of the unit tests we had, could we please follow the same pattern and rename this as unit so it is more clear

.github/workflows/nightly.yml
miguelgfierro: got it thanks for the clarification
1. would you mind explaining in the file that some test are being run on the hosted machine and that some tests on the self-hosted one?
2. For nightly, in the current pipeline we are running only smoke and integration, see https://github.com/microsoft/recommenders/blob/main/tests/ci/azure_pipeline_test/dsvm_nightly_linux_cpu.yml#L30, and these two are sequential. The reason why they are sequential is that the smoke tests are just a small version of the integration tests. Is it possible to replicate this behavior in github?

We are making so many changes in test, I think we need also to review the test readme, there might be info that is not up to date: https://github.com/microsoft/recommenders/blob/main/tests/README.md

.github/workflows/pr-gate.yml
laserprec: > @laserprec do you have the time saving between doing the old pipeline vs this one?

@miguelgfierro, need to fix some spark bugs on the self-hosted machine. Will get back to you later :)

laserprec: > if this pr-gate.yml is the equivalent of the unit tests we had, could we please follow the same pattern and rename this as unit so it is more clear

@miguelgfierro, if I understand correctly, you are suggesting that we can rename `pr-gate.yml` to `unit-test.yml`. Is that right? In that case, I think it would be best we stick with a more functional name describing utility of the pipeline like `gating a PR from merging` because in the future, depending on how we structure tests, we can possibly run smoke/e2e/integration tests in the PR-gate itself.

.github/workflows/nightly.yml
laserprec: Thanks for the great feedbacks! 

1. I already added a comment on these mixed runs for integration tests and link to a new issue (#1507) on whether we should optimize/trim these integration tests :
```yml
 include:
          # Integration tests need a powerful machine with more memory. GitHub-hosted agents only have 7GB memory.
          # TODO: Optimization/refactoring should be done to ensure that these test can run in the default CI agent (#1507)
          - os: self-hosted
```
@miguelgfierro is there anywhere else you would like me to comment?

2. If I understand correctly, in the existing pipeline, the smoke tests are running before integration tests so we can "fail fast" and avoid running the full integration test which is destined to fail. In the similar manner, github action has the `[fail-fast](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)` keyword which will cancel running jobs in the matrix if any failed.  If this isn't what you are going after and we need to run both of them sequentially, I think we would need to refactor the CI structure because is designed to run everything in parallel as much as possible.

> We are making so many changes in test, I think we need also to review the test readme
This is a great document when I was onboarding :). All of the CI changes is compatible with the what's written in the document in terms of test execution. One thing that we can change is that we no longer need the `--duration 0` flag from 
`pytest tests/integration -m "integration and not spark and not gpu" --durations 0`, because these flags are stored in [tox.ini](https://github.com/microsoft/recommenders/blob/main/tox.ini#L53) and pytest will pick the flags up from tox.ini without you specifying it. Of course, it doesn't hurt to repeat yourself add them in when running. 😃

laserprec:
> @laserprec do you have the time saving between doing the old pipeline vs this one?

@miguelgfierro, please interpret the numbers with a grain of salt. It's a not apples-to-apples comparison:
The current nightly build on ADO takes about **4hrs50min** to run sequentially the 3 builds and smoke & integration tests.

image
[Link to pipeline runs](https://dev.azure.com/best-practices/recommenders/_build)

This new nightly build on GitHub Action is about **4hrs40min** to run the 3 builds and the full test suites (unit/smoke/integration) in partially parallel. 

image
[link to pipeline run](https://github.com/microsoft/recommenders/actions/runs/1163637279)

As you see the GH action is bottlenecked by the longest running thread which is the GPU integration tests (4hr40min) and is considerably longer than the same run in ADO. This is because we are not using the same Azure SKUs to run them. On ADO we are using NC6s_v2 (which has P100 GPU but we no longer have any quota to allocate more) and in GitHub we are using NC12_promo (which has two P40 GPUs).

So at this point, we don't have much immediate gains, but with the new CIs in GitHub Action we are in better grounds to scale up the infra for distributed loads when we have more environments to support and centralized interface to interpret the run and optimize the individual jobs in the next step.

miguelgfierro (approved):
This is really good job @laserprec

PR #1505, opened 2021-08-20
add logic and example of item feature vector based item-item similari…

YanZhangADS:
…ty calculation

### Description

In the scenarios when item features are available, we may want to calculate item-item similarity based on item feature vectors. In this PR, we add logic and example on how to calculate diversity metrics using item feature vector based item-item similarity.

### Checklist:


examples/03_evaluate/als_movielens_diversity_metrics.ipynb
```diff
@@ -774,6 +774,162 @@
     "df_results"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Conclustion\n",
```
gramhagen, line 781:
```suggestion
    "#### Conclusion\n",
```

recommenders/evaluation/spark_evaluation.py
```diff
@@ -560,6 +565,21 @@ def __init__(
...
                 col_user, col_item, F.col(self.col_relevance).cast(DoubleType())
             )
+            
+        if item_feature_df is None:
+            self.item_feature_df = None
+        else:
+            required_schema = StructType(
+               (
+                 StructField(self.col_item, IntegerType()),
+                 StructField("features", VectorUDT()),
```
gramhagen, line 575: can we create self.col_features so the user can control the column name just like user, item, relevance

recommenders/evaluation/spark_evaluation.py
```diff
@@ -630,6 +659,39 @@ def _get_cosine_similarity(self, n_partitions=200):
...
         return self.df_cosine_similarity
 
+    @staticmethod
+    @udf(returnType=DoubleType())
+    def sim_cos(v1,v2):
+        try:
+            p = 2
+            return float(v1.dot(v2))/float(v1.norm(p)*v2.norm(p))
+        except:
+            return 0
```
gramhagen, line 669: do we want to indicate that two vectors are identical if there's a failure to compute the distance? can we return a NaN or something?

gramhagen (changes requested):
few small comments, this looks like a nice addition.

recommenders/evaluation/spark_evaluation.py
YanZhangADS: if we vectors are identical, their cosine similarity will be 1. On which situation do you think the calculation will fail? @gramhagen

gramhagen: sorry you're right, we're not computing a distance here. i think in this case returning a 0 is okay in the exception. my concern is just about silent failures, if for some reason there is an exception, most likely because the vectors are not the correct size or have invalid values, then we don't ever surface the fact that one of the item vectors is incorrectly defined.

YanZhangADS: added logic to check the size of vectors. If the sizes of two input vector are different, raise an exception

examples/03_evaluate/als_movielens_diversity_metrics.ipynb
```diff
@@ -36,13 +36,16 @@
     "We evaluate the results obtained with two approaches: using the ALS recommender algorithm vs. a baseline of random recommendations. \n",
```
miguelgfierro, line 36: <pre class="ql-syntax" spellcheck="false">Line #10.    als_diversity
</pre><p>this is very interesting, Yan do you know why the diversity varies in comparison between the coocurrence method?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1505/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='45'/>

miguelgfierro, line 36: <p>Something it would be nice is to talk about why diversity metrics are important for a customer point of view. Maybe also talk about the long tail items: https://github.com/microsoft/recommenders/blob/main/GLOSSARY.md</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1505/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

YanZhangADS: addressed

YanZhangADS: the calculation of diversity is based on item-item similarity values. It varies when using different item-item similarity calculation approaches (1) based on item co-occurrence count; and (2) based on item feature vectors.

anargyri: <p>I am not sure that items in the tail are always highly profitable, they could be cheap because there is low demand for them. Maybe it'd be better to qualify the statement e..g say "oftentimes" or something to that effect.   </p>

anargyri, line 36: <p>The above warning is strange, because there are 6 columns in the data frame, not 4. Any idea what happened? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1505/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

anargyri, line 36: <pre class="ql-syntax" spellcheck="false">Line #14.        #inputCols=["year_made", "text_features", "genres_features"],
</pre><p>Is this comment needed?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1505/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

anargyri: <p>What is the 1043 in the start of the "features" column? Can we remove it?</p>

anargyri: <p>It's interesting that diversity decreased when using the features, for both ALS and random. Maybe you could mention this.</p>

YanZhangADS: > What is the 1043 in the start of the "features" column? Can we remove it?

1043 is the length of the "features" vector. I added explanation about it.

YanZhangADS: > Line #14.        #inputCols=["year_made", "text_features", "genres_features"],
> Is this comment needed?
> 
> 
> _Reply via [ReviewNB](https://app.reviewnb.com/microsoft/recommenders/pull/1505/discussion/)_

removed

YanZhangADS: > The above warning is strange, because there are 6 columns in the data frame, not 4. Any idea what happened?
> 
> 
> _Reply via [ReviewNB](https://app.reviewnb.com/microsoft/recommenders/pull/1505/discussion/)_

fixed schema definition. Thanks for catching it!

recommenders/evaluation/spark_evaluation.py
```diff
@@ -560,6 +567,20 @@ def __init__(
             self.reco_df = reco_df.select(
                 col_user, col_item, F.col(self.col_relevance).cast(DoubleType())
             )
+            
+        if self.item_feature_df is not None:
+            self.col_item_features = col_item_features
+            required_schema = StructType(
+               (
+                 StructField(self.col_item, IntegerType()),
```
anargyri, line 575: We use double type for the itemID, see line 105.

recommenders/evaluation/spark_evaluation.py
```diff
@@ -630,6 +660,43 @@ def _get_cosine_similarity(self, n_partitions=200):
             )
         return self.df_cosine_similarity
 
+    @staticmethod
+    @udf(returnType=DoubleType())
+    def sim_cos(v1,v2):
+        try:
+            p = 2
+            return float(v1.dot(v2))/float(v1.norm(p)*v2.norm(p))
```
anargyri, line 667: Do you know whether using `pyspark.ml.stat.Correlation.corr()` would be faster or slower?

recommenders/evaluation/spark_evaluation.py
```diff
@@ -630,6 +660,43 @@ def _get_cosine_similarity(self, n_partitions=200):
             )
         return self.df_cosine_similarity
 
+    @staticmethod
+    @udf(returnType=DoubleType())
+    def sim_cos(v1,v2):
```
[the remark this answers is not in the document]
YanZhangADS: > There are a few style issues (missing spaces etc.). Could you use black to reformat this file?

fixed with black

recommenders/evaluation/spark_evaluation.py
gramhagen: i'm a still a bit confused on this part, instead of both catching a particular exception and raising an exception and providing a generic except that silently swallows the exception and returns a 0. I think we just do one or the other. I'm fine with using a generic exception and returning a NaN or not even using a try-except and throwing the exception.

anargyri: It's probably better practice to pass the exception through instead of returning 0, because people interpret 0 as the two vectors being orthogonal. Also passing NaN in the case of an exception is not great because people would expect NaN to be the output when both vectors are zero.

anargyri: Although the latter is not so clear in this case, e.g. if both feature vectors are zero then they may be considered similar.

recommenders/evaluation/spark_evaluation.py
```diff
@@ -588,6 +608,15 @@ def _get_pairwise_items(self, df):
         )
 
     def _get_cosine_similarity(self, n_partitions=200):
+        if self.item_feature_df is None:
```
anargyri, line 611: It would be better to select the method for calculating similarity using an argument instead of whether `item_feature_df` is present or not.
Because if the item features are available in the dataframe, one would need to create a new SparkDiversityEvaluation object if one wishes to compute the co-occurrence based similarities.

recommenders/evaluation/spark_evaluation.py
```diff
@@ -630,6 +659,43 @@ def _get_cosine_similarity(self, n_partitions=200):
...
+                self.item_feature_df.select(
+                    F.col(self.col_item).alias("i1"),
+                    F.col(self.col_item_features).alias("f1"),
+                )
+                .join(
+                    self.item_feature_df.select(
+                        F.col(self.col_item).alias("i2"),
+                        F.col(self.col_item_features).alias("f2"),
+                    ),
+                    (F.col("i1") < F.col("i2")),
```
anargyri, line 691: We had <= in intralist similarity.

recommenders/evaluation/spark_evaluation.py
```diff
@@ -630,6 +659,43 @@ def _get_cosine_similarity(self, n_partitions=200):
...
+                )
+                .select("i1", "i2", self.sim_cos("f1", "f2").alias("sim"))
```
anargyri, line 693: This function uses the udf, whereas `_get_cooccurrence_similarity()` uses SQL operations. At the core, both functions compute the same thing (cosine similarity) but in different ways. It probably makes sense to be consistent between the two i.e. to use the most efficient approach in both. Also the computation could then be factored out to remove duplication.

tests/unit/recommenders/evaluation/test_spark_evaluation.py
```diff
@@ -501,3 +525,35 @@ def test_serendipity(spark_diversity_data, target_metrics):
         col_relevance="Relevance",
     )
     assert target_metrics["serendipity"] == evaluator.serendipity()
+
+
+@pytest.mark.spark
+def test_user_diversity2(spark_diversity_data, target_metrics):
```
anargyri, line 531: I would suggest a more informative name so that it is clear to someone looking at output from pytest that this method uses item features. Similarly for the method below.

examples/03_evaluate/als_movielens_diversity_metrics.ipynb
```diff
@@ -722,85 +732,239 @@
...
+   ],
+   "metadata": {}
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "It's interesting that diversity changes when using different item-item similarity calculation approach, for both ALS algorithm and random recommender. But the diversity of random recommender is still higher than ALS algorithm. "
+   ],
+   "metadata": {}
   },
```
anargyri, line 939: Another thing you could add here is how serendipity changes now that cosine similarity is computed using the item feature vectors.

recommenders/evaluation/spark_evaluation.py
anargyri: +1

recommenders/evaluation/spark_evaluation.py
YanZhangADS: Do you mean we want to only create a single SparkDiversityEvaluation object, and be able to get diversity value from two different item similarity approaches if item features are available? 
e.g.
eval = SparkDiversityEvaluation(args, ...)
diversity_item_coocurrence = eval.diversity(method="item_coocurrence")
diversity_item_similarity=eval.diversity(method="item_similarity")

recommenders/evaluation/spark_evaluation.py
YanZhangADS: addressed

recommenders/evaluation/spark_evaluation.py
YanZhangADS: removed try/catch block

examples/03_evaluate/als_movielens_diversity_metrics.ipynb
YanZhangADS: added discussion in the notebook

recommenders/evaluation/spark_evaluation.py
YanZhangADS: If using this approach, it doubles the code in multiple functions. It looks very confusing. Therefore I kept the previous method unchanged, i.e. only one "item_sim_measure" can be chosen for one object 
eval = SparkDiversityEvaluation(args, ...). 

"select the method for calculating similarity using an argument instead of whether item_feature_df is present or not."  -- this is implemented.

recommenders/evaluation/spark_evaluation.py
YanZhangADS: fixed

tests/unit/recommenders/evaluation/test_spark_evaluation.py
YanZhangADS: modified the method name to "test_user_diversity_item_feature_vector", and the same change made for the method below.

examples/03_evaluate/als_movielens_diversity_metrics.ipynb
YanZhangADS: added calculation for serendipity.

recommenders/evaluation/spark_evaluation.py
YanZhangADS: For function "_get_item_feature_similarity" I have to use a udf. I have not found a good way to combine both functions.

recommenders/evaluation/spark_evaluation.py
```diff
@@ -554,6 +556,7 @@ def __init__(
         self.df_user_diversity = None
         self.avg_diversity = None
         self.item_feature_df = item_feature_df
+        self.item_sim_measure = item_sim_measure
```
anargyri, line 559: Also please add the new argument in the docstring too.

recommenders/evaluation/spark_evaluation.py
```diff
@@ -608,12 +618,12 @@ def _get_pairwise_items(self, df):
...
 
     def _get_cosine_similarity(self, n_partitions=200):
-        if self.item_feature_df is None:
-            # calculate item-item similarity based on item co-occurrence count
-            self._get_cooccurrence_similarity(n_partitions)
-        else:
+        if self.item_sim_measure == "item_feature_vector":
             # calculate item-item similarity based on item feature vectors
             self._get_item_feature_similarity(n_partitions)
+        else:
```
anargyri, line 624: Here it is better to check whether it is equal to "item_cooccurrence_count". If it is a third value then you could raise an exception.

recommenders/evaluation/spark_evaluation.py
anargyri: Yes, it would be a bit more complicated. If you require that the features are sparse vectors (as in the example notebook), an alternative approach would be to first put them in tabular form inside a dataframe and then use SQL operations (similar to the co-occurrence case but with an additional SQL multiplication). I was curious whether that would be faster or slower than the UDF. Let's leave it for another occasion, after all the features could be dense in certain scenarios.

recommenders/evaluation/spark_evaluation.py
anargyri: Why would it increase the code significantly? I think you would just need to add an extra argument in some functions and just pass this argument from calling to called function. So that you could write
`als_diversity=als_eval.diversity(method="cooccurrence")`
or something similar.
And you would remove the argument from the constructor that you added in the most recent commit.

anargyri: The main thing I would like to avoid is the user creating two SparkDiversityEvaluation objects when they want to compute both types of diversity or serendipity metrics. This is more user-unfriendly than the alternative. And imagine what would happen if we added more item similarity methods: they would need to create as many objects as the methods available.

YanZhangADS: it will looks something like 
image

YanZhangADS: @anargyri @gramhagen what do you think?  Should we make change to use the above method?

recommenders/evaluation/spark_evaluation.py
YanZhangADS: added into docstring.

recommenders/evaluation/spark_evaluation.py
YanZhangADS: addressed

gramhagen (approved):
lgtm

recommenders/evaluation/spark_evaluation.py
anargyri: Actually, you don't need to duplicate the code in this way. What I meant was to change some of the member DFs into dictionaries indexed by the similarity method. I.e. do something like this instead:
```
def user_diversity(self, item_sim_measure="item_cooccurrence_count"):
        if self.df_user_diversity[item_sim_measure] is None:
            self.df_intralist_similarity[item_sim_measure] = self._get_intralist_similarity(self.reco_df, item_sim_measure=item_sim_measure)
            self.df_user_diversity[item_sim_measure] = (
                self.df_intralist_similarity[item_sim_measure].withColumn(
                    "user_diversity", 1 - F.col("avg_il_sim")
                )
                .select(self.col_user, "user_diversity")
                .orderBy(self.col_user)
            )
        return self.df_user_diversity[item_sim_measure]
```
So the code doesn't really increase but it becomes less readable. I am fine both ways but it's a trade-off, i.e. shifting convenience from the end-user to the developer or vice versa.
@yueguoguo @miguelgfierro any opinions on this?

anargyri (approved):

Issue #1511, opened 2021-08-27
Python 3.7 Support


laserprec:
### Description
We should extend the support of `recommenders` to python 3.7.

### Expected behavior with the suggested feature
We should modify our CIs to run our checks in py37 and fix any relevant issue raised from it.

### Other Comments
@anargyri, I think you mentioned problems with numba in python3.7 in venv installation. Is there an existing issue we can link to? I can also take a look.

anargyri:
@laserprec I found a solution to this (see the above PR) by installing Surprise from source (the issue was caused by the .pyx files in Surprise which apparently were compiled with an incompatible numpy version).

laserprec:
@anargyri this is AWESOME! Thanks Andreas!

PR #1518, opened 2021-09-02
Try py37

laserprec:
### Description
Testing out our build in python3.7 and try to replicate the numba import issues that @anargyri mentioned. **(This is just a draft PR)**

### Related Issues
#1511 

### Checklist:

### DIFF
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -73,7 +73,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python: [3.6]
+        python: [3.6, 3.7]
         # different kinds of tests are located in tests/<unit|integration|smoke> folders
         test-kind: ['unit'] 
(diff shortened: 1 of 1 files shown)


anargyri:
> @anargyri, if I understand from this [thread](https://github.com/microsoft/recommenders/pull/1503#discussion_r694847748) correctly, we have this numba import issue in py37 because of a specific numpy version fixed by Tensorflow, so if we somehow update tensorflow would this go away?
> 
> Not sure if there are other pip-native solutions to this without involving conda.
> 
> **(This is just a draft PR to replicate the issue. Hopefully we can find a simple solution :))**

I suspect that tensorflow upgrade would allow us to upgrade python for venv & virtualenv too, but I am not completely sure. The reason I am not sure is that, while we need numba for some libraries with algos, numba is not included in [TF dependencies](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py). But hopefully the higher numpy version will be more compatible with the algos we import.

PR #713, opened 2019-04-05
python evaluation improvement

gramhagen:
### Description
cleanup to python evaluation - ranking section
moved decorators to merging methods in preparation of creating new dataframe caching decorator

### Checklist:

- [x] My code follows the code style of this project, as detailed in our [contribution guidelines](../CONTRIBUTING.md).
- [x] I have added tests.
- [x] I have updated the documentation accordingly.

PR #1522, opened 2021-09-10
addition of item similarity measure - python version

YanZhangADS:
### Checklist:

- [x ] I have added tests covering my contributions.
- [x ] I have updated the documentation accordingly.
- [x ] This PR is being made to `staging branch` and not to `main branch`.

### DIFF
--- a/recommenders/evaluation/python_evaluation.py
+++ b/recommenders/evaluation/python_evaluation.py
@@ -698,48 +698,131 @@ metrics = {
 }
 
 # diversity metrics
-class PythonDiversityEvaluation:
-    """Python Diversity Evaluator"""
+def check_column_dtypes_diversity_serendipity(func):
+    """Checks columns of DataFrame inputs
 
-    def __init__(
-        self,
+    This includes the checks on:
+
+    * whether the input columns exist in the input DataFrames
+    * whether the data types of col_user as well as col_item are matched in the two input DataFrames.
+    * whether reco_df contains any user_item pairs that are already shown in train_df
+    * check relevance column in reco_df
+    * check column names in item_feature_df
+
+    Args:
+        func (function): function that will be wrapped
+
+    Returns:
+        function: Wrapper function for checking dtypes.
+    """
+
+    @wraps(func)
+    def check_column_dtypes_diversity_serendipity_wrapper(
         train_df,
         reco_df,
         item_feature_df=None,
         item_sim_measure=DEFAULT_ITEM_SIM_MEASURE,
+        col_item_features=DEFAULT_ITEM_FEATURES_COL,
         col_user=DEFAULT_USER_COL,
         col_item=DEFAULT_ITEM_COL,
+        col_sim=DEFAULT_SIMILARITY_COL,
         col_relevance=None,
+        *args,
+        **kwargs
     ):
-        """Initializer.
+        """Check columns of DataFrame inputs
 
-        This is the Python version of diversity metrics evaluator.
-        The methods of this class calculate the following diversity metrics:
+        Args:
+          train_df (pandas.DataFrame): Data set with historical data for users and items they
+              have interacted with; contains col_user, col_item. Assumed to not contain any duplicate rows.
+          reco_df (pandas.DataFrame): Recommender's prediction output, containing col_user, col_item,
+              col_relevance (optional). Assumed to not contain any duplicate user-item pairs.
+          item_feature_df (pandas.DataFrame): (Optional) It is required only when item_sim_measure='item_feature_vector'. It contains two columns: col_item and features (a feature vector).
+          item_sim_measure (str): (Optional) This column indicates which item similarity measure to be used. Available measures include item_cooccurrence_count (default choice) and item_feature_vector.
+          col_item_features (str): item feature column name.
+          col_user (str): User id column name.
+          col_item (str): Item id column name.
+          col_relevance (str): This column indicates whether the recommended item is actually
+              relevant to the user or not.
+        """
 
-        * Coverage - it includes two metrics:
-            1. catalog_coverage, which measures the proportion of items that get recommended from the item catalog;
-            2. distributional_coverage, which measures how unequally different items are recommended in the
-               recommendations to all users.
+        if not has_columns(train_df, [col_user, col_item]):
+            raise ValueError("Missing columns in train_df DataFrame")
+        if not has_columns(reco_df, [col_user, col_item]):
+            raise ValueError("Missing columns in reco_df DataFrame")
+        if not has_same_base_dtype(train_df, reco_df, columns=[col_user, col_item]):
+            raise ValueError("Columns in provided DataFrames are not the same datatype")
+        if col_relevance is None:
+            col_relevance = DEFAULT_RELEVANCE_COL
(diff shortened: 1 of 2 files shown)


recommenders/evaluation/python_evaluation.py
```diff
@@ -743,6 +747,8 @@ def __init__(
                 Interaction here follows the *item choice model* from Castells et al.
             reco_df (pandas.DataFrame): Recommender's prediction output, containing col_user, col_item,
                 col_relevance (optional). Assumed to not contain any duplicate user-item pairs.
+              item_feature_df (pandas.DataFrame): (Optional) It is required only when item_sim_measure='item_feature_vector'. It contains two columns: col_item and features (a feature vector).
+            item_sim_measure (str): (Optional) This column indicates which item similarity measure to be used. Available measures include item_cooccurrence_count (default choice) and item_feature_vector.
             col_user (str): User id column name.
             col_item (str): Item id column name.
             col_relevance (str): This column indicates whether the recommended item is actually
```
miguelgfierro, line 754: There is one thing I wanted to discuss with you all @anargyri, @YanZhangADS, @gramhagen, @yueguoguo, @loomlike. 

There is a disparity between the way we are doing rating and ranking metrics, and diversity metrics. In rating and ranking we use functions, whereas in diversity we use a class. I think it makes sense to write a class for pyspark, but I would like to get the views of people on if you guys are ok to do the same in python.

miguelgfierro (commented):
Yan this is really nice, I have one point for discussion

recommenders/evaluation/python_evaluation.py
anargyri: Good point. I agree, we should be consistent throughout the file. I think it's better to have them as functions, is there any good reason to have classes?

YanZhangADS: The main advantage of using class is, when one class object is initialized, several commonly shared variables do not need to be calculated multiple times.  e.g. both diversity and serendipity need to use item similarity matrix. By using class, we only need to calculate item similarity matrix once.

anargyri: You can avoid the recomputations without using classes. This is what we do in this file, using LRU cache https://github.com/microsoft/recommenders/blob/9b094e32550adbb326f40b02a7056e179a3e3da0/recommenders/evaluation/python_evaluation.py#L96 
See https://docs.python.org/3/library/functools.html#functools.lru_cache

anargyri: Which suggests that you could use LRU cache functionality to cache the similarity computations.

anargyri: This issue was the topic of some discussion in the past, see e.g. https://github.com/microsoft/recommenders/pull/713

laserprec: This is probably a related issue on optimization, but there is a substantial increase in the runtime of the test for the spark version. 
image
from
image

I hope we can reduce this runtime if we a smaller mock movielens dataset as we discussed before instead of the "100k" dataset for the notebook. Would welcome thoughts on other angles of optimization.

anargyri: What do the above screenshots compare? Computation time before and after the similarity based metrics were added?

laserprec: > What do the above screenshots compare? Computation time before and after the similarity based metrics were added?
They were comparing before and after the diversity metrics were added. I will open a different thread to discuss them as they are not related to the changes in this PR. Sorry for confusing the two.

miguelgfierro (commented):
This is really good Yan, you have changed the code super quickly. 
The only thing I see is that these changes also affect the notebooks right? but the tests didn't fail, so are we not testing the diversity notebook?

YanZhangADS:
> This is really good Yan, you have changed the code super quickly.
> The only thing I see is that these changes also affect the notebooks right? but the tests didn't fail, so are we not testing the diversity notebook?

diversity notebook only use spark version as an example. We do not have example diversity notebook for python version.

recommenders/evaluation/python_evaluation.py
```diff
@@ -696,46 +698,131 @@ def get_top_k_items(
 }
 
 # diversity metrics
-class PythonDiversityEvaluation:
-    """Python Diversity Evaluator"""
+def check_column_dtypes_diversity_serendipity(func):
```
anargyri, line 701: Shouldn't we have this decorator private (with an '_' at the front of the name)?

recommenders/evaluation/python_evaluation.py
```diff
@@ -696,46 +698,131 @@ def get_top_k_items(
...
-            P. Castells, S. Vargas, and J. Wang, Novelty and diversity metrics for recommender systems:
-            choice, discovery and relevance, ECIR 2011
+    Returns:
+        function: Wrapper function for checking dtypes.
+    """
 
-            Eugene Yan, Serendipity: Accuracy’s unpopular best friend in Recommender Systems,
-            eugeneyan.com, April 2020
+    @wraps(func)
+    def check_column_dtypes_novelty_coverage_wrapper(
```
anargyri, line 817: Similar, should this be private?

recommenders/evaluation/python_evaluation.py
```diff
@@ -696,46 +698,131 @@ def get_top_k_items(
...
-            2. distributional_coverage, which measures how unequally different items are recommended in the
-               recommendations to all users.
+    return check_column_dtypes_diversity_serendipity_wrapper
 
-        * Novelty - A more novel item indicates it is less popular, i.e. it gets recommended less frequently.
-        * Diversity - The dissimilarity of items being recommended.
-        * Serendipity - The "unusualness" or "surprise" of recommendations to a user. When 'col_relevance' is used, it indicates how "pleasant surprise" of recommendations is to a user.
 
-        The metric definitions/formulations are based on the following references with modification:
+def check_column_dtypes_novelty_coverage(func):
```
anargyri, line 800: Similar, should this be private?

anargyri (approved):
Good job, thanks!

anargyri:
There are some long lines (caught by flake).

recommenders/evaluation/python_evaluation.py
YanZhangADS: We did not use "_" in existing code, e.g. the function "check_column_dtypes" does not have "_" in front of the function name. Therefore I don't add "_" to be consistent.

anargyri: No, I meant adding underscore (_), not quotes. Currently you see these methods in [readthedocs](https://microsoft-recommenders.readthedocs.io/en/staging/evaluation.html#recommenders.evaluation.python_evaluation.check_column_dtypes), where they are probably not needed.

anargyri: There is another issue with the docstrings. The args are missing in the python functions e.g. [here](https://microsoft-recommenders.readthedocs.io/en/staging/evaluation.html#recommenders.evaluation.python_evaluation.serendipity) because before they were inside the encapsulating class but now they are required.

YanZhangADS:
Thanks @anargyri for catching many issue!

Issue #1439, opened 2021-06-14
[BUG] Move MMLSPARK info from tools to pip installable package


anargyri:
### Description

`pytest  tests/unit/examples/test_notebooks_pyspark.py` requires info from directory `tools/`:

```
E           papermill.exceptions.PapermillExecutionError:
E           ---------------------------------------------------------------------------
E           Exception encountered at "In [1]":
E           ---------------------------------------------------------------------------
E           ModuleNotFoundError                       Traceback (most recent call last)
E           <ipython-input-1-a0f1f3501b03> in <module>
E                17 if not is_databricks():
E                18     # get the maven coordinates for MML Spark from databricks_install script
E           ---> 19     from tools.databricks_install import MMLSPARK_INFO
E                20     packages = [MMLSPARK_INFO["maven"]["coordinates"]]
E                21     repo = MMLSPARK_INFO["maven"].get("repo")
E
E           ModuleNotFoundError: No module named 'tools'

/anaconda/envs/reco/lib/python3.6/site-packages/papermill/execute.py:234: PapermillExecutionError
```
We could move the required MMLSPARK_INFO from `tools/` to a definition built inside the pip package.

### In which platform does it happen?

Any platform

### How do we replicate the issue?

 * Create a conda / virtual environment 
 * `pip install ms-recommenders[examples,spark]`
* Run unit test `tests/unit/examples/test_notebooks_pyspark.py` with `pytest -k criteo` 

### Expected behavior (i.e. solution)

The tests should pass successfully even if `tools/` directory is not present

gramhagen:
it would be nice to have the mmlspark version set in setup.py with the rest of the dependencies, but i think that would mean setup.py would have to modify a separate file at installation (reco_utils/common/spark_utils.py?) to inject the location of the spark package & maven repo. Alternatively we could just put that information directly into spark_utils and put a note in setup.py that if you want to change the default version of mmlspark to modify spark_utils.py?

were you thinking that there would be an additional flag in start_or_get_spark() that would add mmlspark, or just store the info and leave it up to the user to retrieve it and pass it to start_or_get_spark() using the packages and repositories inputs?

anargyri:
I am open to suggestions. I was thinking of defining MMLSPARK_INFO somewhere in `reco_utils/common/spark_utils.py` where `databricks_install.py` would import it from, and it will also be available from the pip package.

[...]

PR #1531, opened 2021-09-20
Fix Databricks installation script and LightGBM test  

anargyri:
### Description

Fixed issues arising in https://github.com/microsoft/recommenders/issues/1439
- moved MMLSPARK_INFO so that the failing LightGBM test passes
- added mmlspark related arguments in `start_or_get_spark()` following Scott's suggestion
- enclosed occurrences of `start_or_get_spark()` inside a check for `not is_databricks()` 

Moreover, 
- changed `databricks_install.py` to install from PyPI instead of egg file
- fixed some azure related dependencies that were breaking the installation on Databricks.
 
### Related Issues

https://github.com/microsoft/recommenders/issues/1439

### Checklist:

### DIFF
--- a/recommenders/utils/spark_utils.py
+++ b/recommenders/utils/spark_utils.py
@@ -2,7 +2,6 @@
 # Licensed under the MIT License.
 
 import os
-import sys
 
 
 try:
@@ -11,6 +10,9 @@ except ImportError:
     pass  # skip this import if we are in pure python environment
 
 
+MMLSPARK_PACKAGE = "com.microsoft.ml.spark:mmlspark_2.11:0.18.1"
+MMLSPARK_REPO = "https://mvnrepository.com/artifact"
+
 def start_or_get_spark(
     app_name="Sample",
     url="local[*]",
@@ -18,18 +20,18 @@ def start_or_get_spark(
     config=None,
     packages=None,
     jars=None,
-    repository=None,
+    repositories=None,
 ):
     """Start Spark if not started
 
     Args:
-        app_name (str): Set name of the application
+        app_name (str): set name of the application
         url (str): URL for spark master
-        memory (str): Size of memory for spark driver
+        memory (str): size of memory for spark driver
         config (dict): dictionary of configuration options
         packages (list): list of packages to install
         jars (list): list of jar files to add
-        repository (str): The maven repository
+        repositories (list): list of maven repositories
 
     Returns:
         object: Spark context.
@@ -40,8 +42,8 @@ def start_or_get_spark(
         submit_args = "--packages {} ".format(",".join(packages))
     if jars is not None:
         submit_args += "--jars {} ".format(",".join(jars))
-    if repository is not None:
-        submit_args += "--repositories {}".format(repository)
+    if repositories is not None:
+        submit_args += "--repositories {}".format(",".join(repositories))
     if submit_args:
         os.environ["PYSPARK_SUBMIT_ARGS"] = "{} pyspark-shell".format(submit_args)
 
--- a/examples/01_prepare_data/data_split.ipynb
+++ b/examples/01_prepare_data/data_split.ipynb
@@ -2,51 +2,39 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {},
    "source": [
     "<i>Copyright (c) Microsoft Corporation. All rights reserved.</i>\n",
     "\n",
     "<i>Licensed under the MIT License.</i>"
-   ]
+   ],
+   "metadata": {}
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
    "source": [
     "# Data split"
-   ]
+   ],
+   "metadata": {}
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
    "source": [
     "Data splitting is one of the most vital tasks in assessing recommendation systems. Splitting strategy greatly affects the evaluation protocol so that it should always be taken into careful consideration by practitioners.\n",
     "\n",
     "The code hereafter explains how one applies different splitting strategies for specific scenarios."
-   ]
+   ],
+   "metadata": {}
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
    "source": [
     "## 0 Global settings"
-   ]
+   ],
+   "metadata": {}
   },
   {
    "cell_type": "code",
    "execution_count": 3,
-   "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "System version: 3.6.0 | packaged by conda-forge | (default, Feb  9 2017, 14:36:55) \n",
-      "[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)]\n",
-      "Pyspark version: 2.2.1\n"
-     ]
-    }
-   ],
    "source": [
     "# set the environment path to find Recommenders\n",
     "import sys\n",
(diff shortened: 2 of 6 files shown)


examples/00_quick_start/als_movielens.ipynb
```diff
@@ -2,46 +2,34 @@
  "cells": [
```
miguelgfierro, line 2: <p>@anargyri, queston about this, why are you checking if we are on databricks? </p><p>are you trying to make sure that the exact same notebook can work on a DSVM and in Databricks?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1531/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

anargyri: Yes, now the notebook won't fail on Databricks (it was failing because `start_or_get_spark()` fails).
This was a suggestion by Le here https://github.com/microsoft/recommenders/issues/1439

miguelgfierro: one thing that is a little bit weird is that we have `start_or_get_spark` but then we are adding an if clause to all notebooks. 

in the code of that function at the end we have:
```
spark_opts.append("getOrCreate()")
```
so if we add `start_or_get_spark` into Databricks, since spark instance already exists, it should return it directly.

Based on that, it should be possible to remove the if in:
```
if not is_databricks():
    spark = start_or_get_spark("ALS PySpark", memory="16g")
```
and leave only:
```
spark = start_or_get_spark("ALS PySpark", memory="16g")
```

anargyri: Actually, if you call `start_or_get_spark()` on Databricks it fails.
We already had the `if not is_databricks():` clause before this PR in a couple of notebooks. I just added it to the rest.

anargyri: But one alternative that may look better is to incorporate the `if not is_databricks()` check inside the definition of `start_or_get_spark()`.

recommenders/utils/spark_utils.py
```diff
@@ -11,37 +11,50 @@
...
     app_name="Sample",
     url="local[*]",
     memory="10g",
     config=None,
     packages=None,
     jars=None,
-    repository=None,
+    repositories=None,
+    mmlspark_package=None,
+    mmlspark_repository=None
```
miguelgfierro, line 26: I saw the comment of @gramhagen and @yueguoguo here: https://github.com/microsoft/recommenders/issues/1439 about adding mmlspark in the signature.

Another approach could be the following, having a generic function that could input any package, jar, etc:
```
def start_or_get_spark(
    app_name="Sample",
    url="local[*]",
    memory="10g",
    config=None,
    packages=None,
    jars=None,
repositories=None
)
```
and then a second function that includes specifically mmlspark, something like:
```
def start_or_get_spark_with_mmlspark(
    app_name="Sample",
    url="local[*]",
    memory="10g",
    config=None,
    packages=None,
    jars=None,
repositories=None

):
packages_full = mmlspark_package
if packages is not None:
    package_full+="".format(",".join(new_packages))
# same with repositories
```
The advantage of this is that we have a generalistic function for spark that can get any package, and another one explicitely for mmlspark.

What do you think @yueguoguo , @gramhagen , @anargyri? Happy to back off if you all guys agree we should go the original approach

anargyri: 🤔 I don't see a clear benefit from duplicating the function (the code of the two functions will be almost identical). 
I think the Python philosophy of specific rather than generalistic applies mainly when the functionalities differ significantly. In this case it's the same functionality essentially, just the repo info changes.

gramhagen: is it not possible to add mmlspark using the current packages/repositories inputs? just wondering if it's worth hiding that from the user and creating a library specific option. my preference would be to let the user provide these values and we can provide examples.

examples/00_quick_start/als_movielens.ipynb
gramhagen: I think this makes sense, if we put the if not is_databricks inside start_or_get_spark we can raise a warning that that function is not supported on spark. that said, do we know why it's failing and would it be better to resolve that?

recommenders/utils/spark_utils.py
anargyri: Yes, I was following your earlier suggestion, but this is another good option, to have just one function and let the user define the list of packages and repositories (If I understand correctly what you suggest).

examples/00_quick_start/als_movielens.ipynb
anargyri: Hmm, I just tried `start_or_get_spark()` on ADB with the PyPI reocmmenders package and it works.
I am pretty sure it failed at some point, but now I wonder if it was failing because of something wrong in the installation process.

recommenders/utils/spark_utils.py
gramhagen: right, would be good to test if this works
```
start_or_get_spark(packages=["com.microsoft.ml.spark:mmlspark_2.11:0.18.1"], repositories=["https://mvnrepository.com/artifact"])
```

examples/00_quick_start/als_movielens.ipynb
anargyri: I think what I was doing was something like
```
%%bash
python -c "from reco_utils.common.spark_utils import start_or_get_spark
start_or_get_spark()"
```
inside a cell. This raises an error in `start_or_get_spark()`. 
I doubt though that this is a recommended way to use Databricks.

anargyri: FYI @yueguoguo in case you have more experience with `start_or_get_spark()` on Databricks.

recommenders/utils/spark_utils.py
anargyri: This was already working [here](https://github.com/microsoft/recommenders/blob/main/examples/02_model_content_based_filtering/mmlspark_lightgbm_criteo.ipynb) (only change was the repos is now a list).

examples/00_quick_start/als_movielens.ipynb
miguelgfierro: > I just tried start_or_get_spark() on ADB with the PyPI recommenders package and it works

ok this is cool, and I guess the same function also works on DSVM spark right? then there is no need to add the if clause

recommenders/utils/spark_utils.py
miguelgfierro: got it makes sense, if you guys think that two functions (one general, one with mmlspark) are unnecessary, then I would lean more towards having just one function where we can have any arbitrary input:
`start_or_get_spark(packages=["com.microsoft.ml.spark:mmlspark_2.11:0.18.1"], repositories=["https://mvnrepository.com/artifact"])`, rather than a function with the specific mmlspark inputs:
```
start_or_get_spark(
    app_name="Sample",
    url="local[*]",
    memory="10g",
    config=None,
    packages=None,
    jars=None,
    repository=None,
    repositories=None,
    mmlspark_package=None,
    mmlspark_repository=None
```

examples/00_quick_start/als_movielens.ipynb
anargyri: Right. I can remove the `if not is_databricks()`, except where it is needed for other reasons.

recommenders/utils/spark_utils.py
anargyri: Ok, I will keep a single, simpler version then.

anargyri: @miguelgfierro @gramhagen if the recent changes look good, could you approve of the PR?

recommenders/utils/spark_utils.py
```diff
@@ -11,25 +10,28 @@
     pass  # skip this import if we are in pure python environment
 
 
+MMLSPARK_PACKAGE = "com.microsoft.ml.spark:mmlspark_2.11:0.18.1"
```
gramhagen, line 13: Do we need these lines now?

anargyri: Yes, for the mmlspark_lightgbm_criteo.ipynb notebook and databricks_install.py

gramhagen (approved):

PR #1538, opened 2021-09-24
Laserprec/opt unit tests exp

laserprec:
### Description
To improve developer experience working with this repo, we will begin migrating our DevOps pipelines into Github Action, leveraging popular DevOps tools like tox and flake8. This will be a sequence of updates to our DevOps infrastructures:

1. ~~Propose and draft the CI pipeline in Github Action~~
2. ~~Setup self-hosted machines to run our GPU workloads~~
3. ~~Create feature parity with the existing CI pipelines (pr-gates & nightly-build)~~
4. ~~Run tests on the appropriate dependency subsets~~
5. Optimize build time for unit tests <---- (Current PR)
6. Enforce flake8 (coding style checks) on the build and clean up coding styles to pass flake8 checks
7. Deprecate CI pipelines in ADO and switch to Github Actions

#### In this PR:

[Message not completed yet]

### Checklist:


examples/02_model_collaborative_filtering/als_deep_dive.ipynb
```diff
@@ -2,32 +2,31 @@
  "cells": [
```
miguelgfierro, line 2: <p>small detail: the output of this cell is not shown</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1538/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

recommenders/datasets/mock/movielens.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+Mock dataset schema to generate fake data for testing use. This will mimic the Movielens Dataset
+"""
+try:
+    import pandera as pa
+except ImportError:
+    raise ImportError("pandera is not installed. Try `pip install recommenders['dev']`")
```
miguelgfierro, line 10: Is this clause to avoid breaking the code if people don't install dev?

recommenders/datasets/mock/movielens.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+Mock dataset schema to generate fake data for testing use. This will mimic the Movielens Dataset
```
miguelgfierro, line 5: @laserprec can we discuss the pros and cons of having this function on a different file vs having it in the movielens.py file?

recommenders/datasets/mock/movielens.py
```diff
@@ -0,0 +1,140 @@
...
+    ):
+        """Return fake movielens dataset as a Spark Dataframe with specified rows
+
+        Args:
+            spark (SparkSession): spark session to load the dataframe into
+            size (int): number of rows to generate
+            seed (int): seeding the pseudo-number generation. Defaults to 100.
+            keep_title_col (bool): remove the title column if False. Defaults to False.
+            keep_genre_col (bool): remove the genre column if False. Defaults to False.
+            tmp_path (str, optional): path to store files for serialization purpose
```
miguelgfierro, line 106: does pandera return pure random data or it follows some kind of distribution?

@yueguoguo and Simon did a random generator based on scikit or similar? what is the benefit of pandera vs creating our own random data with a library that we already use in the repo like sklearn or numpy?

tests/unit/examples/test_notebooks_pyspark.py
```diff
@@ -28,12 +30,20 @@ def test_data_split_runs(notebooks, output_notebook, kernel_name):
...
     sys.platform == "win32", reason="Takes 2764.50s in Windows, while in Linux 124.35s"
 )
-def test_als_deep_dive_runs(notebooks, output_notebook, kernel_name):
+@pytest.mark.parametrize("data_size", ["mock100"])
+def test_als_deep_dive_runs(notebooks, output_notebook, kernel_name, data_size):
     notebook_path = notebooks["als_deep_dive"]
-    pm.execute_notebook(notebook_path, output_notebook, kernel_name=kernel_name)
+    pm.execute_notebook(notebook_path, output_notebook, kernel_name=kernel_name,
+                        parameters=dict(
+                            MOVIELENS_DATA_SIZE=data_size,
```
miguelgfierro, line 42: trying to think of simpler solutions, why not just input `MOVIELENS_DATA_SIZE="mock100"` instead of adding `@pytest.mark.fake_movielens` and `@pytest.mark.parametrize("data_size", ["mock100"])`

tests/unit/recommenders/datasets/mock/test_movielens.py
```diff
@@ -0,0 +1,86 @@
+import os
```
miguelgfierro, line 1: Traditionally we have tried to not add too many folder levels, just to reduce the complexity of the repo. Do we really need the mock folder, or can we directly add the test inside datasets?

recommenders/datasets/mock/movielens.py
laserprec: Yup.

examples/02_model_collaborative_filtering/als_deep_dive.ipynb
laserprec: Thanks for pointing that out! Will regenerate it.

recommenders/datasets/mock/movielens.py
```diff
@@ -0,0 +1,140 @@
...
+                when transferring data from python to java.
+                If None, a temporal path is used instead
+
+        Returns:
+            pyspark.sql.DataFrame: a mock dataset
+        """
+        pandas_df = cls.get_df(size=size, seed=seed, keep_title_col=True, keep_genre_col=True)
+
+        # generate temp folder
```
gramhagen, line 115: why are we writing the pandas dataframe to disk to read into spark dataframe?
you should be able to do something like `spark.createDataFrame(pandas_df, schema=schema)`

recommenders/datasets/mock/movielens.py
gramhagen: i think we might be able to simplify a few things if we just push this directly into movielens.py

recommenders/datasets/movielens.py
```diff
@@ -100,6 +96,11 @@ def item_has_header(self):
     "20m": _DataFormat(",", "ml-20m/ratings.csv", True, ",", "ml-20m/movies.csv", True),
 }
 
+# Fake data for testing only
+MOCK_DATA_FORMAT = {
```
gramhagen, line 100: why not just add this to data_format?

tests/unit/examples/test_notebooks_pyspark.py
laserprec: `@pytest.mark.fake_movielens` and the parametrization were artifacts of producing the offline experiment results included in this PR. It's mainly there so I can do something like "pytest tests/unit -m 'fake_movielens'" 😄. If you think the results are convincing enough and we don't need to regenerate them, I can remove them.

recommenders/datasets/mock/movielens.py
laserprec: Yes, I've tried exactly that (`spark.createDataFrame(pandas_df)`) and this actually incurs a very expensive communication between java and python each time the spark.dataframe is queried that actually becomes a new bottleneck. I think what happens is that when we create a spark.dataframe from pandas.dataframe, data remains in the memory as Python objects, and the spark.dataframe is merely a wrapper on top the Python object (thus the expensive java <-> python communication). To rid of this wrapper behavior, we need to serialized the data and reload into java. I think this is what pyarrow is trying to resolved.

This is at least what I observed in Spark 2.4.4 without [pyarrow ](https://docs.databricks.com/spark/latest/spark-sql/spark-pandas.html)enabled. Not sure if there's a better way to do this.

recommenders/datasets/movielens.py
laserprec: The composition of the object is quite different: `DATA_FORMAT` uses an instance of `_DataFormat` object which contains data deserialization controls that doesn't fit the need to create a mock data, where you just need to provide the data size and seed for the pRNG. I figured it's simpler to separate the two to make the client of these objects easier.

recommenders/datasets/mock/movielens.py
laserprec: > does pandera return pure random data or it follows some kind of distribution?

Under the hood, [pandera](https://pandera.readthedocs.io/en/latest/data_synthesis_strategies.html#data-synthesis-strategies) uses hypothesis, a popular data generation tool, to generate the fake data. By default, hypothesis uses the global pRNG in python to generate the random data, so we can provide deterministic random data with a fixed seed (it is what we are doing right now). For more detail [see this doc](https://hypothesis.readthedocs.io/en/latest/details.html#making-random-code-deterministic).

For the pros & cons of pandera vs implementation, I think there are a few arguments here:
1. Avoid re-inventing the wheel (as always 😝).
2. Easier use: data generation follows the validation rules you declared in the schema (for example, you can restrict the range of certain columns and random data will NOT provide anything outside the range).
3. Pandera's data schema is not used only for data generation, but also for [data validation](https://pandera.readthedocs.io/en/latest/schema_models.html). This is a bit biased anecdote, but data validation is such an important piece and a paint-point in a data pipeline, some of my previous work would have been longer to debug without neat validation like one provided by Pandera. This tool is very similar to [pydantic](https://pydantic-docs.helpmanual.io/) in data model validation but designed for `pandas.DataFrame` specially.

recommenders/datasets/mock/movielens.py
laserprec: I think this is my mental activity planning for this change 😄:

The sentimental part of me is like:
It pains me to add another 100+ lines to an already long 500 document.

The analytics part is like:
1. The new folder structure can contain other "mock" dataset schemas besides `movielens` (although one can argue that this over engineering for something that may not happen)
2. A clear separation between the code for testing purpose and the normal intended use

I am okay with keeping it separated or combined. But here is a summary:
- If we think this mocking strategy can be applied to other datasets for optimization, I think keeping it separated is good.
- If we think this is over-engineering and unnecessary complexity I think we can push into movielens.py

recommenders/datasets/movielens.py
gramhagen: fair, i'm just lazy i guess =)

recommenders/datasets/mock/movielens.py
gramhagen: can you enable pyarrow?
`spark.conf.set("spark.sql.execution.arrow.enabled", "true")`

laserprec: Let me try :).

examples/02_model_collaborative_filtering/als_deep_dive.ipynb
laserprec: @miguelgfierro, it turns out the output of this cell is still present, as you can see from [my branch](https://github.com/microsoft/recommenders/blob/laserprec/opt-unit-tests-exp/examples/02_model_collaborative_filtering/als_deep_dive.ipynb). Not sure what ReviewNB is reporting a removal there. Maybe a bug on their end 🤣 .

recommenders/datasets/mock/movielens.py
laserprec: @gramhagen, here's the result of using pyarrow. Unfortunately, it didn't bring any performance improvements when compared to using explicit serialization.

Here's the runtime of a notebook WITHOUT any serialization/pyarrow enabled (vanilla `spark.createDataFrame(pandas_df, schema=schema)`):
diversity_metric_opt_no_serialization

Here's the runtime with pyarrow (v0.14.0) enabled:
diversity_metric_opt_no_serialization_w_pyarrow3_no_error

Here's the runtime with explicit serialization (i.e. writing a CSV to disk and `spark.read.csv`):
diversity_metric_opt_serialization

I did [encounter this warning](https://www.py4u.net/discuss/1630531) when trying to enable pyarrow in spark 2, but I managed overcome it by downgrading to pyarrow==0.14.0, following the solution in the thread. Still it doesn't bring any improvements 🤣. I would we can we can revisit pyarrow after we migrate to spark 3.

gramhagen: ok, makes sense. thanks for checking this.

examples/02_model_collaborative_filtering/als_deep_dive.ipynb
miguelgfierro: weird

recommenders/datasets/mock/movielens.py
miguelgfierro: wow, that's a big difference. @laserprec do you think spark 3 will improve the situation?

recommenders/datasets/mock/movielens.py
miguelgfierro: got it, makes sense

tests/unit/examples/test_notebooks_pyspark.py
miguelgfierro: We are trying not to add too many custom marks, otherwise, we will clutter the repo too much. 
I think in general, whenever we can reduce or simplify the code in the repo, we are facilitating the work of our users

recommenders/datasets/mock/movielens.py
miguelgfierro: @laserprec I think we should do the approach that is easier for users, what approach do you think they would prefer the most?

recommenders/datasets/mock/movielens.py
laserprec: I hope so. Ultimately, this java-to-python inefficiency is what [pyarrow is designed to optimize for](https://spark.apache.org/docs/latest/api/python/user_guide/arrow_pandas.html#apache-arrow-in-pyspark). We could have some join operations in our notebooks that made it difficult to leverage pyarrow, but I think it's a good idea to revisit with Spark 3.

recommenders/datasets/mock/movielens.py
laserprec: I suspect the users of this code would be folks developing with the repo. Since this mocking strategy is only applied to movielens dataset, it would be better to keep it together and avoid adding more folder complexity without an immediate clear scenario for reusability. I can refactor the code and push into `movielens.py`.

recommenders/datasets/movielens.py
```diff
@@ -154,14 +149,17 @@ def load_pandas_df(
     To load movie information only, you can use load_item_df function.
 
     Args:
-        size (str): Size of the data to load. One of ("100k", "1m", "10m", "20m").
-        header (list or tuple or None): Rating dataset header.
-        local_cache_path (str): Path (directory or a zip file) to cache the downloaded zip file.
+        size (str): Size of the data to load. One of ("100k", "1m", "10m", "20m", "mock100").
+        header* (list or tuple or None): Rating dataset header.
```
anargyri, line 153: What is the meaning of * here? Is it described in the Google style?

recommenders/datasets/movielens.py
```diff
@@ -196,6 +194,17 @@ def load_pandas_df(
         warnings.warn(WARNING_MOVIE_LENS_HEADER)
         header = header[:4]
 
+    if size in MOCK_DATA_FORMAT:
+        # function-wide import to isolate extra dependencies from the mock schema will use
+        from recommenders.datasets.mock.movielens import MockMovielensSchema
```
anargyri, line 199: This import statement is probably better at the beginning of the file according to PEP guidelines.

recommenders/datasets/mock/movielens.py
anargyri: I am not sure I follow all the stuff happening here, but Isn't there any simpler and faster way to create a mock Spark DF? A way that avoids going through Pandas DF and read / write to file. How about creating a random numpy array and initializing a Spark DF using that array? I suspect that would take less time, it would avoid the I/O overhead at least and probably numpy would be faster than pandas.

examples/03_evaluate/als_movielens_diversity_metrics.ipynb
```diff
@@ -153,7 +153,7 @@
     "import pyspark\n",
```
anargyri, line 153: <p>Is there a reason we want to change COL_ITEM? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1538/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

anargyri, line 153: <pre class="ql-syntax" spellcheck="false">Line #7.            StructField("Timestamp", StringType()),
</pre><p>Why do we need the type change here? Is there a reason to change the signature of load_spark_df() ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/recommenders/pull/1538/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

laserprec: <p>Not particular :D. I think this is an artifact of me pasting from the default column constant. I can change it back to "MovieId".</p>

recommenders/datasets/mock/movielens.py
laserprec: @anargyri, that's a good point, thanks for bring that up. I think the main problem here is how we initialize the Spark DF. If we initialize from any Python object (whether it's pandas DF or numpy arrays), during code excecution, inevitably there will be a data exchange from python to Java because the data is in memory as some python object but spark runs on a JVM. This data exchange is source of this bottleneck, and I don't see a way around it as long as the data is a Python object. To your question, I think the easiest way to initialize a Spark DF would be reading from a CSV file. However, storing static data files in the repo will have its own challenge and inconvenience.

I find reusing existing libraries for fake data generation being a more flexible approach, so we don't have to reinvent the wheel. I find [pandera's schema](https://pandera.readthedocs.io/en/stable/schema_models.html#schema-models) definition by far the most intuitive and simple interface for fake data generation. @anargyri, if you have any other suggestions, I am open to other options too 😃.

examples/03_evaluate/als_movielens_diversity_metrics.ipynb
laserprec: Oops, this is some artifacts of me testing whether the data types has any impact to the runtime. I can revert the change.

recommenders/datasets/movielens.py
laserprec: I intended to use it like a footnote notation for function arguments that are ignored when a mock data size is specified. I can remove it and replace with a explicit statement.

recommenders/datasets/movielens.py
laserprec: Removed

examples/03_evaluate/als_movielens_diversity_metrics.ipynb
laserprec: Done

recommenders/datasets/movielens.py
laserprec: Done

tests/unit/recommenders/datasets/mock/test_movielens.py
laserprec: File merged into movielens.

recommenders/datasets/mock/movielens.py
laserprec: Done

PR #1549, opened 2021-10-14
Use spark3 by default

laserprec:
### Description
1. ~~Set `pyspark` dependency to `>=3.0.0, <4.0.0`~~
    - Update to  `>=2.4.5, <3.2.0`
2. Update CIs to run the build in spark3

### Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ extras_require = {
     "spark": [
         "databricks_cli>=0.8.6,<1",
         "pyarrow>=0.12.1,<6.0.0",
-        "pyspark>=2.4.5,<4.0.0",
+        "pyspark>=2.4.5,<3.2.0",
     ],
     "xlearn": [
         "cmake>=3.18.4.post1",


anargyri:
@laserprec @miguelgfierro I wonder whether it is possible to specify a custom command line flag that tells `pip install` to install v2 instead of v3 (I don't mean using an additional extras option). In case we would like to allow v2 as an option to the users.

gramhagen:
> @laserprec @miguelgfierro I wonder whether it is possible to specify a custom command line flag that tells `pip install` to install v2 instead of v3 (I don't mean using an additional extras option). In case we would like to allow v2 as an option to the users.

I don't know of an approach other than using extras. Was the issue with this change that version 3 is not installed by default?

gramhagen:
> > :-> Yes, I agree, let's do `<3.2.0`
> 
> Sounds good. I can apply this change.
> 
> I think we have a consensus on not introducing an extra dependency solely for spark2 support (e.g. ` pip install recommenders[spark2]` ), @gramhagen and @miguelgfierro, please let us know if you have other thoughts. If no objections, we will run our pipelines in Java 11 + pyspark v3.1.2 (default version that satisfies `>=2.4.5, <3.2.0`) and drop support for spark 2 & java 8 in the pipeline.
> 
> > And `>= 2.4.5` if everyone else agrees`

2.4.5<=pyspark<3.2.0 is fine with me, can we add an issue capturing what needs to change to support >=3.2.0?

laserprec:
> can we add an issue capturing what needs to change to support >=3.2.0?

Done. Please see #1553. The errors are accessible in the pipeline history and are linked in the issue.

PR #1612, opened 2022-01-17
Update NCF Dataset code

angusrtaylor:
### Description
- **Fixed out-of-memory error**: Users of the original code encountered out-of-memory errors when creating the Dataset object for large datasets. This is because the original code attempts to create a matrix of users and "negative items" (items the user has not interacted with) for negative sampling purposes. This matrix can have very large memory footprint. The updated code removes the need for creating this matrix by only identifying a user's negative item set when processing a user's data during training.
- **Improved memory efficiency**: The original code expands the original training dataset (containing only positive examples) by performing negative sampling for all users at the start of each training epoch and holding this full dataset in memory. This creates a memory bottleneck and causes problems when training on very large datasets. The updated code iterates over a csv file sorted by user and performs negative sampling for each user sequentially, never storing the full dataset in memory. A shuffle buffer is introduced to the train_loader function such that a subset of the full dataset is loaded into memory for shuffling purposes before returning batches. The size of the shuffle buffer is configurable.
- **Fixed sampling problem**: The original code samples user negative items every time it processes a positive example in the original training set. This is inefficient but also amounts to sampling with replacement which does not follow the approach of the original paper. The updated code samples negative items once per user without replacement. A new parameter is added to the Dataset class to enable sampling with replacement if desired.
- **Fixed training/test set information leakage**: in the original code, embeddings are created in the model for all users and items in both the training and test sets. This creates some information leakage for users/items that appear in the training set but not the test set. The updated code filters out unseen users/items from the test set in the notebook and additional filters are added to the Dataset object to prevent these users/items from being added to the test set.
- **Fixed leave-one-out evaluation example**: the original code runs the leave-one-out evaluation on the last 25% of the dataset using a chronological split which does not correspond to the approach taken in the original paper. The updated code aligns with the approach of the paper by only taking the last item for each user as the test set.

Note: updated unit tests are to follow.

### Checklist:

- [ x] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ x] I have added tests covering my contributions.
- [ x] I have updated the documentation accordingly.
- [ x] This PR is being made to `staging branch` and not to `main branch`.


miguelgfierro (commented):
looks great!

miguelgfierro:
@angusrtaylor there are some errors with flake, are you able to access the logs?

angusrtaylor:
> @angusrtaylor there are some errors with flake, are you able to access the logs?

Thanks I'll address this

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
+            test_batch = int(row[self.col_test_batch])
+        return {self.col_user: user, self.col_item: item, self.col_rating: rating, self.col_test_batch: test_batch}
+
+    def _init_data(self):
+        # Compile lists of unique users and items, assign IDs to users and items,
+        # and ensure file is sorted by user (and batch index if test set)
+        logger.info("Indexing {} ...".format(self.filename))
+        with self:
+            user_items = []
+            self.item2id, self.user2id = OrderedDict(), OrderedDict()
```
anargyri, line 110: Do you need an `OrderedDict`? I don't see any of its specialized functionality used in the code.

angusrtaylor: I'm using OrderedDict to be memory efficient. It serves two purposes:
1. I need a dict to hold the mapping between original user/item values and their IDs
2. I need an ordered list to hold users so I can iterate through them in the order they appear in the training data file
Ordered dict allows me to achieve both these requirements in one object (the `users` attribute can be obtained from the `users` property which just retrieves the OrderedDict keys). If I don't use OrderedDict I think I would have to have both a dictionary and a list each containing all users.

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
+
+
+class DataFile():
+    """
+    DataFile class for NCF. Iterator to read data from a csv file.
+    Data must be sorted by user. Includes utilities for loading user data from
+    to file, formatting it and returning a Pandas dataframe.
+    """
+
+    def __init__(
```
anargyri, line 26: Black formatting of the arguments etc. would be nice.

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
+        self, filename, col_user, col_item, col_rating,
+        col_test_batch=None, binary=True
+    ):
+        self.filename = filename
```
anargyri, line 30: Docstring would be nice too (and would be visible in readthedocs).

recommenders/models/ncf/dataset.py
anargyri: `users` returns the output of `keys()`. You could do the same with the standard `dict`, couldn't you?
The LIFO or FIFO functionality provided by OrderedDict isn't used as far as I can see.

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
+        self.col_user = col_user
+        self.col_item = col_item
+        self.col_rating = col_rating
+        self.col_test_batch = col_test_batch
+        self.expected_fields = [self.col_user, self.col_item, self.col_rating]
+        if self.col_test_batch is not None:
+            self.expected_fields.append(self.col_test_batch)
+        self.binary = binary
+        self.user2id, self.item2id, self.id2user, self.id2item = None, None, None, None
```
anargyri, line 39: id2user, id2item are not used in `DataFile`
Also no need to initialize the other two, since they are initialized in the following line.

angusrtaylor: id2user and id2item are not used in DataFile but are included as a helper to make it easy to check a user/item against a given ID. I'll remove the unnecessary initialization.

anargyri: But this is in DataSet, not DataFile. In DataFile they are not intialized.

recommenders/models/ncf/dataset.py
anargyri: Yes, we require 3.6 - 3.8 in fact 
https://github.com/microsoft/recommenders/blob/9898ec6f5bc611a23657a6f3395ef9403f210344/setup.py#L142

anargyri: Actually, maybe it's better to keep OrderedDict just to be on the safe side. Insertion order became a platform independent feature in Python 3.7 https://mail.python.org/pipermail/python-dev/2017-December/151283.html. So maybe there will be a problem if someone uses 3.6 elsewhere than CPython.

anargyri:
Is there any testing for the new functionality that should be added in test_ncf_dataset.py ?

angusrtaylor:
> Is there any testing for the new functionality that should be added in test_ncf_dataset.py ?

Yes I need to revisit the unit tests. I'm working on that now.

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
+        # if sampling without replacement, check sample population is sufficient and reduce
+        # n_samples if not.
+        n_neg_var = "n_neg" if training else "n_neg_test"
+        dataset_name = "training" if training else "test"
+
+        new_n_samples = n_samples
+        if not self.sample_with_replacement:
+            population_size = len(user_negative_item_pool)
+            if n_samples > population_size:
+                new_n_samples = min(n_samples, population_size)
```
anargyri, line 301: How about incorporating this as `k=min(n_samples, len(user_negative_item_pool))` inside `_sample_negatives_without_replacement()` and do away with this function?
Another question: how is the negative sampling according to the paper, with a fixed sample size or proportional to the user's population size?

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
+        prepare_batch = self._prepare_batch_with_id if yield_id else self._prepare_batch_without_id
+        shuffle_buffer_df = pd.concat(shuffle_buffer)
+        shuffle_buffer_df = shuffle_buffer_df.sample(shuffle_buffer_df.shape[0])  # shuffle the buffer
+        for batch in self._split_into_batches(shuffle_buffer_df, batch_size):
+            if batch.shape[0] == batch_size:
+                if write_to:
+                    batch.to_csv(write_to, mode='a', header=False, index=False)
+                yield prepare_batch(batch)
+            else:
+                return batch
```
anargyri, line 383: Why is the return statement needed?

recommenders/models/ncf/dataset.py
angusrtaylor: It's possible. The reason I do this outside of `_sample_negatives_without_replacement` is that in `_create_test_file` I have to call `_sample_negatives_without_replacement` for every positive example for a user. If the sample size needs to be adjusted I raise a warning and ideally I'd only want to do this once per user rather than for every positive example. However, I should be able to parametrize `_sample_negatives_without_replacement` such that it would only raise a warning the first time it is called for each user. I'll make this change.

In the paper, 4 negative examples are sampled for every positive example so it's the ratio of positive to negative examples that are fixed for each user. A user with more positive examples will therefore have more negative examples. In the original code, this sampling is achieved by calling the sampling function once per positive example. However, I changed this because it's less efficient but also because it effectively amounts to sampling _with_ replacement as the same negative sample can appear more than once per user. In my code (during training and not for test file creation), I perform the sampling once per user without replacement so that the negatives for a user are unique. This also means however that the number of requested samples sometimes exceeds the population of negative items for a user and so some users have a smaller ratio of positives to negatives. I'm not sure which approach is better. The choices are:

1. Sample without replacement per user, guaranteeing complete diversity of negative samples, but having a reduced ratio of positive/negative samples for some users [my current code]
2. Sample without replacement per positive example, ensuring the ratio of positive/negative samples are the same for all users, but some negative examples will appear twice for a given user [old code]
3. Sample _with_ replacement per user, which is the same as option 2 but more efficient and guarantees equal ratios across users.
4. Sample without replacement per user, and if the resulting number of negative samples is less than required, sample again to make up the shortfall. This would be efficient and the keep the ratios constant across users but some negative samples will appear more than once (but only for users that have limited population sizes).
Do you have any thoughts on what is the best approach?

recommenders/models/ncf/dataset.py
angusrtaylor: When the shuffle buffer is released, it yields batches of fixed size. However, there is likely to be a remainder when splitting the shuffle buffer into fixed size batches and so the final return passes this remainder back to the train_loader so that it can start filling the fresh shuffle buffer with this remainder.

recommenders/models/ncf/dataset.py
anargyri: It looks like they do #2 in [their code](https://github.com/hexiangnan/neural_collaborative_filtering/blob/4aab159e81c44b062c091bdaed0ab54ac632371f/GMF.py#L92) but with a fixed num_negatives (not a proportion). I am not sure how they get their test data, it is available as a raw file. The paper says there are 100 test samples for each user.
I think with the data sets they use it is feasible to sample without duplicates (since there are many items and large sparsity).
I agree that duplicate examples are not a good idea, for reasons of efficiency but also because they effectively give more importance to some training examples compared to others. So I think your approach is good. The problems arise only when the sparsity is low (at least for some users), but then it's a matter for the user to decide what to do as a pre-processing step (it depends on each specific case).

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
+            records.append(row)
+            if not self.end_of_file:
+                next(self)
+            else:
+                break
+
+        return pd.DataFrame.from_records(records)
+
 
 class Dataset(object):
```
miguelgfierro, line 172: there is an error in the PR:

```
=================================== FAILURES ===================================
___________________________ test_data_preprocessing ____________________________

python_dataset_ncf = (     userID  itemID  rating  timestamp
299       1      96       1 2018-01-19
655       1      83       3 2018-08-24
...    3 2020-01-28
730      99      55       4 2020-02-20
728      99      11       4 2020-06-25

[399 rows x 4 columns])

    def test_data_preprocessing(python_dataset_ncf):
        train, test = python_dataset_ncf
        data = Dataset(
>           train=train, test=test, n_neg=N_NEG, n_neg_test=N_NEG_TEST, seed=SEED
        )
E       TypeError: __init__() got an unexpected keyword argument 'train'

tests/unit/recommenders/models/test_ncf_dataset.py:21: TypeError
______________________________ test_train_loader _______________________________

python_dataset_ncf = (     userID  itemID  rating  timestamp
299       1      96       1 2018-01-19
655       1      83       3 2018-08-24
...    3 2020-01-28
730      99      55       4 2020-02-20
728      99      11       4 2020-06-25

[399 rows x 4 columns])

    def test_train_loader(python_dataset_ncf):
        train, test = python_dataset_ncf
        data = Dataset(
>           train=train, test=test, n_neg=N_NEG, n_neg_test=N_NEG_TEST, seed=SEED
        )
E       TypeError: __init__() got an unexpected keyword argument 'train'

tests/unit/recommenders/models/test_ncf_dataset.py:45: TypeError
_______________________________ test_test_loader _______________________________

python_dataset_ncf = (     userID  itemID  rating  timestamp
299       1      96       1 2018-01-19
655       1      83       3 2018-08-24
...    3 2020-01-28
730      99      55       4 2020-02-20
728      99      11       4 2020-06-25

[399 rows x 4 columns])

    def test_test_loader(python_dataset_ncf):
        train, test = python_dataset_ncf
>       data = Dataset(train=train, test=test, n_neg=N_NEG, n_neg_test=N_NEG_TEST)
E       TypeError: __init__() got an unexpected keyword argument 'train'
```
see logs: https://github.com/microsoft/recommenders/runs/4944649632?check_suite_focus=true. 

However, I realised that apart from the error, these tests should be in the GPU set, but are in the CPU. So we need to add `@pytest.mark.gpu` to the tests

angusrtaylor: Yes the tests only work for the old code. I am still working on the revised set of tests. I'll be sure to place them in the GPU set.

recommenders/models/ncf/dataset.py
anargyri: But train_loader could get that from `yield`, couldn't it? In fact, I think the return syntax will not provide the remainder to the caller of this function, only yield does. Return has no effect in an iterator like this.

angusrtaylor: The `_release_shuffle_buffer` iterator does return batch if it does not have the required number of examples (which can only be the case for the last batch in the shuffle buffer). When the iterator hits the return statement it will stop the execution and return the batch to train_loader. The train_loader uses `yield from` syntax to yield batches from `_release_shuffle_buffer` to the calling function (the NCF model training function) so full batches are simply passed through the train_loader function and only the final incomplete batch is returned to train_loader. If the final incomplete batch were to be yielded, it would immediately be yielded to the model training function which we don't want.

I did try writing the code without this nested iterator approach initially but it was very very messy.

recommenders/models/ncf/dataset.py
```diff
@@ -1,321 +1,450 @@
...
-                        [self.id2item[x] for x in self.test_items],
-                        self.test_ratings,
-                    ]
+                user_negative_examples = self._get_negative_examples(
+                    user, user_negative_item_pool, n_samples
                 )
+                user_examples = pd.concat([user_positive_examples, user_negative_examples])
+                shuffle_buffer.append(user_examples)
+                shuffle_buffer_len = sum([df.shape[0] for df in shuffle_buffer])
+                if (shuffle_buffer_len >= shuffle_size):
```
anargyri, line 426: In this code `shuffle_buffer` reads data in the same order as users, right? So, if `shuffle_size` is small enough the training epoch will see the first users only according to the ordering of `train_datafile.users`. As a result the last users will never be used, will they? So I think you need to permute `train_datafile.users` in line 413.

angusrtaylor: Yes the `train_loader` always reads data in the same order, adding users' data to shuffle buffer. If shuffle buffer is small, it will get shuffled and released more often, but the buffer will always contain all the data for at least one user. After being released, the shuffle buffer will be initialized with any remainder following the last release and start appending next set of users' data. Line 433 will release any remaining data if there is still data in the shuffle buffer after iterating through all users. The only problem with small `shuffle_size` is that the quality of shuffle will be poor because the same users are likely to appear in the same batch over every epoch.

anargyri:
Another thing to check may be the benchmark runs, maybe try to run NCF in that notebook to check that it doesn't fail.

recommenders/models/ncf/dataset.py
angusrtaylor: @anargyri  I've tried moving the logic of `_check_sample_size` into the `_sample_negatives_without_replacement` function but the result is quite messy. If we want to just reduce the number of samples as you suggest it is no problem but if we also want to _warn_ users about it then I have to pass a bunch of extra information to the sampling function (as well as the `_get_negative_examples` function that calls it) in order to make the warnings meaningful. Plus I have to add an additional if statement to ensure we only print warnings once per user when creating the test set. I'm not sure it's improved anything by doing this but I can make the change if you want. What do you think?

recommenders/models/ncf/dataset.py
anargyri: I see. It looks correct then.

recommenders/models/ncf/dataset.py
anargyri: Ok then, better keep it as is.

recommenders/models/ncf/dataset.py
```diff
@@ -165,10 +187,99 @@ def load_data(self, key, by_user=True):
                 next(self)
             else:
                 break
-
         return pd.DataFrame.from_records(records)
 
 
+class NegativeSampler:
+    """
+    NegativeSampler class for NCF. Samples a subset of negative items from a given population of items.
```
anargyri, line 195: I think sphinx won't show this description, you need to remove the newline.

recommenders/models/ncf/dataset.py
```diff
@@ -165,10 +187,99 @@ def load_data(self, key, by_user=True):
...
+            self._sample_negatives_with_replacement
+            if self.sample_with_replacement
+            else self._sample_negatives_without_replacement
+        )
+        if not self.sample_with_replacement:
+            self._check_sample_size()
+
+    def sample(self):
+        """
+        Method for sampling uniformly from a population of negative items
```
anargyri, line 239: The docstring is not formatted properly.

tests/unit/recommenders/models/test_ncf_dataset.py
```diff
@@ -1,120 +1,190 @@
...
+    # test the data loaded from the file is the same as original data
+    datafile_df = datafile_df.sort_values(by=[DEFAULT_USER_COL, DEFAULT_ITEM_COL])
+    train = train.sort_values(by=[DEFAULT_USER_COL, DEFAULT_ITEM_COL])
+    train[DEFAULT_RATING_COL] = train[DEFAULT_RATING_COL].apply(lambda x: float(x > 0))
+    train = train.drop(DEFAULT_TIMESTAMP_COL, axis=1)
+    assert train.equals(datafile_df)
+
+    # test data can be loaded for a valid user and it throws exception for invalid user
+    user = train[DEFAULT_USER_COL].iloc[0]
+    missing_user = train[DEFAULT_USER_COL].iloc[0] + 1
```
anargyri, line 57: Why is this user missing? Isn't it inside the valid range of user ids (1 - 100)?

tests/unit/recommenders/models/test_ncf_dataset.py
```diff
@@ -1,120 +1,190 @@
...
+    with datafile as f:
+        user_data = f.load_data(user)
+        assert user_data[DEFAULT_USER_COL].iloc[0] == user
+        with pytest.raises(Exception):
```
anargyri, line 61: Maybe the exceptions from NCF Dataset should be more specific e.g. ValueError, see https://docs.python.org/3.8/library/exceptions.html
What do you think?

tests/unit/recommenders/models/test_ncf_dataset.py
```diff
@@ -1,120 +1,190 @@
...
+        batch_data = {
+            DEFAULT_USER_COL: [dataset.id2user[user] for user in batch[0]],
+            DEFAULT_ITEM_COL: [dataset.id2item[item] for item in batch[1]],
+            DEFAULT_RATING_COL: batch[2]
+        }
+        batch_records.append(pd.DataFrame(batch_data))
+    
+    assert len(batch_records) == expected_batches
+    train_loader_df = pd.concat(batch_records).reset_index(drop=True)
+    assert train_loader_df.shape[0] == expected_batches * batch_size
```
anargyri, line 151: Isn't it possible that the last batch will be smaller than batch_size and hence that this assertion is false?

tests/unit/recommenders/models/test_ncf_dataset.py
```diff
@@ -1,120 +1,190 @@
...
+    test_users = leave_one_out_test[DEFAULT_USER_COL].unique()
+
+    n_neg = 1
+    n_neg_test = 1
+    dataset = Dataset(train_path, test_file=leave_one_out_test_path, n_neg=n_neg, n_neg_test=n_neg_test)
+    assert set(dataset.test_full_datafile.users) == set(test_users)
+
+    # test number of batches and data size is as expected after loading all test data
+    expected_test_batches = leave_one_out_test.shape[0]
+    assert max([x for x in dataset.test_full_datafile.batch_indices_range]) + 1 == expected_test_batches
```
anargyri, line 174: I think `max(dataset.test_full_datafile.batch_indices_range)` would also work.

anargyri (approved):
Great job @angusrtaylor ! the code is really neat now.

tests/unit/recommenders/models/test_ncf_dataset.py
angusrtaylor: Yes you're right... I was being lazy. I'll make them more specific.

tests/unit/recommenders/models/test_ncf_dataset.py
angusrtaylor: I'll check this.

tests/conftest.py
```diff
@@ -392,6 +394,74 @@ def random_date_generator(start_date, range_in_days):
...
+    test.to_csv(test_path, index=False)
+    leave_one_out_test.to_csv(leave_one_out_test_path, index=False)
+    return train_path, test_path, leave_one_out_test_path
+
+
+@pytest.fixture
+def dataset_ncf_files_unsorted(data_paths, dataset_ncf_files):
+    train_path, test_path, leave_one_out_test_path = data_paths
+    train, test, leave_one_out_test = dataset_ncf_files
+    train = train.sort_values(by=DEFAULT_ITEM_COL)
```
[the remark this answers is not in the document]
angusrtaylor: Because if we leave the data frame as is there is an infinitesimally small chance the data is sampled in user order. Sorting by item guarantees it's not sorted by user. But I can remove this if you think it just adds confusion?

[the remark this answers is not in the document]
anargyri: Ah, it's a bit unclear. Maybe make it more explicit e.g. move the highest user to the first position.

anargyri:
There was some error in the notebook, see here https://dev.azure.com/best-practices/recommenders/_build/results?buildId=56649&view=ms.vss-test-web.build-test-results-tab&runId=53844&resultId=100002&paneView=debug
It looks like it's in the construction of the Dataset.

anargyri (approved):

angusrtaylor:
> There was some error in the notebook, see here https://dev.azure.com/best-practices/recommenders/_build/results?buildId=56649&view=ms.vss-test-web.build-test-results-tab&runId=53844&resultId=100002&paneView=debug It looks like it's in the construction of the Dataset.

Thanks. I saw that this morning but couldn't make sense of it. In my PR, the constructor is different (`data = NCFDataset(train_file=train_file, test_file=leave_one_out_test_file, seed=SEED)` in my branch vs `data = NCFDataset(train=train, test=test, seed=SEED)` in staging). So it seems the build is running the version of the notebook that's in staging rather than my branch.

anargyri:
> > There was some error in the notebook, see here https://dev.azure.com/best-practices/recommenders/_build/results?buildId=56649&view=ms.vss-test-web.build-test-results-tab&runId=53844&resultId=100002&paneView=debug It looks like it's in the construction of the Dataset.
> 
> Thanks. I saw that this morning but couldn't make sense of it. In my PR, the constructor is different (`data = NCFDataset(train_file=train_file, test_file=leave_one_out_test_file, seed=SEED)` in my branch vs `data = NCFDataset(train=train, test=test, seed=SEED)` in staging). So it seems the build is running the version of the notebook that's in staging rather than my branch.

Strange. I am rerunning the job.

miguelgfierro:
there is a weird OOM error:
```
E           /home/runner/work/recommenders/recommenders/.tox/gpu/lib/python3.7/site-packages/torch/utils/data/dataloader.py in _next_data(self)
E               561         data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
E               562         if self._pin_memory:
E           --> 563             data = _utils.pin_memory.pin_memory(data)
E               564         return data
E               565 
E           
E           /home/runner/work/recommenders/recommenders/.tox/gpu/lib/python3.7/site-packages/torch/utils/data/_utils/pin_memory.py in pin_memory(data)
E                56         return type(data)(*(pin_memory(sample) for sample in data))
E                57     elif isinstance(data, collections.abc.Sequence):
E           ---> 58         return [pin_memory(sample) for sample in data]
E                59     elif hasattr(data, "pin_memory"):
E                60         return data.pin_memory()
E           
E           /home/runner/work/recommenders/recommenders/.tox/gpu/lib/python3.7/site-packages/torch/utils/data/_utils/pin_memory.py in <listcomp>(.0)
E                56         return type(data)(*(pin_memory(sample) for sample in data))
E                57     elif isinstance(data, collections.abc.Sequence):
E           ---> 58         return [pin_memory(sample) for sample in data]
E                59     elif hasattr(data, "pin_memory"):
E                60         return data.pin_memory()
E           
E           /home/runner/work/recommenders/recommenders/.tox/gpu/lib/python3.7/site-packages/torch/utils/data/_utils/pin_memory.py in pin_memory(data)
E                56         return type(data)(*(pin_memory(sample) for sample in data))
E                57     elif isinstance(data, collections.abc.Sequence):
E           ---> 58         return [pin_memory(sample) for sample in data]
E                59     elif hasattr(data, "pin_memory"):
E                60         return data.pin_memory()
E           
E           /home/runner/work/recommenders/recommenders/.tox/gpu/lib/python3.7/site-packages/torch/utils/data/_utils/pin_memory.py in <listcomp>(.0)
E                56         return type(data)(*(pin_memory(sample) for sample in data))
E                57     elif isinstance(data, collections.abc.Sequence):
E           ---> 58         return [pin_memory(sample) for sample in data]
E                59     elif hasattr(data, "pin_memory"):
E                60         return data.pin_memory()
E           
E           /home/runner/work/recommenders/recommenders/.tox/gpu/lib/python3.7/site-packages/torch/utils/data/_utils/pin_memory.py in pin_memory(data)
E                48 def pin_memory(data):
E                49     if isinstance(data, torch.Tensor):
E           ---> 50         return data.pin_memory()
E                51     elif isinstance(data, string_classes):
E                52         return data
E           
E           RuntimeError: cuda runtime error (2) : out of memory at ../aten/src/THC/THCCachingHostAllocator.cpp:280
```

angusrtaylor:
> cuda runtime error (2)

There are a bunch of tests that are failing in the build, even after merging in staging. I'm still trying to make sense of it. @anargyri's latest PR build was successful on staging so I'm trying to work out what is different compared to that...
image

anargyri:
> > cuda runtime error (2)
> 
> There are a bunch of tests that are failing in the build, even after merging in staging. I'm still trying to make sense of it. @anargyri's latest PR build was successful on staging so I'm trying to work out what is different compared to that... image

It could be something with the host. I rebooted and restarted the pipeline.

angusrtaylor:
I figured out the problem with the NCF test: the failure was related to the ncf quickstart notebook which I had forgotten to update. I've updated it and pushed the changes. I think all the other failures are OOM errors.

Issue #1510, opened 2021-08-26
[FEATURE] Export RBM model


miguelgfierro:
### Description

Problem when saving RBM model, following the Wide&Deep model implementation https://nbviewer.jupyter.org/github/microsoft/recommenders/blob/staging/examples/00_quick_start/wide_deep_movielens.ipynb#4.3-Export-Model but when I exported the RBM model I received the following error

`AttributeError: 'RBM' object has no attribute 'experimental_export_all_saved_models'`

### In which platform does it happen?

Linux VM

### How do we replicate the issue?

```
train_fn = tf_utils.pandas_input_fn(
    df=train_df,
    y_col='rating',
    batch_size=100,
    num_epochs=None,  .
    shuffle=True,
    seed=42,
)
# Unique users in the dataset
user_ids = tf.feature_column.categorical_column_with_vocabulary_list("userID", user["userID"].values)
item_ids = tf.feature_column.categorical_column_with_vocabulary_list("movieID", item["movieID"].values)

fn = [
        user_ids,
        item_ids,
        tf.feature_column.crossed_column(
                [user_ids, item_ids], hash_bucket_size=1000
        ),
    ]

exported_path = tf_utils.export_model(
    model=model,
    train_input_fn=train_fn,
    eval_input_fn=tf_utils.pandas_input_fn(
        df=test_df, y_col='rating'
    ),
    tf_feat_cols= fn
    base_dir=EXPORT_DIR_BASE,
)
```

### Other Comments
related to #878

miguelgfierro:
FYI @pradnyeshjoshi

miguelgfierro:
fixed in #1618

Issue #878, opened 2019-07-26
[FEATURE] Model save in RBM


ghost:
### Description
Add feature to allow the RBM model to save itself to disk.

### Expected behavior with the suggested feature
RBM encapsulates the tensorflow details within its implementation. It would be nice to have the code to save the trained model to disk so that tensorflow.utils and other modules do not have to be imported into the scripts that create the RBM instance.

something like:

rbm_model = RBM(.....)
train_time = rbm_model.fit(...)
rbm_model.save(DIR, overwrite=True)

atimesastudios:
Bumping my original post - is there any way to save the RBM model similar to the implementation of Wide Deep? If someone could provide some tips on how to get started, it would be much appreciated.

atimesastudios:
@WessZumino - I've implemented Azure HyperDrive hyperparameter optimization for RBM using the Surprise and Wide Deep hyperdrive notebook examples. I'm looking for tips on how to save the best model so that I can use it later for scoring. Any guidance would be much appreciated

mtayyab-numtra:
Did you find any way to export the RBM model?

WessZumino:
Hi @mtayyab-numtra, I am not actively working on this anymore. Weights can be saved and loaded as per any other tensorflow model, if this is what you are asking. If you mean saving weights when using AML or AML's Hyperdrive, the issue is the same. You need to create a script following examples from the Wide & Deep model, and add the save weight method. Once hyperdrive is done, you will see the plot of the metric(s) you chose and you can select the one with the best performance. For each model, AML will save the weights in a directory of your choice. Than you need to retrieve the weights for the model you selected (again following the example for Wide & Deep) and load them in the RBM model. I believe you need to add a method for this, but this is also standard. Let me know if this helps.

miguelgfierro:
fixed in #1618

PR #1618, opened 2022-01-20
Model save in RBM

pradnyeshjoshi:
### Description

1. Added `save` function to `rbm.py`: This writes model checkpoints to a specified directory.
2. Added `load` function to `rbm.py`: This gives users ability to instantiate a model class, load checkpoints from a pretrained model and resume training.
3. Added code snippet for how to use `save` and `load` functions to the RBM quick start notebook.

### Related Issues

1. https://github.com/microsoft/recommenders/issues/878
2. https://github.com/microsoft/recommenders/issues/1510

### Checklist:


examples/00_quick_start/models/checkpoint
```diff
@@ -0,0 +1,2 @@
+model_checkpoint_path: "rbm_model.ckpt"
+all_model_checkpoint_paths: "rbm_model.ckpt"
```
miguelgfierro, line 2: is this needed, or can we remove it?

recommenders/models/rbm/rbm.py
```diff
@@ -691,3 +699,27 @@ def predict(self, x):
         vp = self.sess.run(v_, feed_dict={self.vu: x})
 
         return vp
+
+    def save(self, dir_name):
+        """Save model parameters in `dir_name`
+        Args:
```
miguelgfierro, line 705: Minor change, please add an intro between the first line and Args, otherwise sphinx will throw an error and the documentation won't compile:
```
"""Save model parameters in `dir_name`

        Args:
             dir_name (str): directory name, which should be a folder name instead of file name
                we will create a new directory if not existing.
"""
```

recommenders/models/rbm/rbm.py
```diff
@@ -691,3 +699,27 @@ def predict(self, x):
         vp = self.sess.run(v_, feed_dict={self.vu: x})
 
         return vp
+
+    def save(self, dir_name):
```
miguelgfierro, line 703: minor change, to give more flexibility to the user we can add a model name parameter:
```
def save(self, dir_name, model_name=MODEL_CHECKPOINT):
```

tests/unit/recommenders/models/test_rbm.py
```diff
@@ -106,7 +119,7 @@ def check_sampled_values(sampled, s):
     r = Xtr.max()  # obtain the rating scale
 
     # fit the model to the data
-    model.fit(Xtr, Xtst)
+    model.fit(Xtr)
 
     # evaluate the activation probabilities of the hidden units and their sampled values
```
miguelgfierro, line 124: can you add a test for the save and load, similar to `def test_class_init(init_rbm):` but create the class, save it, loading and then make sure that you get the right:
```
    # number of hidden units
    assert model.n_hidden == init_rbm["n_hidden"]
    # number of training epochs
    assert model.epochs == init_rbm["epochs"] + 1
    # minibatch size
    assert model.minibatch == init_rbm["minibatch"]
    # keep probability for dropout regulrization
    assert model.keep == init_rbm["keep_prob"]
    # learning rate
    assert model.learning_rate == init_rbm["learning_rate"]
    # standard deviation used to initialize the weight matrix from a normal distribution
    assert model.stdv == init_rbm["init_stdv"]
    # sampling protocol used to increase the number of steps in Gibbs sampling
    assert model.sampling_protocol == init_rbm["sampling_protocol"]
    # number of epochs after which the rmse is displayed
    assert model.display_epoch == init_rbm["display_epoch"]
```

recommenders/models/rbm/rbm.py
```diff
@@ -691,3 +699,27 @@ def predict(self, x):
...
+        """Save model parameters in `dir_name`
+        Args:
+            dir_name (str): directory name, which should be a folder name instead of file name
+                we will create a new directory if not existing.
+        """
+        # save trained model
+        if not os.path.exists(dir_name):
+            os.makedirs(dir_name)
```
miguelgfierro, line 711: Minor proposal, to me this is more elegant: `os.makedirs(work_directory, exist_ok=True)` and you save one line of code

miguelgfierro (commented):
this is amazing @pradnyeshjoshi, brutal timing!

Just a minor comment with tests

recommenders/models/rbm/rbm.py
pradnyeshjoshi: Added a `file_path` parameter instead of the two. If the parent directory in the file path does not exist, it will be created.

recommenders/models/rbm/rbm.py
pradnyeshjoshi: Done.

examples/00_quick_start/models/checkpoint
pradnyeshjoshi: Removed.

tests/unit/recommenders/models/test_rbm.py
pradnyeshjoshi: Done.

recommenders/models/rbm/rbm.py
pradnyeshjoshi: Done.

miguelgfierro (approved):
Vamos!!!

recommenders/models/rbm/rbm.py
```diff
@@ -14,6 +16,9 @@ class RBM:
 
     def __init__(
         self,
+        n_users,
```
anargyri, line 19: Maybe you could avoid `n_users` in the constructor of the model. As far as I can see it is used only inside `fit()` for the minibatch, isn't it? Is there a scenario in which `n_users` should be other than the number of users in the training data? If not, then you could remove `n_users` from the constructor and compute the minibatch inside `fit()`. 
This would also make the model simpler to handle for the end user, since now if you want to retrain with more data you need to do something like
```
model = RBM(n_users=10000)
model.fit(Xtr)
Xtr2 = pd.concat([Xtr, Xtr_new]) 
model = RBM(n_users=15000)
model.fit(Xtr2)
```
Without `n_users` in `__init()__` you would avoid the second call to `__init()__`. It may also cause some problems with save and load.

miguelgfierro: @pradnyeshjoshi FYI in case you didn't see this

pradnyeshjoshi: @miguelgfierro I have created a [PR](https://github.com/microsoft/recommenders/pull/1622) to remove `n_users` from the constructor. Looks like Andreas has approved it, you may merge it into staging if all looks good.

PR #1637, opened 2022-02-10
Scenario gaming

miguelgfierro:
### Description

Add scenario for gaming industry

### Checklist:


scenarios/gaming/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Similar items
+
+A related scenario to personalized item recommendation is similar item recommendations. In this case the recommendations are based on item feature similarity. Some features that can be taken into account are price, item category, item usage, etc. The business usecase for similar items can be increasing the likelihood of buying or avoiding losing a sale by showing an alternative item with a lower price (down-selling). The set of algorithms that can be used in this scenario are those based on content-based filtering, such as [LightGBM](../../examples/00_quick_start/lightgbm_tinycriteo.ipynb), [VW](../../examples/02_model_content_based_filtering/vowpal_wabbit_deep_dive.ipynb) or even [xDeepFM](../../examples/00_quick_start/xdeepfm_criteo.ipynb).
```
angusrtaylor, line 19:
```suggestion
A related scenario to personalized item recommendation is similar item recommendations. In this case the recommendations are based on item feature similarity. Some features that can be taken into account are price, item category, item usage, etc. The business usecase for similar items can to increase the likelihood of buying or avoiding losing a sale by showing an alternative item with a lower price (down-selling). The set of algorithms that can be used in this scenario are those based on content-based filtering, such as [LightGBM](../../examples/00_quick_start/lightgbm_tinycriteo.ipynb), [VW](../../examples/02_model_content_based_filtering/vowpal_wabbit_deep_dive.ipynb) or even [xDeepFM](../../examples/00_quick_start/xdeepfm_criteo.ipynb).

```

scenarios/gaming/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Next best action prediction
+
+An interesting scenario is next best action prediction. In this scenario, what is recommended is the most beneficial next action for the player. From the technical point of view, this can be implemented using collaborative filtering algorithms, such as [SAR](../../examples/00_quick_start/sar_movielens.ipynb), [BPR](../../examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb), and [NCF](../../examples/00_quick_start/ncf_movielens.ipynb).
```
anargyri, line 23: Isn't this task closer to reinforcement learning rather than CF?

scenarios/gaming/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Data and evaluation
+
+Datasets used in gaming recommendations usually include [user information](../../GLOSSARY.md), [item information](../../GLOSSARY.md) and [interaction data](../../GLOSSARY.md), among others.
```
anargyri, line 27: Is any context based information (e.g. session) available?

scenarios/gaming/README.md
miguelgfierro: Good point, there is a way to do next best action in RL, a lot of people using MCTS. This is done in AlphaGo, AlphaZero, etc. But in that case, the selection of the optimal next action is automatically taken. If we pose the problem as a reco system, where actions are treated like items in the typical CF setup, then the user can choose among a set of actions.

scenarios/gaming/README.md
miguelgfierro: That's a good point. @angusrtaylor do you know if there is session data in the current engagement you are looking at? I'll ask internally as well

angusrtaylor: There is data on when players logged in, added/removed an item to their inventory, matched with a multiplayer game etc.

scenarios/gaming/README.md
anargyri: I meant something about the state i.e. in CF there is no state. But here is the current state of the game relevant for which next actions to recommend? "Next" implies that you have at least some info about what has happened before. With CF you would always recommend the same action, wouldn't you?

scenarios/gaming/README.md
miguelgfierro: updated

scenarios/gaming/README.md
miguelgfierro: The way I see it is like a session based recommender, but with actions instead of items. The info that has happened before are the previous actions, the next recommended action is an option that the user chooses among a list of actions available.

miguelgfierro:
I'll merge this now, we can iterate over the content if needed

PR #1663, opened 2022-03-08
Danb27/sar literal fix

danb27:
### Description
Change how the package checks for similarity type from 'is' to ==
This is more correct python code - check for the same value in the string using ==, avoids problems when unpacking parameters from other files using json package.

### Checklist:

### DIFF
--- a/tests/unit/recommenders/models/test_sar_singlenode.py
+++ b/tests/unit/recommenders/models/test_sar_singlenode.py
@@ -392,9 +392,9 @@ def test_get_normalized_scores(header):
     assert np.isclose(expected, np.asarray(actual)).all()
 
 
-def test_match_similarity_type(header):
+def test_match_similarity_type_from_json_file(header):
     # store parameters in json
-    params_str = json.dumps({'similarity_type': 'jaccard'})
+    params_str = json.dumps({'similarity_type': 'lift'})
     # load parameters in json
     params = json.loads(params_str)
 
@@ -411,4 +411,5 @@ def test_match_similarity_type(header):
         }
     )
 
+    # make sure fit still works when similarity type is loaded from a json file
     model.fit(train)
--- a/tests/unit/recommenders/models/test_sar_singlenode.py
+++ b/tests/unit/recommenders/models/test_sar_singlenode.py
@@ -394,11 +394,9 @@ def test_get_normalized_scores(header):
 
 def test_match_similarity_type(header):
     # store parameters in json
-    with open('similarity_type_test.json', 'w') as f:
-        json.dump({'similarity_type': 'jaccard'}, f)
+    params_str = json.dumps({'similarity_type': 'jaccard'})
     # load parameters in json
-    with open('similarity_type_test.json') as f:
-        params = json.load(f)
+    params = json.loads(params_str)
 
     params.update(header)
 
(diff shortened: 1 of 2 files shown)


danb27:
More information on the problem with using 'is': https://lerner.co.il/2015/06/16/why-you-should-almost-never-use-is-in-python/

gramhagen (approved):
good catch!

tests/unit/recommenders/models/test_sar_singlenode.py
```diff
@@ -389,3 +390,25 @@ def test_get_normalized_scores(header):
...
+    train = pd.DataFrame(
+        {
+            header["col_user"]: [1, 1, 1, 1, 2, 2, 2, 2],
+            header["col_item"]: [1, 2, 3, 4, 1, 5, 6, 7],
+            header["col_rating"]: [3.0, 4.0, 5.0, 4.0, 3.0, 2.0, 1.0, 5.0],
+            header["col_timestamp"]: [1, 20, 30, 400, 50, 60, 70, 800],
+        }
+    )
+
+    model.fit(train)
```
miguelgfierro, line 414: thanks for the input @danb27, not sure I understand what you are trying to check with this test. We already have a test that checks the fit function: https://github.com/microsoft/recommenders/blob/staging/tests/unit/recommenders/models/test_sar_singlenode.py#L92
There is another test that checks the init: https://github.com/microsoft/recommenders/blob/staging/tests/unit/recommenders/models/test_sar_singlenode.py#L74, however, it only checks it with the jaccard option. Maybe a good addition would be to add checks around lift and concurrence.

What do you think?

danb27: @miguelgfierro This test isn't really testing the init or the fit method (sort of - it is really just testing 'is' vs. == in the fit function). This test would fail currently because the string comparison is done with 'is' instead of ==. Currently, if I do `SARSingleNode(similarity_type='lift', **other_params)` I will be able to fit successfully. BUT if I load in the similarity type from a JSON file, the fit method will give this error: 

https://github.com/microsoft/recommenders/blob/staging/recommenders/models/sar/sar_singlenode.py#L287 

even though I have spelled lift/jaccard/cooccurrence correctly. 

The purpose of this test is to ensure that this error is not raised regardless of whether the similarity type is provided directly in your code or if you are loading the desired similarity type from a json file (assuming you selected a valid similarity type)

danb27: I changed the name of the test, added a comment to help clarify the purpose, and changed the similarity type used to test the functionality to 'lift' instead of 'jaccard' (just to test something other than jaccard in this one). 

Does that help clarify?

miguelgfierro: this is fantastic Dan, I'll approve the PR

miguelgfierro (approved):
Great work!

miguelgfierro:
hey @danb27 I have just realized that you haven't signed your commits, so your name won't appear on GitHub. If you want to keep contributing to the repo and that your name appears on it, you can sign your commits following: https://github.com/microsoft/recommenders/wiki/How-to-add-your-name-as-a-contributor-to-the-repo#make-sure-that-your-github-user-is-setup-correctly
```
git config --global user.name "Your Name"
git config --global user.email your-github-username@users.noreply.github.com
```

PR #1665, opened 2022-03-09
Codecov Fix

laserprec:
### Description
Resolves the following issue when merging code coverage reports generated in self-hosted (aml) runners.

```
"CoverageWarning: Couldn't parse 
 '/home/azureuser/runner/work/recommenders/recommenders/recommenders/evaluation/__init__.py': No source for code"
```
Also add `-i` flag to ignore the "No source for code" error.

### Checklist:

### DIFF
--- a/.github/workflows/actions/merge-cov/action.yml
+++ b/.github/workflows/actions/merge-cov/action.yml
@@ -33,9 +33,18 @@ runs:
   # different parallelized runs
   - name: Merge coverage reports
     shell: bash
+    # NOTE: Merging code reports generated from self-hosted (aml) runners can be problematic
+    #       as reports may reference the source code at an unresolvable absolute path.
+    #       For example, you may encounter errors like:
+    #            "CoverageWarning: Couldn't parse 
+    #            '/home/azureuser/runner/work/recommenders/recommenders/recommenders/evaluation/__init__.py':
+    #            No source for code"
+    # Work-around: Creating a symlink at the root that points to the default local runner folder: '/home/runner/work/recommenders'
     run: |
+      sudo mkdir -p /home/azureuser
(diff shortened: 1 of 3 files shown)


.github/workflows/actions/merge-cov/action.yml
```diff
@@ -33,9 +33,18 @@ runs:
...
+    # NOTE: Merging code reports generated from self-hosted (aml) runners can be problematic
+    #       as reports may reference the source code at an unresolvable absolute path.
+    #       For example, you may encounter errors like:
+    #            "CoverageWarning: Couldn't parse 
+    #            '/home/azureuser/runner/work/recommenders/recommenders/recommenders/evaluation/__init__.py':
+    #            No source for code"
+    # Work-around: Creating a symlink at the root that points to the default local runner folder: '/home/runner/work/recommenders'
     run: |
+      sudo mkdir -p /home/azureuser
+      sudo ln -s /home/runner /home/azureuser/runner
```
anargyri, line 45: /home/runner may not exist, don't you need to create it first?

laserprec: `/home/runner` is a default path that should be present as long as we are using github-hosted runners to run this step. If that path doesn't exist, it will be a different but big problem (a lot of things can break for github action alone). Of course, there's no guarantee that this wouldn't change, but I think this is relatively safe assumption to make it simpler, unless you think otherwise.

anargyri: Ok, let's see if it works.

laserprec: hmm but we get 23% coverage, did we change the sets of test we run?

anargyri (approved):

PR #1588, opened 2021-12-17
Remove drop_duplicates() from SAR method fix #1464

miguelgfierro:
### Related Issues

Fix #1464

### Checklist:

- [ ] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions.
- [ ] I have updated the documentation accordingly.
- [ ] This PR is being made to `staging branch` and not to `main branch`.

PR #1666, opened 2022-03-09
danb27/sar_single_node_improvements: allow for getting most frequent …

danb27:
### Description
1) Bug fix: The documentation says that `self.item_frequencies` should contain the frequencies of every item, but in fact, it currently shows a different value, generated by using the diagonal of the response from `self.compute_cooccurrence_matrix`, which is not actually equal to item frequencies when there are duplicates in the dataframe. The test we had written before (`test_get_popularity_based_topk`) was only passing because the test it has no duplicates. If you add any duplicates to the test, it will fail, and nothing in the fit() method will stop you or warn you that you are using the object incorrectly. Many of the tests we have written do include duplicates also.
2) Improvement: Allow users to get similar users using the new method, `self.get_user_based_topk`, similarly to how users are currently able to get similar items using `self.get_item_based_topk`
3) Improvement: Allow users to use the `self.get_popularity_based_topk` method with `items=False` to get the most popular users rather than the most popular items. The parameter defaults to True (return items, not users) for backwards compatibility
4) Add tests for 1-3

### Checklist:

Looking forward to your feedback!


danb27:
the issue (I'm no longer sure that it is a bug) with the frequencies was introduced here: #https://github.com/microsoft/recommenders/pull/1588 

The problem happens when there are duplicate user-item pairs in the dataset. A note was added to the docstring telling the user not to provide duplicates, but there is no warning or error thrown if duplicates are present. If there were an error thrown, then a lot of our current tests would fail also. 

Without any warning or without dropping duplicates, you can get weird results like this: 

image

I added a warning if there are duplicates found in the dataset, but I think we can still improve by changing how we calculate frequency to be unrelated to the cooccurrence matrix (just count occurrences in the user provided dataframe). 

Any thoughts?

recommenders/models/sar/sar_singlenode.py
```diff
@@ -446,6 +490,80 @@ def get_item_based_topk(self, items, top_k=10, sort_top_k=True):
...
+        based on the similarity matrix fit during training.
+
+        Args:
+            users (pandas.DataFrame): DataFrame with user, item (optional), and rating (optional) columns
+            top_k (int): number of top users to recommend
+            sort_top_k (bool): flag to sort top k results
+
+        Returns:
+            pandas.DataFrame: similar users to the users provided in users
+        """
```
miguelgfierro, line 505: One question @danb27, the computation above of the user coocurrence is only used when we want to use the method `get_user_based_topk`?

miguelgfierro (commented):
@danb27 this is fantastic work, I added some questions and suggestions. I'll add more people to review because the work you are doing is super good.

Something I would like to ask Dan, please be patient with the review process, SAR is a key algo for us that has a lot of implications, so it will take us some time to review

recommenders/models/sar/sar_singlenode.py
```diff
@@ -192,27 +203,32 @@ def set_index(self, df):
...
-        Please make sure that `df` has no duplicates.
+        Please make sure that `df` has no duplicate user-item pairs.
 
         Args:
-            df (pandas.DataFrame): User item rating dataframe (without duplicates).
+            df (pandas.DataFrame): User item rating dataframe (without duplicate user-items).
+            compute_user_similarity (bool): If true, compute user similarity using the self.similarity_type. Item
+            similarity is always calculated. This must be true to use self.get_user_based_topk to get similar users.
         """
+        if sum(df[[self.col_user, self.col_item]].duplicated()) > 0:
```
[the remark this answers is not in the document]
danb27: I think either option is better than just having a docstring. If there is a use case for running SAR with duplicates, then I think warning is best. But, if we believe that the algorithm should NEVER see duplicates, then I would choose error. 

If we do choose to go with error, we will need to fix multiple tests. If we go with warning, we don't need to fix the tests (but still probably should IMO). I was just hesitant to change a bunch of tests without first having this discussion. 

Let me know what you all think.

recommenders/models/sar/sar_singlenode.py
danb27: correct. in `fit()`, if we set `compute_user_similarity=True`, `user_cooccurrence` will be calculated and used to create `self.user_similarity`, which is required to run `get_user_based_topk`. 

I went back and forth on whether to compute user similarity always, but decided to create the `compute_user_similarity` parameter so that the user could keep performance the same as before the PR if they do not require the added functionality. 

Happy to have my mind changed if you all would prefer to have this computation occur every time so that you can always call any of the methods, regardless of how you fit.

recommenders/models/sar/sar_singlenode.py
```diff
@@ -157,7 +161,7 @@ def compute_time_decay(self, df, decay_column):
         # group time decayed ratings by user-item and take the sum as the user-item affinity
         return df.groupby([self.col_user, self.col_item]).sum().reset_index()
 
-    def compute_cooccurrence_matrix(self, df):
+    def compute_cooccurrence_matrix(self, df, reverse=False):
```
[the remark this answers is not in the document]
danb27: I agree with your critique. In hindsight, reverse was not a clear name for this. Personally, I prefer booleans because they leave less room for error.

recommenders/models/sar/sar_singlenode.py
miguelgfierro: what would be the change required and the performance difference if we do all the computation of user_coocurrence inside the function `get_user_based_topk`?

tests/unit/recommenders/models/test_sar_singlenode.py
```diff
@@ -413,3 +418,41 @@ def test_match_similarity_type_from_json_file(header):
 
     # make sure fit still works when similarity type is loaded from a json file
     model.fit(train)
+
+def test_get_user_and_item_based_topk(header):
```
miguelgfierro, line 422: minor comment, typically we try to test each function separately, so we would have `test_get_user_based_topk` and `test_get_item_based_topk`, then it is easier for us to identify which test is failing

recommenders/models/sar/sar_singlenode.py
miguelgfierro: > I am envisioning the use case (which is my current use case) where someone wants to store similar users for every user in their dataset, so it is necessary to call this method n_users times.

Ideally, it would be great that given a list of user_ids (or DF), the system returns a DF (or list of list, whatever) but the top k most similar users for each input user_id in the list. It looks that is not happening, right?

Some questions:
1. How can we create this result?
2. Right now we can input a list of user_ids, what is the meaning of the output?

tests/unit/recommenders/models/test_sar_singlenode.py
```diff
@@ -413,3 +418,41 @@ def test_match_similarity_type_from_json_file(header):
...
+            header["col_rating"]: [3.0, 4.0, 5.0, 4.0, 3.0, 2.0, 1.0, 5.0, 1.0, 1.0]
+        }
+    )
+    model.fit(train, compute_user_similarity=True)
+
+    similar_users = model.get_user_based_topk(pd.DataFrame({header['col_user']: [1]}))
+    assert similar_users[header['col_user']].iloc[0] == 4
+
+    similar_users = model.get_user_based_topk(pd.DataFrame({header['col_user']: [1, 2]}))
+    assert similar_users[header['col_user']].iloc[0] == 4
```
miguelgfierro, line 437: I've been doing some debugging, if I do:
```
train = pd.DataFrame(
        {
            header["col_user"]: [1, 1, 2, 2, 3, 3, 3, 3, 4, 4],
            header["col_item"]: [1, 2, 1, 3, 3, 4, 5, 6, 1, 2],
            header["col_rating"]: [3.0, 4.0, 5.0, 4.0, 3.0, 2.0, 1.0, 5.0, 1.0, 1.0]
        }
    )
```
I get:
```
	UserId	MovieId	prediction
0	4	0	1.000000
1	2	0	0.333333
2	3	0	0.000000
```
but if I do the following, making the user 1 and 2 the same:
```
train = pd.DataFrame(
        {
            header["col_user"]: [1, 1, 2, 2, 3, 3, 3, 3, 4, 4],
            #header["col_item"]: [1, 2, 1, 3, 3, 4, 5, 6, 1, 2],
            #header["col_rating"]: [3.0, 4.0, 5.0, 4.0, 3.0, 2.0, 1.0, 5.0, 1.0, 1.0]
            header["col_item"]: [1, 2, 1, 2, 3, 4, 5, 6, 1, 2],
            header["col_rating"]: [3.0, 4.0, 3.0, 4.0, 3.0, 2.0, 1.0, 5.0, 5.0, 1.0]
        }
    )
```
I get:
UserId | MovieId | prediction
-- | -- | --
2 | 0 | 1.0
4 | 0 | 1.0
3 | 0 | 0.0

But user 2 is equal to user 1, and it gets a prediction of 1, which is the same as user 4, that has seen the same movies as user 1, but rated them completely different. User 2 should get a higher score than user 4 in this case. Do you know why this is happening?

recommenders/models/sar/sar_singlenode.py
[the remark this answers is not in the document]
danb27: So currently, yes duplicates will give us weird results in the cooccurrence matrix. We can fix this by simply removing duplicates from a copy of the dataframe before calculating cooccurrence matrix like this:

recommenders/models/sar/sar_singlenode.py
danb27: 1)

Currently, I call `get_item_based_topk()` like this: 

```
for item_id in item_ids:
    temp_df = pd.DataFrame({'item_id': [item]}, dtype=DTYPE)
    similar_items_df = model.get_item_based_topk(temp_df, k=K)
    ...
```
With my PR, I would be able to do something very similar for `get_user_based_topk()` 

We could implement this as a different method, because what currently occurs if you give this method a list of user ids is the following:

2. From the docstring of `get_item_based_topk()`: " This method will take a set of items and use them to recommend the most similar items to that set based on the similarity matrix fit during training. This allows recommendations for cold-users (unseen during training), note - the model is not updated." The same is true for the new `get_user_based_topk()` method, it returns the most similar users given a set of other users.

tests/unit/recommenders/models/test_sar_singlenode.py
danb27: I believe this is because user similarity is calculated from user cooccurrence, which does not actually look at the ratings.  The same is true for items, so I believe you would see the same issue with items.

danb27:
@miguelgfierro Sorry for the back and forth, but your comments made me realize a much simpler + better way to achieve what I was trying to do. Here is a summary of what the PR does now: 

- change how we calculate user + item frequencies to come directly from the user's data
- remove need for user to drop duplicates ahead of calling .fit(). The issue was when we calculated item cooccurrences and there were duplicates, but we can instead just temporarily drop any duplicates in the user dataframe before calculating cooccurrences. this ensures that our unit tests are still working AND our cooccurrences are accurate
- allow someone to retrieve the most frequent users, the same way we can already retrieve the most popular items
- allow someone to retrieve the k most similar users based on our user affinity matrix. This allows ratings to be factored into the results of grabbing the most similar users and also avoids the weird situation I had created where we needed to calculate user cooccurrences if we wanted to call a specific method. 

I think these changes are a lot better and hopefully make the PR more palatable. 

What do you think?

Is there any reason we should still tell users to not include duplicates? I am assuming that if our unit test dataset has duplicates, then we should also be accepting duplicates, but I don't know where this test data comes from.

danb27:
@anargyri Thank you for your clarification on point 1. This is very helpful. I will think about if I can adapt my code or if I should just change those parts back to how they were before. 

2. Of course this could be misused if you are leaking information to other users. But I believe there are legitimate use cases also. For example, you want to make sure that your model has some overlap in the recommendations for similar users, but that similar users are not getting the exact same recommendations. Please note that this sort of functionality is featured within recommenders for LightFM, see the [LightFM notebook](https://github.com/microsoft/recommenders/blob/main/examples/02_model_hybrid/lightfm_deep_dive.ipynb) and the [code](https://github.com/microsoft/recommenders/blob/main/recommenders/models/lightfm/lightfm_utils.py#L128)

3. You might want to look at how your model performs for users with a lot of data points vs. very few to understand your model biases and at what number of datapoints you know enough about a user to make good recommendations.

tests/unit/recommenders/models/test_sar_singlenode.py
```diff
@@ -413,3 +418,37 @@ def test_match_similarity_type_from_json_file(header):
...
+    train = pd.DataFrame(
+        {
+            header["col_user"]: [1, 1, 2, 2, 3, 3, 3, 3, 4, 4],
+            header["col_item"]: [1, 2, 1, 2, 3, 4, 5, 6, 1, 2],
+            header["col_rating"]: [3.0, 4.0, 3.0, 4.0, 3.0, 2.0, 1.0, 5.0, 5.0, 1.0]
+        }
+    )
+    model.fit(train)
+
+    similar_users = model.get_topk_most_similar_users(1, 1)
```
miguelgfierro, line 435:
```suggestion
    similar_users = model.get_topk_most_similar_users(user=1, top_k=1)
```
suggestion to improve readibility

tests/unit/recommenders/models/test_sar_singlenode.py
```diff
@@ -413,3 +418,37 @@ def test_match_similarity_type_from_json_file(header):
...
+    assert similar_users[header['col_user']].iloc[0] == 2
+
+    similar_users = model.get_topk_most_similar_users(2, 1)
```
miguelgfierro, line 438:
```suggestion
    similar_users = model.get_topk_most_similar_users(user=2, top_k=1)
```
same as before

tests/unit/recommenders/models/test_sar_singlenode.py
```diff
@@ -413,3 +418,37 @@ def test_match_similarity_type_from_json_file(header):
...
+    assert similar_users[header['col_user']].iloc[0] == 1
+
```
miguelgfierro, line 440: would it be possible to add another line with:
`similar_users = model.get_topk_most_similar_users(user=1, top_k=2)`
then we should check that the most similar user is 2, and then 4. 
Also, it would be good to return the prediction

danb27: sure! for consistency I am evaluating both columns using assert_frame_equal now

danb27:
So given Andreas' comment on the frequencies, I went back to the old way of calculating item_frequencies, without using value_counts, and also rewrote how we calculate user_frequencies to not use the dataframe directly. 

I think there are two remaining questions: 
1) do we want to have some sort of warning / error if a data scientist does provide duplicates? 
2) are we okay with providing these methods for grabbing the most frequent users and grabbing similar users, similar to the functionality provided in the LightFM code? Of course these could be misused, but I think there are legitimate uses for them as well.

If the answer to both is no, then we can probably just close this PR. 

@miguelgfierro @anargyri Please let me know how to proceed when you get a chance.

recommenders/models/sar/sar_singlenode.py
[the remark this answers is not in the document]
miguelgfierro: @danb27 for this I would say that it is better to check whether there are duplicates, and if so, throw an error.
In my experience `df.duplicated().any()` is faster than `any(df.duplicated())` but not sure if your implementation is faster. In `sum(df[[self.col_user, self.col_item]].duplicated())` I would consider also ratings column. I can see a case where the ratings  are related to clicks and purchase, where both signals are valid, so we want to keep both.

recommenders/models/sar/sar_singlenode.py
miguelgfierro: I think it would be valuable to have both item and user similar. Given a list of items, I would like to get a list of the most similar ones for each item, and given a list of users, I would like to get a list of the most similar users for each user.

miguelgfierro:
@danb27 sorry for not answering earlier. Answering your questions:

> do we want to have some sort of warning / error if a data scientist does provide duplicates?

I think we should throw an error if there are duplicates, see details of faster implementations in [this comment](https://github.com/microsoft/recommenders/pull/1666#discussion_r840331225)

> are we okay with providing these methods for grabbing the most frequent users and grabbing similar users, similar to the functionality provided in the LightFM code? Of course these could be misused, but I think there are legitimate uses for them as well.

I think this is ok.

Any other perspective @anargyri?

recommenders/models/sar/sar_singlenode.py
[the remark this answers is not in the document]
danb27: image
@miguelgfierro You are correct, I shouldn't be using sum because then it needs to add up all the True/False in the array resulting from df.duplicated(). Using any, we only need to check until there is a single True - so our best case is a lot better. 

The bigger issue here is that if I implement this error like this (to accommodate your comments and @anargyri 's comments):
```
        select_columns = [self.col_user, self.col_item, self.col_rating]
        if self.time_decay_flag:
            select_columns += [self.col_timestamp]

        if df[select_columns].duplicated().any():
            raise ValueError("There should not be duplicates in the dataframe")
```
then four tests will fail: test_user_affinity, test_recommend_k_items[3-cooccurrence-count], test_recommend_k_items[3-jaccard-jac], test_recommend_k_items[3-lift-lift]

It seems odd to change these tests, since the dataset being used [https://github.com/microsoft/recommenders/blob/main/tests/conftest.py#L91](url) was designed for SAR. I don't know enough about this dataset to know what to do here. 

What do you think?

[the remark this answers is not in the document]
miguelgfierro: @simonzhaoms please see Dan's message above. It looks some of the tests we have both in SAR CPU and SAR+ rely on a dataset with duplicates. 

@danb27 can you tell me exactly which csv have duplicate values? I can create a non duplicate version and upload it to the blob.

FYI @anargyri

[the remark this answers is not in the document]
danb27: @miguelgfierro The csv file with duplicates is located at  https://recodatasets.z20.web.core.windows.net/sarunittest/demoUsage.csv 

I tested locally after deduping with this code: 

```
df = df.sort_values(
    "Timestamp", ascending=False
)
df = df.drop_duplicates(
    ["UserId", "MovieId", "Timestamp"],
    keep="first"
)
```

which is enough to pass all of the unit tests with the code I just pushed. The tests will fail for now, but once you get the deduped version up in storage I can swap the link and everything should be ready to go.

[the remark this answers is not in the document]
danb27: I'm getting this error: `urllib.error.HTTPError: HTTP Error 404: The requested content does not exist.` with the above link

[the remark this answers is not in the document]
miguelgfierro: Sorry, I didn't set the right permissions. Now it should work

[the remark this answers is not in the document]
danb27: tests passed locally! PR is hopefully good to go now after latest commit.

tests/unit/recommenders/models/test_sar_singlenode.py
```diff
@@ -413,3 +418,43 @@ def test_match_similarity_type_from_json_file(header):
 
     # make sure fit still works when similarity type is loaded from a json file
     model.fit(train)
+
```
miguelgfierro, line 421: Can we add a test that fails when there are duplicates?
```
def test_dataset_with_duplicates(header):

    model = SARSingleNode(**header)
    train = pd.DataFrame(
        {
            header["col_user"]: [1, 1, 2, 2, 2],
            header["col_item"]: [1, 2, 1, 2, 2],
            header["col_rating"]: [3.0, 4.0, 3.0, 4.0, 4.0]
        }
    )
with pytest.raises(ValueError):
        model.fit(train)
    
```

miguelgfierro, line 421: Can we add a test that fails when there are duplicates?
```
def test_dataset_with_duplicates(header):

    model = SARSingleNode(**header)
    train = pd.DataFrame(
        {
            header["col_user"]: [1, 1, 2, 2, 2],
            header["col_item"]: [1, 2, 1, 2, 2],
            header["col_rating"]: [3.0, 4.0, 3.0, 4.0, 4.0]
        }
    )
with pytest.raises(ValueError):
        model.fit(train)
    
```

miguelgfierro (approved):
I think the PR is good to go, minor addition for Dan.

@anargyri is there anything else?

tests/unit/recommenders/models/test_sar_singlenode.py
danb27: good call. added.

recommenders/models/sar/sar_singlenode.py
```diff
@@ -267,6 +268,8 @@ def fit(self, df):
         # free up some space
         del temp_df
 
+        # creates an array with the frequency of every unique user/item
+        self.user_frequencies = np.count_nonzero(self.user_affinity.toarray(), axis=1)
```
anargyri, line 272: This is an inefficient way to count these, because it converts a sparse csr matrix to a dense array.

anargyri (changes requested):
See the comment above about avoiding the dense matrix.

recommenders/models/sar/sar_singlenode.py
anargyri: Another thing is that it would be much better if this computation were "on demand" i.e. it should not be computed inside `fit()` since it is not going to be used in production or in most other settings.

danb27: updated with the latest commit. does that work?

recommenders/models/sar/sar_singlenode.py
```diff
@@ -565,3 +566,17 @@ def predict(self, test):
             }
         )
         return df
+
+
+if __name__ == '__main__':
```
anargyri, line 571: This piece of code seems to have been checked in by mistake.

danb27: oops! fixed.

anargyri (approved):

PR #1691, opened 2022-03-31
Use numpy divide in explained variance

anargyri:
### Description

Useful in cases when all true labels are the same and hence explained variance should equal minus infinity.

### Checklist:

### DIFF
--- a/recommenders/evaluation/spark_evaluation.py
+++ b/recommenders/evaluation/spark_evaluation.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+import numpy as np
 try:
     from pyspark.mllib.evaluation import RegressionMetrics, RankingMetrics
     from pyspark.sql import Window, DataFrame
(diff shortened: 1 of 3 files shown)


recommenders/evaluation/spark_evaluation.py
```diff
@@ -158,7 +159,8 @@ def exp_var(self):
             0
         ]
         var2 = self.y_pred_true.selectExpr("variance(label)").collect()[0][0]
-        return 1 - var1 / var2
+        # numpy divide is more tolerant to var2 being zero
+        return 1 - np.divide(var1, var2)
```
miguelgfierro, line 163: I'm looking at the test here: https://github.com/microsoft/recommenders/blob/main/tests/unit/recommenders/evaluation/test_spark_evaluation.py#L239, should we maybe add the edge case for when var2 is zero?

miguelgfierro (approved):

recommenders/evaluation/spark_evaluation.py
anargyri: Maybe, although it is a very unusual case in practice (everything in the data is rated equally).

PR #1704, opened 2022-04-20
Review BPR

miguelgfierro:
### Description

Better explanation of BPR, in particular the pairwise ranking

### Checklist:


miguelgfierro:
FYI @fazamani

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
```diff
@@ -88,16 +88,38 @@
...
+    "\n",
+    "<p align=\"center\">\n",
+    "<img src=\"https://recodatasets.z20.web.core.windows.net/images/bpr1.png\" width=\"400px\">\n",
+    "</p>\n",
+    "\n",
+    "In the implicit feedback scenario the observed data is positive feedback. As it is shown in the figure above, the observed data (image to the left) is known, but the unobserved data is unknown. The negative data is usually generated by filling the matrix with 0 (image to the right). \n",
+    "\n",
+    "Using the pointwise loss, the model is trained to predict as 1 the elements in the observed dataset and 0 for the rest. The problem with this approach is that all elements the model should rank in the future are presented as negative feedback during training. Therefore, an overfitted model, a model that learns the training data exactly, will predict all new elements as 0s. The only reason these models can predict rankings is by using techniques to prevent overfitting, like regularization.\n",
```
anargyri, line 101: This doesn't hold in general: "an overfitted model, a model that learns the training data exactly, will predict all new elements as 0s". Some predictions may be zeros, other ones, depending on the algorithm and the hyperparameters. Also, filling the rest of the matrix with zeros is not what happens in a collaborative filtering scenario, since there is nothing left to predict.

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
```diff
@@ -88,16 +88,38 @@
...
+    "<img src=\"https://recodatasets.z20.web.core.windows.net/images/bpr2.png\" width=\"400px\">\n",
+    "</p>\n",
+    "\n",
+    "In the figure above, pairwise learning is explained. For each user, a pairwise matrix of items is generated. The plus indicates that the user prefers item *i* over item *j*, and the minus is that the user prefers item *j* over *i*. The interrogation indicates that we don't know the user's preference.\n",
+    "\n",
+    "The pairwise setup has several advantages:\n",
+    "* The training set is defined as positive, negative and unknown preferences. \n",
+    "* The missing values between two non-observed items are exactly the item pairs that have to be ranked by the model.\n",
+    "* The training data is created for the actual objective of ranking.\n",
     "\n",
```
anargyri, line 121: Btw the pairwise loss of BPR is not specific to implicit feedback. It could easily be applied when the data contains ratings, since they can be ordered ("user u prefers movie a more than movie b").

anargyri (changes requested):
See comment.

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
miguelgfierro: This comes from the paper: "Machine learning approaches for item recommenders
[5, 10] typically create the training data from S by
giving pairs (u, i) ∈ S a positive class label and all
other combinations in (U × I) \ S a negative one (see
Figure 1). Then a model is fitted to this data. That
means the model is optimized to predict the value 1 for
elements in S and 0 for the rest. The problem with this
approach is that all elements the model should rank in
the future ((U × I) \ S) are presented to the learning
algorithm as negative feedback during training. That
means a model with enough expressiveness (that can
fit the training data exactly) cannot rank at all as it
predicts only 0s. The only reason why such machine
learning methods can predict rankings are strategies
to prevent overfitting, like regularization."
https://arxiv.org/ftp/arxiv/papers/1205/1205.2618.pdf

do you think this doesn't hold true? A perfect network will predict everything unknown as zero, because we are saying that both negatives and unknowns are zero. Regularization is what makes the trick so this doesn't happen in practice.

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
miguelgfierro: yeah, that's a good point. Maybe we can add another point in the line of:

* Pairwise ranking is not restricted to implicit feedback and can be used in explicit feedback by setting user preferences as pairs (i.e. user u prefers item a more than item b).

What do you think?

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
anargyri: Yes, I think what they are saying is that some methods like the SVD method of [13] will overfit (because you specify the whole matrix and interpolate on that). But other methods like [5] (which is similar to ALS) use regularization, so the reg. parameters can be selected so as to avoid overfitting.
I am not sure this says much though. After all the BPR method will overfit too (in a different way) if you set the lambda_Theta to zero. The main difference from methods like [5] is the loss function. So, with BPR you can overfit the pairs of items i.e. overfit the differences in sec. 4.3 (which, up to an additive constant, implies that you overfit the observed ones and zeros).

anargyri: I guess what I am saying is that overfitting is not an issue just with pointwise loss, it will also appear with BPR if you set the reg. param to zero. The text in the notebook gives another impression though.

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
anargyri: Yes, maybe also link to Cornac or some paper that does this.

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
miguelgfierro: got it, makes sense. I have modified the text based on what you said. Let me know if it is more accurate now

examples/02_model_collaborative_filtering/cornac_bpr_deep_dive.ipynb
miguelgfierro: changed

miguelgfierro:
@anargyri could you please review this again?

anargyri (approved):

PR #1715, opened 2022-05-12
Add test to check if CuDNN is enabled

miguelgfierro:
### Description

To make sure we have enabled in our test system cudnn. 

This might help with the problem @pradnyeshjoshi is having

### Checklist:

### DIFF
--- a/tests/unit/recommenders/utils/test_gpu_utils.py
+++ b/tests/unit/recommenders/utils/test_gpu_utils.py
@@ -47,6 +47,11 @@ def test_get_cudnn_version():
 
 
 @pytest.mark.gpu
+def test_cudnn_enabled():
+    assert torch.backends.cudnn.enabled == True
+
+
+@pytest.mark.gpu
 def test_tensorflow_gpu():
     assert tf.test.is_gpu_available()
 


tests/unit/recommenders/utils/test_gpu_utils.py
```diff
@@ -46,6 +46,11 @@ def test_get_cudnn_version():
     assert get_cudnn_version() > "7.0.0"
 
 
+@pytest.mark.gpu
+def test_cudnn_enabled():
+    assert torch.backends.cudnn.enabled == True
```
fazamani, line 51: `assert torch.backends.cudnn.enabled` would also work here.

fazamani (approved):

tests/unit/recommenders/utils/test_gpu_utils.py
miguelgfierro: The problem with just using assert is that it also takes other values. For exampe if you have something like this:
```
def test_simple():
    assert 1
    assert True
    assert "a"
    assert False
```
you'll get 3 oks and one error.

fazamani: That is correct, in this scenario `assert torch.backends.cudnn.enabled` can only be `True` or `False`, so we still get the desired behavior in either cases. But I agree, your way is better for being more explicit and readable. Dropping `==True` is only more concise.

PR #1748, opened 2022-06-15
Rewrote get_top_k_items() to improve runtime

ChuyangKe:
### Description

Rewrote get_top_k_items().

Optimize the runtime of ranking metrics.

### Checklist:

- [x] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions.
- [x] I have updated the documentation accordingly.
- [x] This PR is being made to `staging branch` and not to `main branch`.

PR #1757, opened 2022-06-22
Added tests for ranking function get_top_k_items()

ChuyangKe:
### Description

Added a unit test for get_top_k_items() to ensure correctness. 

### Related Issues

This is a follow-up PR of #1748. See https://github.com/microsoft/recommenders/pull/1748#issuecomment-1157306289.

### Checklist:

### DIFF
--- a/tests/unit/recommenders/evaluation/test_python_evaluation.py
+++ b/tests/unit/recommenders/evaluation/test_python_evaluation.py
@@ -22,6 +22,7 @@ from recommenders.evaluation.python_evaluation import (
     mae,
     rsquared,
     exp_var,
+    get_top_k_items,
     precision_at_k,
     recall_at_k,
     ndcg_at_k,
@@ -75,6 +76,29 @@ def rating_nohit():
             DEFAULT_PREDICTION_COL: [12, 14, 13, 12, 11, 10, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 14, 13],
         }
     )
+
+
+@pytest.fixture
+def top_3_items_true():
+    # Both are correct top 3 items of rating_true
+    return [
+        pd.DataFrame(
+            {
+                DEFAULT_USER_COL: [1, 1, 1, 2, 2, 2, 3, 3, 3],
+                DEFAULT_ITEM_COL: [1, 2, 3, 1, 4, 5, 2, 5, 6],
+                DEFAULT_RATING_COL: [5, 4, 3, 5, 5, 3, 5, 5, 5],
+                "rank": [1, 2, 3, 1, 2, 3, 1, 2, 3],
+            }
+        ),
+        pd.DataFrame(
+            {
+                DEFAULT_USER_COL: [1, 1, 1, 2, 2, 2, 3, 3, 3],
+                DEFAULT_ITEM_COL: [1, 2, 3, 1, 4, 6, 2, 5, 6],
+                DEFAULT_RATING_COL: [5, 4, 3, 5, 5, 3, 5, 5, 5],
+                "rank": [1, 2, 3, 1, 2, 3, 1, 2, 3],
+            }
+        ),
+    ]
 # fmt: on
 
 
@@ -198,6 +222,18 @@ def test_python_exp_var(rating_true, rating_pred):
         == pytest.approx(1.0, TOL)
     )
     assert exp_var(rating_true, rating_pred) == pytest.approx(-6.4466, TOL)
+
+
+def test_get_top_k_items(rating_true, top_3_items_true):
+    top_3_items_df = get_top_k_items(
+        dataframe=rating_true,
+        col_user=DEFAULT_USER_COL,
+        col_rating=DEFAULT_RATING_COL,
+        k=3,
+    )
+    assert (top_3_items_df.equals(top_3_items_true[0])) or (
+        top_3_items_df.equals(top_3_items_true[1])
+    )
 
 
 def test_python_ndcg_at_k(rating_true, rating_pred, rating_nohit):


tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -75,6 +76,29 @@ def rating_nohit():
...
+        ),
+        pd.DataFrame(
+            {
+                DEFAULT_USER_COL: [1, 1, 1, 2, 2, 2, 3, 3, 3],
+                DEFAULT_ITEM_COL: [1, 2, 3, 1, 4, 6, 2, 5, 6],
+                DEFAULT_RATING_COL: [5, 4, 3, 5, 5, 3, 5, 5, 5],
+                "rank": [1, 2, 3, 1, 2, 3, 1, 2, 3],
+            }
+        ),
+    ]
```
miguelgfierro, line 101: This is good @ChuyangKe, some questions. 

The rank list is always [1,2,3, 1,2,3...] in that case, why do we need it?

In the assert you have an or clause because user 2 can get items 1,4,5 or 1,4,6, are you sure that you can get 1,4,6? can you explain how the ordering is done? and also, in this context, how the ordering is done with the rating is the same, for example, user 3 has items 2,5,6 with 5 stars. In which case the rank swaps, for example, getting items 6,5,2?

miguelgfierro (commented):
Some questions

tests/unit/recommenders/evaluation/test_python_evaluation.py
ChuyangKe: This rank list is the expected output of get_top_k_items() 

> "Returns: as.DataFrame: DataFrame of top k items for each user, sorted by `col_user` and `rank`

For user 2, both 1,4,5 and 1,4,6 are correct. This is because item 1,4 have the highest score of 5, and after that item 5 and 6 have the second highest score of 3, so either one should be a valid choice.

What is missing, is the permutation of 2,5,6 for user 3. All cases (2,5,6; 2,6,5; 5,2,6; 5,6,2; 6,2,5; 6,5,2) should be accepted following the same rationale. I will submit a change for this.

```
def rating_true():
    return pd.DataFrame(
        {
            DEFAULT_USER_COL: [1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 1, 1],
            DEFAULT_ITEM_COL: [3, 1, 4, 5, 6, 7, 2, 5, 6, 8, 9, 10, 11, 12, 13, 14, 1, 2],
            DEFAULT_RATING_COL: [3, 5, 5, 3, 3, 1, 5, 5, 5, 4, 4, 3, 3, 3, 2, 1, 5, 4],
        }
    )
```

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,45 @@ def test_python_exp_var(rating_true, rating_pred):
...
+def test_get_top_k_items(rating_true):
+    top_3_items_df = get_top_k_items(
+        dataframe=rating_true,
+        col_user=DEFAULT_USER_COL,
+        col_rating=DEFAULT_RATING_COL,
+        k=3,
+    )
+    top_3_user_true = pd.Series([1, 1, 1, 2, 2, 2, 3, 3, 3])
+    top_3_rating_true = pd.Series([5, 4, 3, 5, 5, 3, 5, 5, 5])
+    top_3_rank_true = pd.Series([1, 2, 3, 1, 2, 3, 1, 2, 3])
```
ChuyangKe, line 213: FIrst 3 lines check the userID, rating, and rank columns

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,45 @@ def test_python_exp_var(rating_true, rating_pred):
...
+    assert(top_3_items_df[DEFAULT_USER_COL].equals(top_3_user_true))
+    assert(top_3_items_df[DEFAULT_RATING_COL].equals(top_3_rating_true))
+    assert(top_3_items_df['rank'].equals(top_3_rank_true))
+    assert(top_3_items_df[DEFAULT_ITEM_COL][:3].equals(pd.Series([1, 2, 3])))
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][3:5]) == set([1, 4]))
+    assert(top_3_items_df[DEFAULT_ITEM_COL][5] in [5, 6])
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][6:]) == set([2, 5, 6]))
```
ChuyangKe, line 220: The lines above check the userID, rating, and rank columns

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,45 @@ def test_python_exp_var(rating_true, rating_pred):
...
+        col_user=DEFAULT_USER_COL,
+        col_rating=DEFAULT_RATING_COL,
+        k=3,
+    )
+    top_3_user_true = pd.Series([1, 1, 1, 2, 2, 2, 3, 3, 3])
+    top_3_rating_true = pd.Series([5, 4, 3, 5, 5, 3, 5, 5, 5])
+    top_3_rank_true = pd.Series([1, 2, 3, 1, 2, 3, 1, 2, 3])
+    assert(top_3_items_df[DEFAULT_USER_COL].equals(top_3_user_true))
+    assert(top_3_items_df[DEFAULT_RATING_COL].equals(top_3_rating_true))
+    assert(top_3_items_df['rank'].equals(top_3_rank_true))
```
ChuyangKe, line 216: The lines above check the userID, rating, and rank columns

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,45 @@ def test_python_exp_var(rating_true, rating_pred):
...
+    assert(top_3_items_df[DEFAULT_ITEM_COL][:3].equals(pd.Series([1, 2, 3])))
```
ChuyangKe, line 217: This checks the itemIDs that are corresponding to user 1. The scores are 5, 4, 3 so it is unique.

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,45 @@ def test_python_exp_var(rating_true, rating_pred):
...
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][3:5]) == set([1, 4]))
+    assert(top_3_items_df[DEFAULT_ITEM_COL][5] in [5, 6])
```
ChuyangKe, line 219: The third itemID of user 2. There are two options: item 5 and 6. Both have a score of 3, so either one is OK.

tests/unit/recommenders/evaluation/test_python_evaluation.py
ChuyangKe, line 220: User 3. All three items have a score of 5, so the ordering is not unique.

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,45 @@ def test_python_exp_var(rating_true, rating_pred):
...
+        k=3,
+    )
+    top_3_user_true = pd.Series([1, 1, 1, 2, 2, 2, 3, 3, 3])
+    top_3_rating_true = pd.Series([5, 4, 3, 5, 5, 3, 5, 5, 5])
+    top_3_rank_true = pd.Series([1, 2, 3, 1, 2, 3, 1, 2, 3])
+    assert(top_3_items_df[DEFAULT_USER_COL].equals(top_3_user_true))
+    assert(top_3_items_df[DEFAULT_RATING_COL].equals(top_3_rating_true))
+    assert(top_3_items_df['rank'].equals(top_3_rank_true))
+    assert(top_3_items_df[DEFAULT_ITEM_COL][:3].equals(pd.Series([1, 2, 3])))
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][3:5]) == set([1, 4]))
```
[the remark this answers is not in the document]
simonzhaoms: @ChuyangKe What do you think to add "First two itemIDs of user 2. The scores are both 5, so the ordering is not unique" as a comment in the code?

tests/unit/recommenders/evaluation/test_python_evaluation.py
simonzhaoms: @ChuyangKe And add this comment in the code as well?

tests/unit/recommenders/evaluation/test_python_evaluation.py
simonzhaoms: @ChuyangKe And this comment?

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,45 @@ def test_python_exp_var(rating_true, rating_pred):
...
+    assert(top_3_items_df[DEFAULT_ITEM_COL][5] in [5, 6])
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][6:]) == set([2, 5, 6]))
+
```
simonzhaoms, line 221:
```suggestion

    # Tests when k is larger than the number of available items
```

simonzhaoms (approved):

tests/unit/recommenders/evaluation/test_python_evaluation.py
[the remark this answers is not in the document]
miguelgfierro: + to Simon's comment, maybe just do the asserts with the exact code that is being executed and then add comments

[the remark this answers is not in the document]
ChuyangKe: The comments are now added.

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,54 @@ def test_python_exp_var(rating_true, rating_pred):
...
+    assert(top_3_items_df['rank'].equals(top_3_rank_true))
+    assert(top_3_items_df[DEFAULT_ITEM_COL][:3].equals(pd.Series([1, 2, 3])))
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][3:5]) == set([1, 4]))
+    # First two itemIDs of user 2. The scores are both 5, so any order is OK.
+    assert(top_3_items_df[DEFAULT_ITEM_COL][5] in [5, 6])
+    # Third itemID of user 2. Both item 5 and 6 have a score of 3, so either one is OK.
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][6:]) == set([2, 5, 6]))
+    # All itemIDs of user 3. All three items have a score of 5, so any order is OK.
+
+    # Tests when k is larger than the number of available items
```
miguelgfierro, line 225: Minor comment, maybe it would help to clarify if you split this test into 2, one for just checking the normal functionality and another one for checking what happens when k is larger than the number of available items

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -200,6 +201,54 @@ def test_python_exp_var(rating_true, rating_pred):
...
+    assert(top_3_items_df[DEFAULT_USER_COL].equals(top_3_user_true))
+    assert(top_3_items_df[DEFAULT_RATING_COL].equals(top_3_rating_true))
+    assert(top_3_items_df['rank'].equals(top_3_rank_true))
+    assert(top_3_items_df[DEFAULT_ITEM_COL][:3].equals(pd.Series([1, 2, 3])))
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][3:5]) == set([1, 4]))
+    # First two itemIDs of user 2. The scores are both 5, so any order is OK.
+    assert(top_3_items_df[DEFAULT_ITEM_COL][5] in [5, 6])
+    # Third itemID of user 2. Both item 5 and 6 have a score of 3, so either one is OK.
+    assert(set(top_3_items_df[DEFAULT_ITEM_COL][6:]) == set([2, 5, 6]))
+    # All itemIDs of user 3. All three items have a score of 5, so any order is OK.
```
miguelgfierro, line 223:
```suggestion
    # First two itemIDs of user 2. The scores are both 5, so any order is OK.
    assert(set(top_3_items_df[DEFAULT_ITEM_COL][3:5]) == set([1, 4]))
    # Third itemID of user 2. Both item 5 and 6 have a score of 3, so either one is OK.
    assert(top_3_items_df[DEFAULT_ITEM_COL][5] in [5, 6])
    # All itemIDs of user 3. All three items have a score of 5, so any order is OK.
    assert(set(top_3_items_df[DEFAULT_ITEM_COL][6:]) == set([2, 5, 6]))
```

It's weird that the comment is below the assert, can you change also the ones below?

miguelgfierro (approved):
Minor comments, looks great

miguelgfierro (approved):
This is great Chuyang

Issue #1750, opened 2022-06-16
[FEATURE] Add time performance benchmark for functions in unit tests


simonzhaoms:
### Description

The idea is from [the pull request for improving the performance of `get_top_k_items()`](https://github.com/microsoft/recommenders/pull/1748#issuecomment-1156972351).  By adding time performance benchmark, we can know whether changes from a PR would affect the performance or not other than the expected correct results.

### Expected behavior with the suggested feature

I find [pytest-benchmark](https://pypi.org/project/pytest-benchmark/) can be used.  But not sure whether there are other ways.

miguelgfierro:
Another way is to meassure the time with the Timer class https://github.com/microsoft/recommenders/blob/main/recommenders/utils/timer.py

PR #1765, opened 2022-06-30
Added time performance tests

ChuyangKe:
### Description

Added time performance tests for rating and ranking metrics. 

The time baselines are provided to ensure that the metrics can be computed in a reasonable amount of time on the same testing server. Currently the thresholds are set as MEAN + 5 * STANDARD DEVIATION based on benchmark results.

List of testing functions:
- test_merge_rating 
- test_merge_ranking 
- test_python_rmse 
- test_python_mae 
- test_python_rsquared 
- test_python_exp_var 
- test_get_top_k_items 
- test_get_top_k_items_largek 
- test_python_ndcg_at_k 
- test_python_map_at_k 
- test_python_precision 
- test_python_recall 
- test_python_auc 
- test_python_logloss

### Related Issues

Related to issue https://github.com/microsoft/recommenders/issues/1750

### Checklist:

- [X] I have added tests covering my contributions.


tests/unit/recommenders/evaluation/test_python_time.py
```diff
@@ -0,0 +1,257 @@
...
+        DEFAULT_USER_COL: random.choices(range(0, DATA_USER_NUM), k=DATA_SAMPLE_NUM),
+        DEFAULT_ITEM_COL: [DATA_ITEM_NUM] * DATA_SAMPLE_NUM,
+        DEFAULT_PREDICTION_COL: random.choices(range(1, DATA_RATING_MAX+1), k=DATA_SAMPLE_NUM),
+    }
+)
+
+
+@pytest.fixture
+def rating_true():
+    return DF_RATING_TRUE.copy()
```
miguelgfierro, line 72: The typical way to do what you want to do is you create a fixture and create the dataframe inside, without the copy and without the global variables.
```
@pytest.fixture
def rating_pred():
    return pd.DataFrame(
    {
        DEFAULT_USER_COL: random.choices(range(0, DATA_USER_NUM), k=DATA_SAMPLE_NUM),
        DEFAULT_ITEM_COL: random.choices(range(0, DATA_ITEM_NUM), k=DATA_SAMPLE_NUM),
        DEFAULT_PREDICTION_COL: random.choices(range(1, DATA_RATING_MAX+1), k=DATA_SAMPLE_NUM),
        # DEFAULT_RATING_COL: random.choices(range(1, RATING_MAX+1), k=SAMPLE_NUM),
    }
)
```
There are several scope levels for fixtures: https://github.com/miguelgfierro/pybase/blob/main/test/pytest_scope.py the default one is function, which means that the df is generated everytime the fixture is called inside a test function. If you are not going to change any of the DF, you can add the scope module, which is that the DF is computed and added to memory once, and then used for all the tests inside this file

tests/unit/recommenders/evaluation/test_python_time.py
```diff
@@ -0,0 +1,257 @@
...
+    return rating_pred
+
+
+# The following time thresholds are benchmarked on Azure
+# Standard_D14_v2 VM, with Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz,
+# 16 cores, 112 GB RAM, 800 GB disk.
+# The thresholds are calculated by MEAN + 5 * STANDARD DEVIATION.
+
+
+def test_merge_rating(rating_true, rating_pred):
```
miguelgfierro, line 110: @pradnyeshjoshi Chuyang needs to add these tests to some of the test groups. I'm thinking that we might want to create a small tutorial to help developers add tests. What do you think?

miguelgfierro (commented):
Looks good, some comments

tests/unit/recommenders/evaluation/test_python_time.py
```diff
@@ -0,0 +1,257 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
miguelgfierro, line 1: @simonzhaoms @ChuyangKe if we are going to start these kinds of tests as a new test pattern, what name would you suggest we use? "time performance tests"? other?

simonzhaoms: I think time performance tests is good

tests/unit/recommenders/evaluation/test_python_time.py
```diff
@@ -0,0 +1,257 @@
...
+    rating_pred[DEFAULT_PREDICTION_COL] = minmax_scale(
+        rating_pred[DEFAULT_PREDICTION_COL].astype(float)
+    )
+    return rating_pred
+
+
+# The following time thresholds are benchmarked on Azure
+# Standard_D14_v2 VM, with Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz,
+# 16 cores, 112 GB RAM, 800 GB disk.
+# The thresholds are calculated by MEAN + 5 * STANDARD DEVIATION.
```
simonzhaoms, line 107: Is it possible to skip the tests if the CPU were inferior to the one mentioned here?

simonzhaoms (changes requested):

tests/unit/recommenders/evaluation/test_python_time.py
```diff
@@ -0,0 +1,257 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
```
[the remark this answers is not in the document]
simonzhaoms: The import is not used in the tests

tests/unit/recommenders/evaluation/test_python_time.py
```diff
@@ -0,0 +1,257 @@
...
+import pandas as pd
+import pytest
+from unittest.mock import Mock
```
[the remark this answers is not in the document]
simonzhaoms: And this either

tests/unit/recommenders/evaluation/test_python_time.py
miguelgfierro: Then maybe @chuyangke, should we rename the file to `test_python_evaluation_time_performance.py` or `test_time_performance_python_evaluation.py` or similar? 

Whatever we choose now, we should replicate the pattern in new files

miguelgfierro: Then maybe @chuyangke, should we rename the file to `test_python_evaluation_time_performance.py` or `test_time_performance_python_evaluation.py` or similar? 

Whatever we choose now, we should replicate the pattern in new files

ChuyangKe: Let's use `test_python_evaluation_time_performance.py` then.

tests/unit/recommenders/evaluation/test_python_time.py
ChuyangKe: Updated

tests/unit/recommenders/evaluation/test_python_time.py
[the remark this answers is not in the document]
ChuyangKe: Updated

miguelgfierro (approved):
really good

tests/unit/recommenders/evaluation/test_python_time.py
ChuyangKe: > Is it possible to skip the tests if the CPU were inferior to the one mentioned here?

Miguel and I had a discussion about this. We were thinking about adding a CPU benchmark function like is_cpu_powerful at the beginning of the module to test the power of the CPU. The function should return true if the target CPU finishes the task faster than our one. Then we add skipif using is_cpu_powerful as the condition to all other time tests.

What do you think?

miguelgfierro: One clarification Simon, the skipif clause is the same as [here](https://github.com/microsoft/recommenders/blob/main/tests/unit/recommenders/tuning/test_nni_utils.py#L55).

Something that would be good to clarify Chuyang is how do you propose to implement the function?

ChuyangKe: I guess a naive way to do this, is to 1) create the synthetic datasets rating_true and rating_pred of the same size, and 2) run map_at_k(). I choose map_at_k() here because its implementation contains the merging function and get_top_k.
We then take the runtime of the two steps combined as the criterion, to decide if the target CPU is more powerful than ours. If it is faster than our baseline, the function returns true and the tests can proceed.

simonzhaoms: Sounds good.

miguelgfierro: This is a very interesting topic. We should think of a general way of approaching this, for example, let's say we also want to use this approach to get the time performance of other tests, like the splitters, we would probably want to have a function that measures the power of the CPU that is agnostic to the specific function we are testing. Maybe a function that breaks a hash? how do people measure the performance of CPUs?

@ChuyangKe if you create an issue in the repo explaining the function that we want to want to create, why we need it, and where we are going to use it, then I can call the community to get feedback and ideas about this.

@simonzhaoms are you ok with leaving the cpu function out of this PR and merge this one once it is finished?

simonzhaoms (approved):

miguelgfierro (approved):
This is great work Chuyang

PR #1788, opened 2022-07-15
Add composite actions to AzureML test workflows

pradnyeshjoshi:
### Description

AzureML test workflows used to call a reusable template workflow inside them. Release pipeline, in turn, calls the test workflows, however this gives an error as nesting of more than two reusable workflows is not supported. This PR refactors the template workflow into composite actions. Composite actions support deeper nesting.
To test that the release pipeline works with this code, tag 1.1.2 is created which triggered [creation of draft release](https://github.com/microsoft/recommenders/actions/runs/2673911478).

### Checklist:

- [x] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions.
- [ ] I have updated the documentation accordingly.
- [x] This PR is being made to `staging branch` and not to `main branch`.

PR #1791, opened 2022-07-18
Add path filter for unit tests

simonzhaoms:
### Description

This PR adds a path filter to check if any changes to the unit tests code, thus the unit tests will not run if only docs are changed.  The paths to trigger the unit tests include (Please confirm @miguelgfierro @pradnyeshjoshi):
* [examples/](https://github.com/microsoft/recommenders/tree/simonz/2022-07-18/unit-tests-file-paths/examples)
* [recommenders/](https://github.com/microsoft/recommenders/tree/simonz/2022-07-18/unit-tests-file-paths/recommenders)
* [tests/](https://github.com/microsoft/recommenders/tree/simonz/2022-07-18/unit-tests-file-paths/tests)
* [setup.py](https://github.com/microsoft/recommenders/blob/simonz/2022-07-18/unit-tests-file-paths/setup.py)

### Checklist:


simonzhaoms:
> it seems there are some conflicts. Simon one question, how can we check that this is effective? do you need the code to be in staging or main? for example, if you change a markdown now, are the tests not triggered?

I forgot you just merged [the PR](https://github.com/microsoft/recommenders/pull/1788).  No wonder all unit tests are triggered.

simonzhaoms:
@miguelgfierro I know what's going on.  As said for [the `pull_request_target` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target), it will run in the context of the base of the pull request, rather than in the context of the merge commit, as [the `pull_request` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) does.
https://github.com/microsoft/recommenders/blob/e16c72ed9d0feb448c399b697dce7eb8cf1fd9c0/.github/workflows/azureml-unit-tests.yml#L8-L21

I think [the `path` filter at the workflow level](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-files-changed-in-a-pull-request-1) works:
https://github.com/microsoft/recommenders/blob/4a6754fd28f9b41556bdc882bd448fb4964d3c04/.github/workflows/azureml-unit-tests.yml#L22-L28 

And [the path filter action at the job level](https://github.com/marketplace/actions/paths-changes-filter) works as well:
https://github.com/microsoft/recommenders/blob/cd9946a4b2d2ee6ea8ffec24f7ab3d423e5e177e/.github/workflows/azureml-unit-tests.yml#L33-L60

miguelgfierro:
@simonzhaoms have you tried this code? do you know if it is not triggered if there is a change in a readme?

.github/workflows/azureml-unit-tests.yml
```diff
@@ -19,6 +19,13 @@ on:
     branches:
       - 'staging'
       - 'main'
+    paths:
+      # Unit tests will be run only when there are changes in the
+      # unit tests related code including:
+      - examples/**
+      - recommenders/**
+      - tests/**
+      - setup.py
```
pradnyeshjoshi, line 28: Perhaps we can use [`paths-ignore`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) to avoid triggering unit tests when we only make changes to markdown files.

miguelgfierro: Either allowing paths or ignoring files, we can use whatever works

miguelgfierro (approved):
ok, let's see if this works

.github/workflows/azureml-unit-tests.yml
pradnyeshjoshi: Looks like the folders mentioned above `examples`, `recommenders`, `tests`  have `.md` files in them. I believe any changes to those will trigger unit tests in the current state.

simonzhaoms: Patched in the PR https://github.com/microsoft/recommenders/pull/1797

PR #1799, opened 2022-07-20
Delete tox.ini

miguelgfierro:
### Description

I think we are not using tox anymore. Earlier today I tried to run the tests with pytest, and I got an error because I have this file. Should we erase it?

FYI @laserprec 

### Checklist:


tox.ini
```diff
@@ -1,104 +0,0 @@
...
-extras = all
-
-[testenv:flake8]
-deps = flake8
-skip_install = True
-commands = flake8 .
-
-
-# Configurations for running pytest
-[pytest]
```
laserprec, line 56: Tox also stores configuration related to other python tools like `pytest` and `flake8`. Maybe the error is relating them. Do you think we would need any of these configurations?

tox.ini
```diff
@@ -1,104 +0,0 @@
...
-    gpu: mark a test as gpu test
-    spark: mark a test as spark test
-    vw: mark a test as vowpal wabbit test
-testpaths =
-    tests
-addopts =
-    # reports all (except passed tests). See https://docs.pytest.org/en/latest/usage.html#detailed-summary-report
-    -ra
-    --durations 10
-    --cov-append --cov=recommenders --cov-report=term --cov-report=xml --junitxml=junit/test-results.xml
```
laserprec, line 80: I think by default we report code coverage at the end of the tests. Maybe the error you are seeing is relating to missing the `coverage` package?

tox.ini
miguelgfierro: I would ask back a question, is there anything in the tox that we can reused in the workflow that @pradnyeshjoshi did https://github.com/microsoft/recommenders/tree/main/.github/workflows

tox.ini
miguelgfierro: @pradnyeshjoshi in the AzureML tests, do we have coverage report?

laserprec (approved):

PR #1802, opened 2022-07-26
Optimized Python splitters

ChuyangKe:
### Description

Rewrote Python splitter functions to improve efficiency. 

Tested on synthetic datasets with 10 million records and 100,000 users. The runtime is reduced from 209.7s to 18.8s, indicating a 11x performance improvement.

### Checklist:

### DIFF
--- a/recommenders/datasets/python_splitters.py
+++ b/recommenders/datasets/python_splitters.py
@@ -98,22 +98,37 @@ def _do_stratification(
 
     groups = data.groupby(split_by_column)
 
+    # data["count"] = groups[split_by_column].transform("count")
+    # data["rank"] = groups.cumcount() + 1
+
+    # data["rank"] = data["rank"] / data["count"]
+
+    # data = data.drop("count", axis=1)
+    # if is_random:
+    #     data = data.drop("random", axis=1)
+
+    # splits = []
+    # prev_threshold = None
+    # for threshold in np.cumsum(ratio):
+    #     condition = data["rank"] <= threshold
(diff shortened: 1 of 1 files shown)


miguelgfierro:
> DO NOT MERGE. More tests are needed.

A small notice. An alternative way to tell the team that you don't want us to merge is by creating a draft PR instead of a normal PR. Simon does it from time to time. There is no need to change this PR, this is just FYI in case you were not familiar :-)

recommenders/datasets/python_splitters.py
```diff
@@ -87,36 +87,33 @@ def _do_stratification(
...
+        data = data.drop("random", axis=1)
 
-    return splits_list
+    splits = []
+    prev_threshold = None
+    for threshold in np.cumsum(ratio):
+        condition = data["rank"] <= round(threshold * data["count"])
+        if prev_threshold is not None:
+            condition &= data["rank"] > round(prev_threshold * data["count"])
+        splits.append(data[condition].drop("rank", axis=1).drop("count", axis=1))
```
miguelgfierro, line 113:
```suggestion
        splits.append(data[condition].drop(["rank", "count"], axis=1))
```
Can you check if this change will get you a better time?

miguelgfierro (approved):
LGTM

Just added an idea to consider

recommenders/datasets/python_splitters.py
ChuyangKe: Yes, this saves an additional 10% of the time. Thanks!

PR #1822, opened 2022-09-20
Tests categories

miguelgfierro:
### Description

Added some of the work that @simonzhaoms and I developed for other projects

### Checklist:


pradnyeshjoshi (approved):

simonzhaoms (approved):

tests/README.md
```diff
@@ -1,24 +1,31 @@
...
 
-These types of tests are integrated in the repo in two ways, via the PR gate, and the nightly builds. 
+## Categories of tests
 
-The PR gate are the set of tests executed after doing a pull request and they should be quick. Here we include the unit tests, that just check that the code doesn't have any errors.
+The tests in this repository are divided into the following categories:
 
-The nightly builds tests are executed asynchronously and can take longer. Here we include the smoke and integration tests, and their objective is to not only make sure that there are not errors, but also to make sure that the machine learning solutions are doing what we expect.
+* **Data validation tests:** In the data validation tests, we ensure that the schema for input and output data for each function in the pipeline matches the desired prespecified schema. They should be fast and can be added to the PR gate.
+* **Unit tests**: In the unit tests we just make sure the python utilities and notebooks run correctly. Unit tests are fast, ideally less than 5min and are run in every pull request. They belong to the PR gate. For this type of tests, synthetic data can be used.
```
fazamani, line 20: one can argue that the unit tests are the tests at the lowest (atomic) level of the code, to isolate issues in specific part of the code. With this definition, running notebooks (e2e) doesn't fall into unit test. But I am not sure where to put 'test notebooks'

fazamani (approved):

setuc:
I would also want to add the following tests or checks as part of the PR / scheduled builds
* Code Scans to include for potential security issues or code quality checks, e.g. unused imports etc. We are currently using LGTM in our repository together with Sonnar cloud
* We also want to enable container scans to find potential security issues either in python packages or the underlying OS as part of PR in addition to scheduled scans in the production pipelines. 
* Would we need to include an endpoint scan using OWASP?
* Consideration for load testing, especially when working with larger models in CV and NLP.

tests/README.md
miguelgfierro: Looking at the Wikipedia: 
> In computer programming, unit testing is a software testing method by which individual units of source code—sets of **one or more computer program modules together with associated control data, usage procedures, and operating procedures—are tested** to determine whether they are fit for use

Based on this definition if we consider notebook as an individual unit, or we consider it as a unit with multiple modules, it could be argued that they fit into unit tests. A different situation would be if instead of just checking if the notebook works, we also check that it returns the right information. Then that would be in the realm of functional. 

It seems one of the pioneers in testing was Erich Gamma https://en.wikipedia.org/wiki/Unit_testing maybe we should contact him for advice :-)

tests/README.md
```diff
@@ -1,24 +1,31 @@
...
+* **Smoke tests:** The smoke tests are gates to the slow tests in the nightly builds to detect quick errors. If we are running a test with a large dataset that takes 4h, we want to create a faster version of the large test (maybe with a small percentage of the dataset or with 1 epoch) to ensure that it runs end-to-end without obvious failures. Smoke tests can run sequentially with functional or integration tests in the nightly builds, and should be fast, ideally less than 20min. They use the same type of data as their longer counterparts.
+* **Performance test:** The performance tests are tests that measure the computation time or memory footprint of a piece of code and make sure that this is bounded between some limits.  For this type of tests, synthetic data can be used.
```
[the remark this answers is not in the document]
miguelgfierro: this is an interesting point of view @abeomor. Will this be in the realm of regression tests when we want to measure whether different versions of a library provide different values, or are you thinking in the context of benchmarking different models for the same dataset like this https://github.com/microsoft/recommenders/blob/main/examples/06_benchmarks/movielens.ipynb

miguelgfierro:
@setuc really good points. We mentioned security in the RAI tests, but not sure if it would be convenient to create a different categories. In fact, here https://en.wikipedia.org/wiki/Software_testing there is a category of security testing. 

About load testing, this is a really good point. We will incorporate it in the text

miguelgfierro:
@setuc @fazamani @AbeOmor I incorporated the feedback in the new commit.

miguelgfierro:
Feedback from Erich Gamma:
Each project needs to come up with a strategy that ensure the quality of the project while preserving agility. It looks like you are already using different test categories following the idea of a testing pyramid https://martinfowler.com/articles/practical-test-pyramid.html (this is a pretty good article on the topic).

Given that you already have a decent amount of tests in the Recommender repo I suggest that you start to reflect about what are your pain points with your current tests. For example,

- Not enough tests: make it easier to implement tests. For example by providing a library of test fixtures/setups
- Flakey tests: Team effort to make tests more stable, establish culture that it is not OK to have flakey tests. File issues against these tests
- Not enough coverage: Make the current coverage visible using a coverage report
Not enough automation typically an issue for smoke testing that involves UI (not an issue for your project): Investigate into a UI testing framework like
…

miguelgfierro:
Quote from the link that Erich sent: 

> Write lots of small and fast unit tests. Write some more coarse-grained tests and very few high-level tests that test your application from end to end.

> Given the shortcomings of the original names it's totally okay to come up with other names for your test layers, as long as you keep it consistent within your codebase and your team's discussions.

PR #1911, opened 2023-03-30
Reduce double nightly tests

miguelgfierro:
### Description

I noticed that when there is a merge to staging and a PR to main, we are doing twice the nightly tests. They are the exact same tests so there is no need to have it twice.

image

By removing the trigger when there is a PR to main, we are still checking that the nightly have passed, because all new code that is merged to staging passes the nightly. 

In addition, every 5 days we are scheduling the nightly in main

### Checklist:

### DIFF
--- a/.github/workflows/azureml-cpu-nightly.yml
+++ b/.github/workflows/azureml-cpu-nightly.yml
@@ -18,12 +18,11 @@ on:
     # cron works with default branch (main) only: # https://github.community/t/on-schedule-per-branch/17525/2
   
   push:
-    # because we can't schedule runs for non-main branches,
+    # Because we can't schedule runs for non-main branches,
     # to ensure we are running the build on the staging branch, we can add push policy for it
     branches: [staging]
     paths:
-      # Unit tests will be run only when there are changes in the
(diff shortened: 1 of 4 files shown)


wutaomsft:
@miguelgfierro the picture above - I did not see tests being duplicated?

miguelgfierro:
When Pradnyesh did the solution, he found that there is an issue with GitHub that it only allows to schedule cron jobs in main https://github.com/microsoft/recommenders/blob/main/.github/workflows/azureml-cpu-nightly.yml#L18
So instead of doing a schedule job in staging, we run the nightlys when we merge into staging. 

The code we had before executed nightly when there was a PR to main (in staging branch). So we are repeating the same tests with the same code twice every time we merge to staging and there is a PR to main. 

This PR removes the nightly in the PR to main, because it is actually repeating the same test twice. 

If you notice, in the screenshot, the 3 nightlys are executed at the same time. 

Also in the figure you'll notice that when we do a PR to main we are executing the unit tests. This makes sense and are not repeated.

Let me know if you need additional details

wutaomsft (approved):

PR #1978, opened 2023-08-25
License update

miguelgfierro:
### Description

Reviewing this with Justin, once he approves, I'll update the rest of the files

### Checklist:


LICENSE
```diff
@@ -1,6 +1,7 @@
     MIT License
 
-    Copyright (c) Microsoft Corporation. All rights reserved.
+    Copyright (c) 2018-present Microsoft Corporation. All rights reserved.
```
miguelgfierro, line 3: This follows the convention of PyTorch, see https://github.com/pytorch/pytorch/blob/main/LICENSE

miguelgfierro: The feedback from Justin is that it is ok, and that we can remove the "All right reserved"

miguelgfierro:
FYI @ibrahimhaddad in case you want to review

LICENSE
```diff
@@ -1,6 +1,7 @@
     MIT License
 
-    Copyright (c) Microsoft Corporation. All rights reserved.
+    Copyright (c) 2018-present Microsoft Corporation.
```
loomlike, line 3: is this `present` correct? or should this be `2018-2023`?

loomlike (approved):
Thank you!

LICENSE
miguelgfierro: Not 2023, because otherwise we will need to update it every year. In PyTorch they put present https://github.com/pytorch/pytorch/blob/main/LICENSE#L15. I have also seen 2018- or 2018.

Issue #2002, opened 2023-09-27
[FEATURE] Spark and python's evaluation metrics should produce the same result.


loomlike:
### Description
Spark and python's evaluation metrics should produce the same result.
For that, we should refactor pytest to check if they produce the same values for the same test.
This issue is to track the pytest refactoring to implement that.

### Expected behavior with the suggested feature
Define test parameters and expected-outputs by using a fixture. Then use the same fixture for both python and pyspark evaluation metrics.

Issue #1990, opened 2023-09-10
[ASK] is map_at_k in spark_evaluation actually at k?


spirosdim:
### Description
Hi, after looking at the [spark_evaluation.py](https://github.com/recommenders-team/recommenders/blob/staging/recommenders/evaluation/spark_evaluation.py#L350) the `map_at_k` method, I noticed that it uses [`meanAveragePrecision`](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/mllib/evaluation/RankingMetrics.scala#L83) from _spark.mllib_.  
Should `map_at_k` use the [`meanAveragePrecisionAt`](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/mllib/evaluation/RankingMetrics.scala#L99) instead? In this way, it will use the min(k, pred.length) restricting the calculation of MAP to the top k predictions. 
Is there something that I miss?

### Other Comments

- The current `map_at_k` could be named as just `map`, using `meanAveragePrecision` from _spark.mllib_. 
- The `map_at_k` could use the `meanAveragePrecisionAt` from _spark.mllib_.

loomlike:
Thank you for the comment. I'm checking it and will get back to you soon! I also created an issue #2002 to enhance the test for the metrics.

Issue #1702, opened 2022-04-19
MAP_AT_K implementation is wrong


abitbetter:
### Description

Hi, after several thoughts on it and performing my own tests, I think that the implementation of map_at_k is not correct.

[https://github.com/microsoft/recommenders/blob/main/recommenders/evaluation/python_evaluation.py](https://github.com/microsoft/recommenders/blob/main/recommenders/evaluation/python_evaluation.py)

The return statement shows the following:

`return (df_merge["rr"] / df_merge["actual"]).sum() / n_users`

Although the function documentation does advice that:

`2. The MAP is meant to calculate Avg. Precision for the relevant items, so it is normalized by the number of
        relevant items in the ground truth data, instead of k.`

I think it is mistaken and should normalize (divide) by the min(k, actual). 

By dividing just by the "actual" column, what is computed is the recall, not the precision.
In other words, if I measure how well I am recommending some items to a user, by dividing by the "actual" makes the MAP for that user dependent on how many items he/she bought.  So if I recommend 3 items and the user purchases those 3 plus 5 more, the MAP metric will be different to a situation in which I recommend 3 items and the user purchases those 3 plus 10 more.  Which doesn't make sense because the mean average precision should be independent of how many items are purchased (unless actual < k) as it is expected to measure precision, not recall.

Having said the above, I think that the return value should be instead something similar to this:
`df_merge["top_k"] = k`
 `df_merge["denominator"] = df_merge[["top_k", "actual"]].min(axis=1)`
` return (df_merge["rr"] / df_merge["denominator"]).sum() / n_users`

I would really appreciate your comments on this!

Thank you in advance!

anargyri:
Well, we have seen different ways to define these metrics in the literature and practice. We opted to follow the Spark MLLib definition of MAP, as we note in the docstring i.e. the definition that can be found here https://spark.apache.org/docs/2.3.0/mllib-evaluation-metrics.html#ranking-systems
They divide by the "actual" number, whereas @abitbetter suggests dividing my min(k, actual) instead.
There are pros and cons for both ways; I see the argument in favor of the latter but on the other side if k is larger than the number of relevant items then it is not possible to achieve 100% precision with the latter definition.
I suggest that @abitbetter implements the alternative definition so that it can be activated with a flag (a boolean argument added in the MAP signature). Also to add a case testing this definition in the MAP test.

anargyri:
[...]

PR #2003, opened 2023-09-28
Announcement LF

miguelgfierro:
### Description

A new hope
image

Imagine the music ...

### Related Issues

https://github.com/recommenders-team/recommenders/issues/1997

### Checklist:

- [ ] I have followed the [contribution guidelines](../CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions.
- [ ] I have updated the documentation accordingly.
- [ ] This PR is being made to `staging branch` and not to `main branch`.

PR #2004, opened 2023-09-28
Refactor ranking metric `map` to be the same as Spark's

loomlike:
### Description

1. Refactor both python and spark ranking metric `map` to be the same as Spark's.
- Rename `map_at_k` to `map` which basically the same as Spark's `meanAveragePrecision`
- Add MAP@K `map_at_k` to be the same as Spark's `meanAveragePrecisionAt`

2. Refactor evaluation tests to use the same data fixtures between python and spark tests.

Note: With this changes, some notebook tests are failing.
I'll need to update the expected values for `map_at_k` metrics for those notebooks, or simply to use `map` that is our previous `map_at_k`.

### Related Issues

#2002 
#1990 
#1702 

### Checklist:


loomlike:
btw, why draft PR triggers tests? @miguelgfierro @anargyri Can we disable that?

miguelgfierro:
> btw, why draft PR triggers tests? @miguelgfierro @anargyri Can we disable that?

This might be a solution: https://github.com/orgs/community/discussions/25722 we need to try it. Issue: #2003 2005

fix.txt
```diff
@@ -0,0 +1,12 @@
+# FAILED unit/examples/test_notebooks_python.py::test_sar_deep_dive_runs - papermill.exceptions.PapermillExecutionError:
```
miguelgfierro, line 1: Here is a trick I use: 
I usually add the errors in the text of the PR and then I go one by one fixing them. 

This also helps other people in the future to fix similar errors.

miguelgfierro, line 1: Another thing, since we joined the LF, there is a new action we need to pass: DCO. You noticed that it failed. The solution here: https://github.com/recommenders-team/recommenders/wiki/How-to-sign-commits

anargyri:
@loomlike are the test failures related to the values of the metrics? I vaguely recall that we had found some bugs in the Spark implementation (or they were not matching the definitions), hence our own implementations for some of the metrics.

loomlike:
> @loomlike are the test failures related to the values of the metrics? I vaguely recall that we had found some bugs in the Spark implementation (or they were not matching the definitions), hence our own implementations for some of the metrics.

I think the error was coming from test group names (as I changed some of the tests names, e.g. test_python_precision -> test_python_precision_at_k). I changed them in the test group lists. Let's see if that works.

Regarding the MAP, Spark added `meanAveragePrecisionAt` since [v3.0.0](https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.mllib.evaluation.RankingMetrics.html#pyspark.mllib.evaluation.RankingMetrics.meanAveragePrecisionAt). So I changed our `map_at_k` to be just `map` and added `map_at_k` which implements "at_k" parts for both python and spark utils.

Regarding the definition of MAP vs. MAP@K, actually what we had before was just [MAP based on Wikipedia definition](https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Mean_average_precision) as well as Spark's implementation.

The [community](https://www.kaggle.com/code/nandeshwar/mean-average-precision-map-k-metric-explained-code) seems to follow the following definition:
> AP@k is typically calculated for one user from all recommendations by averaging P@1, P@2, P@3, ... P@k
which the Spark's `meanAveragePrecisionAt` also follows it (and the codes in this PR also follow that definition).

anargyri:
Good, as long as the evaluation notebooks give the same metrics as before we should be fine.

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -164,10 +150,10 @@ def test_python_rmse(rating_true, rating_pred):
         )
         == 0
     )
-    assert rmse(rating_true, rating_pred) == pytest.approx(7.254309, TOL)
+    assert rmse(rating_true, rating_pred) == target_metrics["rmse"]
```
miguelgfierro, line 153: In other tests we hard code the results, see https://github.com/recommenders-team/recommenders/tree/main/tests#how-to-create-tests-for-the-recommenders-library with the pattern computation equals value.

I would argue it's more clear and explicit

tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -100,12 +67,31 @@ def rating_pred_binary(rating_pred):
 
 
 def test_column_dtypes_match(rating_true, rating_pred):
-    # Change data types of true and prediction data, and there should type error produced
-    rating_true[DEFAULT_USER_COL] = rating_true[DEFAULT_USER_COL].astype(str)
-    rating_true[DEFAULT_RATING_COL] = rating_true[DEFAULT_RATING_COL].astype(str)
+    mocked_fn = Mock(return_value=True)
```
miguelgfierro, line 70: A question here, what is the function of the mock here?

miguelgfierro:
@loomlike if you get an error on DCO you can go to details and approve it. Only admins can do this. I've approved this one.

miguelgfierro (dismissed):
@loomlike hey we got an internal security email and it seems that it is due to an unsecured version of TF. 

That caused a lot of problems, as expected, some of the TF algos with the v1 API break. I tried to figure out a way to fix it, but I couldn't. See https://github.com/recommenders-team/recommenders/pull/2017

Until this is merged to main, we can't start new VMs for the tests because they will have the old TF version and will trigger the security issue. 

So would you mind stop doing any new commit to this PR until the TF situation is fixed?

TF=💩

miguelgfierro:
@loomlike the issue is fixed https://github.com/recommenders-team/recommenders/pull/2017

Can you please merge the latest staging into your branch before committing?

tests/unit/recommenders/evaluation/test_python_evaluation.py
loomlike: I agree with you when an expected value is clear and easy to understand. E.g.
```
# check if the metric is the best number (rmse == 0) if the prediction is perfect, i.e true == prediction.
assert rmse(rating_true, rating_true) == 0
```

I also agree that it is explicit to put the number directly when we also specify the input in the same way, like:
```
output = subtract_two(2.2345, 1.0)  
assert output == 1.2345  # this is explicit

# versus

output = subtract_two(input_fixture_data_a, input_fixture_data_b)
assert output == 1.2345  # our input is already in a blackbox for reusability
```

So, I still prefer using an expected value fixture, the same reason we use the fixture for the input data:
1. By using a single fixture value, we can reuse them across many tests and easy to maintain the test cases with the expected values.
2. We can assure both python and spark implementation produces the same metrics for the same inputs (in addition to the test we directly check both metrics' equalities, `test_spark_python_match`)

What do you think?

tests/unit/recommenders/evaluation/test_python_evaluation.py
loomlike: that is to test if `_check_column_dtypes` returns the same output as the input function when there's no exceptions.
I'll add a comment to the test.

tests/unit/recommenders/evaluation/test_python_evaluation.py
miguelgfierro: > I still prefer using an expected value fixture, the same reason we use the fixture for the input data

There are several things to consider here:
- Fixtures are not designed for the results of computations it is designed for data. Because you want to control how the data is loaded (test level, module, or session). For example, maybe we want to load movielens at a module level and not every test because it adds 1 or 2 seconds for every test. A result of a computation doesn't need that, because it is just one number.
- The pattern should be computation == value. Because a common failure is to do a computation1 = computation2, people don't realize that computation2 can also fail, so then they think they are testing computation1, but the error can come from computation1 or computation2. I think it is important to make clear this to less experienced developers (after all, we are trying to teach good coding practices).
- I agree with you that it is convenient to have common results for python and spark, but then it shouldn´t be a fixture, but constants. We can add them as constants or as a dict, or we can directly add the number to keep the consistency with the rest of the repo. 

Thoughts?

loomlike: @miguelgfierro per your suggestion, I changed the way the tests check values. Please review again and see if this PR is good to merge!

miguelgfierro (approved):
Awesome!

Issue #2018, opened 2023-10-13
[BUG] Upgrade to Tensorflow>2.10.1 breaks several of the DeepRec algos


miguelgfierro:
### Description

We had a security alert https://github.com/recommenders-team/recommenders/pull/2017 and we had to update the TF version.

When we moved to >2.10.1, several algos stopped working: xDeepFM, SUM, LightGCN and SliRec

### In which platform does it happen?

```
=========================== short test summary info ============================
FAILED tests/unit/recommenders/models/test_deeprec_model.py::test_xdeepfm_component_definition
FAILED tests/unit/recommenders/models/test_deeprec_model.py::test_slirec_component_definition
FAILED tests/unit/recommenders/models/test_deeprec_model.py::test_sum_component_definition
============= 3 failed, 12 passed, 7 warnings in 153.99s (0:02:33) =============
```

```
=================================== FAILURES ===================================
______________________ test_xdeepfm_component_definition _______________________

deeprec_resource_path = PosixPath('/mnt/azureml/cr/j/e49a1dc1154c4ee3b42372de2b064c9c/exe/wd/tests/resources/deeprec')

    @pytest.mark.gpu
    def test_xdeepfm_component_definition(deeprec_resource_path):
        data_path = os.path.join(deeprec_resource_path, "xdeepfm")
        yaml_file = os.path.join(data_path, "xDeepFM.yaml")
    
        if not os.path.exists(yaml_file):
            download_deeprec_resources(
                "https://recodatasets.z20.web.core.windows.net/deeprec/",
                data_path,
                "xdeepfmresources.zip",
            )
    
        hparams = prepare_hparams(yaml_file)
>       model = XDeepFMModel(hparams, FFMTextIterator)
E       NameError: name 'XDeepFMModel' is not defined

tests/unit/recommenders/models/test_deeprec_model.py:131: NameError
----------------------------- Captured stdout call -----------------------------
INFO:recommenders.datasets.download_utils:Downloading https://recodatasets.z20.web.core.windows.net/deeprec/xdeepfmresources.zip
----------------------------- Captured stderr call -----------------------------

  0%|          | 0.00/10.3k [00:00<?, ?KB/s]
 68%|██████▊   | 6.97k/10.3k [00:00<00:00, 69.7kKB/s]
100%|██████████| 10.3k/10.3k [00:00<00:00, 81.3kKB/s]
------------------------------ Captured log call -------------------------------
[...]

Issue #2012, opened 2023-10-08
[FEATURE] Alternative to scrapbook to execute notebooks programmatically for tests


miguelgfierro:
### Description

Scrapbook is not being developed anymore, and it doesn't support Python 3.10 (See https://github.com/recommenders-team/recommenders/pull/1988#issuecomment-1712425248)

miguelgfierro:
WIP: https://github.com/miguelgfierro/pybase/pull/73

SimonYansenZhao:
@miguelgfierro I don't find any alternatives to scrapbook after googling.  The solutions, I think, may be
1. Use nbformat as you do in https://github.com/miguelgfierro/pybase/pull/73
2. Borrow and modify the functions like [`sb.glue()`](https://github.com/nteract/scrapbook/blob/main/scrapbook/api.py) as @anargyri said in the meeting and make them part of Recommenders.

For the 1st option, I saw you written the functions to do that, what else should we do?  I can try tomorrow.

miguelgfierro:
Actually, option 2 is kind of what I was doing in 1. If you have time it would be great if you could try 1.

miguelgfierro:
I think I have finished the functions, see https://github.com/miguelgfierro/pybase/pull/73

Man, they say the most difficult language in the world is Chinese, but I think it is regex. Thank goodness that we have ChatGPT.

@SimonYansenZhao  I'll start a PR soon to replace all the scrapbook

miguelgfierro:
Fixed!

PR #2022, opened 2023-10-17
Resolve issue #2018

SimonYansenZhao:
### Description

This PR resolves the issue #2018 by commenting out the tests in [tests/ci/azureml_tests/test_groups.py](https://github.com/recommenders-team/recommenders/blob/simonz/tf-security-alert-issue-2018/tests/ci/azureml_tests/test_groups.py) to disable testing for [deeprec](https://github.com/recommenders-team/recommenders/blob/simonz/tf-security-alert-issue-2018/tests/unit/recommenders/models/test_deeprec_model.py). 

### Related Issues

* #2018 

### References

* TensorFlow
  + [TensorFlow release history@PyPI](https://pypi.org/project/tensorflow/#history)
  + [Requirements for older TensorFlow versions](https://www.tensorflow.org/install/source#tested_build_configurations)
* Versioning
  + [PEP 440 – Version Identification and Dependency Specification](https://peps.python.org/pep-0440/#version-specifiers)
  + [Version comparison](https://packaging.pypa.io/en/stable/version.html)

### Checklist:


SimonYansenZhao:
First, I got `NameError: name 'XDeepFMModel' is not defined` (See https://github.com/recommenders-team/recommenders/actions/runs/6544943740/job/17772448740#step:3:2992).  Then I got `ModuleNotFoundError: No module named 'keras.layers.legacy_rnn'` after removing the `try-except` in the commit (https://github.com/recommenders-team/recommenders/pull/2022/commits/88e43a5f112c411cf6801b356ee690f2a0ab5cb4#) (See https://github.com/recommenders-team/recommenders/actions/runs/6545498077/job/17774104408?pr=2022#step:3:3012).

So `Keras.layers.legacy_rnn` imported in https://github.com/recommenders-team/recommenders/blob/b000b78ceb3cbe52a0200922f2b2412d830274af/recommenders/models/deeprec/models/sequential/sum_cells.py#L6 is removed from Keras.  Need to find a way to replace it.

anargyri:
> First, I got `NameError: name 'XDeepFMModel' is not defined` (See https://github.com/recommenders-team/recommenders/actions/runs/6544943740/job/17772448740#step:3:2992).
> 
> I think something might be wrong when importing `XDeepFMModel`, then after removing the `try-except` in the commit ([88e43a5](https://github.com/recommenders-team/recommenders/commit/88e43a5f112c411cf6801b356ee690f2a0ab5cb4)) (See https://github.com/recommenders-team/recommenders/actions/runs/6545498077/job/17774104408?pr=2022#step:3:3012) I got `ModuleNotFoundError: No module named 'keras.layers.legacy_rnn'` .
> 
> It means `Keras.layers.legacy_rnn` imported in
> 
> https://github.com/recommenders-team/recommenders/blob/b000b78ceb3cbe52a0200922f2b2412d830274af/recommenders/models/deeprec/models/sequential/sum_cells.py#L6
> 
> is removed from Keras. Need to find a way to replace it.
> FYI @miguelgfierro @anargyri

It looks like this class is deprecated in the new versions; if I recall, it was one of the temporary fixes when we upgraded from v1 to v2. Now you would have to rewrite the sum cell in the new syntax using [this](https://www.tensorflow.org/api_docs/python/tf/keras/layers/RNN), which may not be straightforward. Maybe we should migrate this code to PyTorch. @Leavingseason what do you think?

That said, I see the code is still in TF master 
https://github.com/tensorflow/tensorflow/tree/r2.14/tensorflow/python/keras/layers/legacy_rnn
They stopped exposing the module but it is still used internally e.g. [here](https://github.com/tensorflow/tensorflow/blob/r2.14/tensorflow/python/keras/layers/rnn_cell_wrapper_v2.py) 
I wonder if copying and pasting the `legacy_rnn` code in our repo would work (provided that the TF license allows for this).

It looks like the conda env is installing 2.13. How about restricting to < 2.13 or < 2.12, do you still get the error? If not, we could restrict the TF version from above.

anargyri:
@SimonYansenZhao I checked locally with different versions of TF and this import
`from keras.layers.legacy_rnn.rnn_cell_impl import LayerRNNCell `
only works up to (and including) version 2.8.

anargyri:
We need to discuss a bit more how to proceed, it is not clear to me what is the best way forward.
For now, let's bound TF with < 2.9.

miguelgfierro:
pinged Jianxun on teams

Master-PLC:
Hi, feen from SimonYansenZhao's discovery, when the TF>=2.9, the reference in the file recommenders/models/deeprec/models/sequential/sum_cells.py, "from keras.layers.legacy_rnn.rnn_cell_impl import LayerRNNCell" becomes invalid, which causes the XDeepFMModel to fail to import and run.
Upon examining the LayerRNNCell and RNNCell classes in the tf.keras.layers.legacy_rnn.rnn_cell_impl file, I found that LayerRNNCell simply inherits from the RNNCell class. When calling, it reduces one step of building scope compared to RNNCell, so we can directly use "from keras.layers.legacy_rnn.rnn_cell_impl import RNNCell" as the parent class of SUM_Cell.
However, in TF>=2.9, the RNNCell class has been deprecated, but it has been rewritten as AbstractRNNCell in tf.keras.layers.recurrent with similar implementation and comments. Therefore, maybe we can try replacing "from keras.layers.legacy_rnn.rnn_cell_impl import LayerRNNCell" with:
try:
    from keras.layers.legacy_rnn.rnn_cell_impl import LayerRNNCell
except:
    from keras.layers.recurrent import AbstractRNNCell as LayerRNNCell

SimonYansenZhao:
I got the following error when adding `from keras.layers.rnn import AbstractRNNCell as LayerRNNCell`.  It looks like that `AbstractRNNCell` does not have `_reuse`.  @Master-PLC 

Screenshot 2024-01-22 at 22 40 36

setup.py
```diff
@@ -58,7 +58,7 @@
 extras_require = {
     "gpu": [
         "nvidia-ml-py3>=7.352.0",
-        "tensorflow==2.8.4",  # FIXME: Temporarily pinned due to issue with TF version > 2.10.1 See #2018
+        "tensorflow>=2.8.4,!=2.9.0.*,!=2.9.1,!=2.9.2,!=2.10.0.*,<2.12",
```
miguelgfierro, line 61:
```suggestion
        "tensorflow>=2.8.4,!=2.9.0.*,!=2.9.1,!=2.9.2,!=2.10.0.*,<3",
```

Can we try removing `<2.12`, and maybe add <3? otherwise we will be constraining TF to old versions

tests/security/test_dependency_security.py
```diff
@@ -7,6 +7,8 @@
 import numpy as np
 import pandas as pd
 
+from packaging.version import Version
```
miguelgfierro, line 10: I would suggest to remove this new package. It adds a new dependency, and it doesn't add too much value.

In the past, we had a lot of problems when adding extra dependencies, I think we should try to minimize external dependencies as much as possible

tests/ci/azureml_tests/test_groups.py
```diff
@@ -426,13 +426,13 @@
...
-        "tests/unit/recommenders/models/test_deeprec_model.py::test_sum_component_definition",  # FIXME: Issue with TF version > 2.10.1 See #2018
-        "tests/unit/recommenders/models/test_deeprec_model.py::test_lightgcn_component_definition",
+        # "tests/unit/recommenders/models/test_deeprec_model.py::test_xdeepfm_component_definition",  # FIXME: Disabled due to the issue with TF version > 2.10.1 See #2018
+        # "tests/unit/recommenders/models/test_deeprec_model.py::test_dkn_component_definition",  # FIXME: Disabled due to the issue with TF version > 2.10.1 See #2018
+        # "tests/unit/recommenders/models/test_deeprec_model.py::test_dkn_item2item_component_definition",  # FIXME: Disabled due to the issue with TF version > 2.10.1 See #2018
+        # "tests/unit/recommenders/models/test_deeprec_model.py::test_slirec_component_definition",  # FIXME: Disabled due to the issue with TF version > 2.10.1 See #2018
+        # "tests/unit/recommenders/models/test_deeprec_model.py::test_nextitnet_component_definition",  # FIXME: Disabled due to the issue with TF version > 2.10.1 See #2018
```
miguelgfierro, line 433: @SimonYansenZhao I thought that the issue was only with xDeepFM. All the algos in deeprec are affected?

tests/security/test_dependency_security.py
SimonYansenZhao: I've checked that **packaging** has no extra dependencies.  See [packaging/pyproject.toml](https://github.com/pypa/packaging/blob/main/pyproject.toml)

tests/ci/azureml_tests/test_groups.py
SimonYansenZhao: Not sure.  I'll check.

tests/security/test_dependency_security.py
SimonYansenZhao: Because packaging is only used for testing, do you have any suggestion on where to packaging.  @miguelgfierro   I saw you added [docs/requirements-doc.txt](https://github.com/recommenders-team/recommenders/blob/staging/docs/requirements-doc.txt) for doc dependencies.

miguelgfierro: > Because packaging is only used for testing, do you have any suggestion on where to packaging

In the past, adding extra dependencies gave us a lot of extra work. For example, the issue we had with snapbook #2012 and for example, a new one we need to fix with pandera #2049. 

Even though this new library doesn't have additional dependencies, it is still one extra dependency. In addition, it is just simpler to do the comparison like this `torch.__version__ >= "1.13.1"` than `Version(torch.__version__) >= Version("1.13.1")`

I think we should only add a new dependency if there is no other alternative and it is really necessary. That would save us a lot of work.

Does it make sense @SimonYansenZhao?

SimonYansenZhao:
When using the latest TensorFlow (2.15), I got [this error](https://github.com/recommenders-team/recommenders/actions/runs/7711491218/job/21048241208#step:3:707):

```
E               713 try:
E               714   # pylint: disable=protected-access
E               715   with self._graph._c_graph.get() as c_graph:
E           --> 716     self._session = tf_session.TF_NewSessionRef(c_graph,opts)
E               717   # pylint: enable=protected-access
E               718 finally:
E               719   tf_session.TF_DeleteSessionOptions(opts)
E           
E           InternalError: cudaGetDevice() failed. Status: CUDA driver version is insufficient for CUDA runtime version
```

It means CUDA driver version used in the VM is lower than the requirement of TensorFlow.  I am trying to figure out how to do that.  @miguelgfierro @anargyri

tests/security/test_dependency_security.py
anargyri: Their dependencies are minimal, see https://github.com/pypa/packaging/blob/4d8534061364e3cbfee582192ab81a095ec2db51/pyproject.toml, so I don't think we should worry so much about this package causing problems.

tests/ci/azureml_tests/test_groups.py
anargyri: Actually, I don't see any class that uses `SUMModel` (which is the class that causes the error). There are a couple of tests that reference it (including the one above), however. So maybe it is not a dependency and has been left somehow in the code? If so, then we could simply remove the tests without breaking any of the notebooks.

anargyri:
> When using the latest TensorFlow (2.15), I got [this error](https://github.com/recommenders-team/recommenders/actions/runs/7711491218/job/21048241208#step:3:707):
> 
> ```
> E               713 try:
> E               714   # pylint: disable=protected-access
> E               715   with self._graph._c_graph.get() as c_graph:
> E           --> 716     self._session = tf_session.TF_NewSessionRef(c_graph,opts)
> E               717   # pylint: enable=protected-access
> E               718 finally:
> E               719   tf_session.TF_DeleteSessionOptions(opts)
> E           
> E           InternalError: cudaGetDevice() failed. Status: CUDA driver version is insufficient for CUDA runtime version
> ```
> 
> It means CUDA driver version used in the VM is lower than the requirement of TensorFlow. I am trying to figure out how to do that. @miguelgfierro @anargyri

Yes, because the latest TF uses cuda 12, all previous ones used cuda 11 or less. See https://www.tensorflow.org/install/source#gpu

miguelgfierro:
@SimonYansenZhao if we upgrade the docker image in the test to cuda 12, will the code with previous versions of cuda 11 work?

tests/security/test_dependency_security.py
miguelgfierro: ok, let's hope this doesn't break in a few years

tests/ci/azureml_tests/test_groups.py
SimonYansenZhao: Uncommeted

tests/ci/azureml_tests/submit_groupwise_azureml_pytest.py
```diff
@@ -178,7 +178,38 @@ def create_run_config(
...
+
+            USER root:root
+
+            ENV NVIDIA_VISIBLE_DEVICES all
+            ENV NVIDIA_DRIVER_CAPABILITIES compute,utility
+            ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
+            ENV DEBIAN_FRONTEND noninteractive
+            
+            # Conda Environment
+            ENV MINICONDA_VERSION py38_23.3.1-0
```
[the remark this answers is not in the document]
SimonYansenZhao: This installs a base Conda env in the Docker image.  `conda_pkg_python` is used to specify the Conda env for testing Recommenders, so when testing, a new Conda environment will be created.  They are different envs.

SimonYansenZhao:
> @SimonYansenZhao if we upgrade the docker image in the test to cuda 12, will the code with previous versions of cuda 11 work?

I think it will work.  Only the NVIDIA GPU Driver in the CUDA installed in the Docker image is used when testing, because a Conda env will be created with a CUDA required by Recommenders, the CUDA in the Conda env will not conflict with the CUDA in the Docker image.  The CUDA installed in the Conda env will interact with the driver which is back compatible.

miguelgfierro (approved):
Amazing work Simon!

PR #2075, opened 2024-03-20
Merged two concats into one

daviddavo:
### Description

Just replaced with a oneliner to make just one concat instead of two

### Related Issues

- https://github.com/recommenders-team/recommenders/pull/1937#discussion_r1530181712

### References

> [...] If function is None, the identity function is assumed, that is, all elements of iterable that are false are removed.
> -- https://docs.python.org/3.9/library/functions.html#filter

### Checklist:

- [x] I have followed the [contribution guidelines](CONTRIBUTING.md) and code style for this project.
- [ ] I have added tests covering my contributions. (not needed)
- [ ] I have updated the documentation accordingly. (not needed)

### DIFF
--- a/recommenders/models/rlrmc/RLRMCdataset.py
+++ b/recommenders/models/rlrmc/RLRMCdataset.py
@@ -68,8 +68,7 @@ class RLRMCdataset(object):
         """
         # Data processing and reindexing code is adopted from https://github.com/Microsoft/Recommenders/blob/main/recommenders/models/ncf/dataset.py
         # If validation dataset is None
-        df = train if validation is None else pd.concat([train, validation])
-        df = df if test is None else pd.concat([df, test])
+        df = pd.concat(filter(lambda x: x is not None, [train, validation, test]))
 
         # Reindex user and item index
         if self.user_idx is None:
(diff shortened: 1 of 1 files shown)


recommenders/models/rlrmc/RLRMCdataset.py
```diff
@@ -68,8 +68,7 @@ def _data_processing(self, train, validation=None, test=None, mean_center=True):
         """
         # Data processing and reindexing code is adopted from https://github.com/Microsoft/Recommenders/blob/main/recommenders/models/ncf/dataset.py
         # If validation dataset is None
-        df = train if validation is None else pd.concat([train, validation])
-        df = df if test is None else pd.concat([df, test])
+        df = pd.concat(filter(lambda x: x is not None, [train, validation, test]))
```
loomlike, line 71: `concat` handles `None` nicely, so that you can simply:
`df = pd.concat([train, test, validation])`

loomlike (commented):
Thank you for making the code clearer and simpler.
I have an alternative suggestion, please check.

recommenders/models/rlrmc/RLRMCdataset.py
daviddavo: > Any None objects will be dropped silently unless they are all None in which case a ValueError will be raised.
It's true! Train won't ever be none so it will never fail.

miguelgfierro (approved):
Thanks for this @daviddavo! Super nice!

Issue #2044, opened 2023-12-19
[FEATURE] Support `steps` to newsrec models


loomlike:
### Description
Currently `newsrec` models only support `epoch` training and the smoke-tests with `epoch = 1` take minutes for each notebook.
With a simple tweak to `base_model.fit()`, we can introduce `steps` parameters that constraint the number of batches to train.
Then, we can set `steps = 1` to the smoke test which will expediate the test time.

### Expected behavior with the suggested feature
Use `steps = 1` to all the smoke tests of the newsrec models.

sumana-2705:
@loomlike @miguelgfierro 
Shall I work on this issue?

miguelgfierro:
feel free

miguelgfierro:
right now, we have a bug that prevents external contributors from running the tests. It is a security limit from MS. We need to find a workaround. See #1853

sumana-2705:
okayy, I understood. Thanks for your reply @miguelgfierro

anargyri:
@loomlike @sumana-2705 should this issue be closed now?

sumana-2705:
Yeah, I guess It can be closed now

PR #2125, opened 2024-07-04
Modified 2 files to update newsrec model

sumana-2705:
### Description

I have modified the file, `base_model.py` in which I added a parameter named step_limit(which is optional) into the funtion `fit` and also modified the file `test_newsrec_model.py`, I have adjusted step_limit=1 which speeds up the smoke test.

### Related Issues

It fixes the issue #2044 

### References

I have done changes as mentioned in the issue

### Checklist:

- [x] I have followed the [contribution guidelines](CONTRIBUTING.md) and code style for this project.

### DIFF
--- a/recommenders/models/newsrec/models/base_model.py
+++ b/recommenders/models/newsrec/models/base_model.py
@@ -186,6 +186,8 @@ class BaseModel:
         valid_behaviors_file,
         test_news_file=None,
         test_behaviors_file=None,
+        steps=None,
+
     ):
         """Fit the model with train_file. Evaluate the model on valid_file per epoch to observe the training status.
         If test_news_file is not None, evaluate it too.
@@ -212,6 +214,8 @@ class BaseModel:
             )
 
             for batch_data_input in tqdm_util:
+                if steps is not None and steps>=steps:
+                    break
 
                 step_result = self.train(batch_data_input)
                 step_data_loss = step_result
--- a/recommenders/models/newsrec/models/base_model.py
+++ b/recommenders/models/newsrec/models/base_model.py
@@ -186,7 +186,7 @@ class BaseModel:
         valid_behaviors_file,
         test_news_file=None,
         test_behaviors_file=None,
-        steps=None,
+        step_limit=None,
 
     ):
         """Fit the model with train_file. Evaluate the model on valid_file per epoch to observe the training status.
@@ -214,7 +214,7 @@ class BaseModel:
             )
 
             for batch_data_input in tqdm_util:
-                if steps is not None and steps>=steps:
+                if step_limit is not None and step_limit>=step_limit:
                     break
 
                 step_result = self.train(batch_data_input)
--- a/recommenders/models/newsrec/models/base_model.py
+++ b/recommenders/models/newsrec/models/base_model.py
@@ -214,7 +214,7 @@ class BaseModel:
             )
 
             for batch_data_input in tqdm_util:
-                if step_limit is not None and step_limit>=step_limit:
+                if step_limit is not None and step>=step_limit:
                     break
 
                 step_result = self.train(batch_data_input)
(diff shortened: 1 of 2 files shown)


recommenders/models/newsrec/models/base_model.py
```diff
@@ -186,6 +186,8 @@ def fit(
         valid_behaviors_file,
         test_news_file=None,
         test_behaviors_file=None,
+        steps=None,
```
miguelgfierro, line 189: I think it would be clearer if you call it `step_limit`, otherwise your step will be confused with the step defined in the loop. Also, please add a comment in the docstring defining the step limit

tests/smoke/recommenders/recommender/test_newsrec_model.py
```diff
@@ -115,7 +115,7 @@ def test_model_naml(mind_resource_path):
     assert model.run_eval(valid_news_file, valid_behaviors_file) is not None
     assert isinstance(
         model.fit(
-            train_news_file, train_behaviors_file, valid_news_file, valid_behaviors_file
+            train_news_file, train_behaviors_file, valid_news_file, valid_behaviors_file,steps=1
```
miguelgfierro, line 118: can you check how long it takes with step limit of 1 versus 5 or other? it would be good if you share the details

recommenders/models/newsrec/models/base_model.py
sumana-2705: okay, sure. I will add it right now

tests/smoke/recommenders/recommender/test_newsrec_model.py
sumana-2705: I have checked time taken for different step_limit, here are the outputs

for step_limit = 1

>> Output:
[Done] exited with code=0 in 0.614 seconds

for step_limit = 3
>> Output:
[Done] exited with code=0 in 0.677 seconds 

for step_limit = 5
>> Output:
[Done] exited with code=0 in 0.984 seconds

for step_limit = 7
>> Output:
[Done] exited with code=0 in 0.488 seconds

for step_limit = 10
>> Output:
[Done] exited with code=0 in 0.846 seconds

- step_limit=7 has taken lesser time than 1

recommenders/models/newsrec/models/base_model.py
```diff
@@ -212,6 +214,8 @@ def fit(
             )
 
             for batch_data_input in tqdm_util:
+                if step_limit is not None and step_limit>=step_limit:
```
miguelgfierro, line 217:
```suggestion
                if step_limit is not None and step>=step_limit:
```
I think you mean here this? You were comparing step_limit with itself which is always the same

tests/smoke/recommenders/recommender/test_newsrec_model.py
```diff
@@ -115,7 +115,7 @@ def test_model_naml(mind_resource_path):
     assert model.run_eval(valid_news_file, valid_behaviors_file) is not None
     assert isinstance(
         model.fit(
-            train_news_file, train_behaviors_file, valid_news_file, valid_behaviors_file
+            train_news_file, train_behaviors_file, valid_news_file, valid_behaviors_file,step_limit=1
```
miguelgfierro, line 118: If 10 steps is less than a second, can we use that instead of 1?

sumana-2705: ofcourse, we can use 10 instead of 1,but as per the output step_limit = 7 is still more preferrable right.

miguelgfierro: I wouldn't take the times too strictly because they could change in other computers. Feel free to choose a number that is round (5, 10, 20) and that doesn't take to long

sumana-2705: okay

miguelgfierro (approved):
Great!

PR #2228, opened 2025-05-02
Feature/pytorch embedding ranker v4

demoncoder-crypto:
### Checklist:

- [ ] I have followed the [contribution guidelines](CONTRIBUTING.md) and code style for this project.
- [ ] I have [signed the commits](https://github.com/recommenders-team/recommenders/wiki/How-to-sign-commits), e.g. `git commit -s -m "your commit message"`. 
- [ ] This PR is being made to `staging branch` AND NOT TO `main branch`.


examples/00_quick_start/nn_embedding_ranker_movielens.ipynb
```diff
@@ -0,0 +1,10 @@
+{
```
miguelgfierro, line 1: not sure if there was an error in the submission of the notebook, but I can't see it

recommenders/evaluation/spark_evaluation.py
```diff
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 import numpy as np
+import warnings # Added for R-Precision warning
```
miguelgfierro, line 5: this file is still from the other PR, it needs to be removed

recommenders/models/embedding_ranker/embedding_ranker_utils.py
```diff
@@ -0,0 +1,208 @@
...
+    batch_size=1024,
+):
+    """Predict ratings for user-item pairs in test data.
+
+    Args:
+        model (NNEmbeddingRanker): Trained embedding ranker model.
+        test_df (pandas.DataFrame): Test dataframe containing user-item pairs.
+        col_user (str): User column name.
+        col_item (str): Item column name.
+        col_rating (str): Rating column name.
```
miguelgfierro, line 33: I think all this can become the notebook. 

Some thoughts for the notebook:
- Here is a good example of a useful notebook: https://github.com/recommenders-team/recommenders/blob/main/examples/02_model_collaborative_filtering/lightgcn_deep_dive.ipynb it explains both the math behind it and an implementation
- Think of the notebook as a way to showcase how to use the ranker and what is the ranker about.
- The objective of the notebook is that it needs to be useful. That's the most important metric.
- Ideally, a person could go to this notebook, add their data, run it, and understand how to use it.
- For the notebook, you can just showcase the content of these functions directly inside the notebook. 
- Something very important is that we follow the principle of explicit is better than implicit. For example, we don't something like `for metric_name, metric_func in metrics.items()`, because it adds a layer of complexity. Instead we show the metrics explicitely:  `rmse(true, pred)`, `precision_at_k(true, pred, params)`, etc. Each person with a quick view can see what is going on.
- Feel free to come to our Monday meeting if you want to understand better how we do the notebooks @demoncoder-crypto

miguelgfierro:
@setuc FYI

miguelgfierro:
@demoncoder-crypto how is this work going? 
@setuc can support

miguelgfierro:
Closing this since it doesn't seem in development and there is already a very similar implementation: https://github.com/recommenders-team/recommenders/pull/2238

miguelgfierro:
@jmarrietar do you think you would be able to take over this work? It is very similar to embdotbias

PR #2278, opened 2026-02-11
Fix ZeroDivisionError in item feature cosine similarity

Mr-Neutr0n:
## Summary
- The `_get_item_feature_similarity` function in `python_evaluation.py` computes cosine similarity via `dot(f1, f2) / (norm(f1) * norm(f2))`. When either feature vector is a zero vector (all elements are zero), `np.linalg.norm` returns 0 and the division raises a `ZeroDivisionError`.
- This fix checks for zero-norm vectors before dividing and returns 0.0, which is the standard mathematical convention for cosine similarity involving zero vectors.

## Details
The bug is in `_get_item_feature_similarity()` at `recommenders/evaluation/python_evaluation.py`. The original code uses a lambda that unconditionally divides by the product of the two vector norms:

```python
lambda x: float(x.f1.dot(x.f2))
    / float(np.linalg.norm(x.f1, 2) * np.linalg.norm(x.f2, 2))
```

If an item has a zero feature vector (e.g., a new/cold-start item with no features populated), this crashes at runtime. The fix extracts the lambda into a named function that guards against zero norms:

```python
def _cosine_sim(x):
    norm_f1 = np.linalg.norm(x.f1, 2)
    norm_f2 = np.linalg.norm(x.f2, 2)
    if norm_f1 == 0 or norm_f2 == 0:
        return 0.0
    return float(x.f1.dot(x.f2)) / float(norm_f1 * norm_f2)
```

## Test plan
- [x] Verified the fix handles zero-norm vectors by returning 0.0
- [x] Verified normal (non-zero) vectors still produce correct cosine similarity values
- [x] No other code paths are affected; the change is isolated to `_get_item_feature_similarity`


miguelgfierro (approved):
Good catch, thanks @Mr-Neutr0n

PR #2279, opened 2026-02-11
Fix mutable default argument in Keras callback methods

Mr-Neutr0n:
## Summary

- Replace `logs={}` with `logs=None` and add `if logs is None: logs = {}` guard in all Keras callback methods in `multinomial_vae.py` and `standard_vae.py`
- Using a mutable default argument like `{}` is a well-known Python anti-pattern ([W0102](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/dangerous-default-value.html)) — the same dict object is shared across all calls, which can lead to unexpected state leakage between invocations
- This also aligns with the Keras `Callback` base class, which uses `logs=None` as the default

## Changes

**`recommenders/models/vae/multinomial_vae.py`** — 7 instances fixed:
- `LossHistory.on_train_begin` (line 19)
- `LossHistory.on_epoch_end` (line 24)
- `Metrics.on_train_begin` (line 65)
- `Metrics.on_epoch_end` (line 103)
- `AnnealingCallback.on_train_begin` (line 161)
- `AnnealingCallback.on_batch_end` (line 165)
- `AnnealingCallback.on_epoch_end` (line 175)

**`recommenders/models/vae/standard_vae.py`** — 7 instances fixed:
- `LossHistory.on_train_begin` (line 20)
- `LossHistory.on_epoch_end` (line 25)
- `Metrics.on_train_begin` (line 66)
- `Metrics.on_epoch_end` (line 104)
- `AnnealingCallback.on_train_begin` (line 161)
- `AnnealingCallback.on_batch_end` (line 165)
- `AnnealingCallback.on_epoch_end` (line 175)

## Test plan

- [ ] Verify all `logs={}` replaced with `logs=None` + guard
- [ ] No functional change — behavior is identical when `logs` is passed by caller (which Keras always does)
- [ ] Existing VAE notebook tests should pass without modification

### DIFF
--- a/recommenders/models/vae/multinomial_vae.py
+++ b/recommenders/models/vae/multinomial_vae.py
@@ -16,13 +16,17 @@ from tensorflow.keras.callbacks import ReduceLROnPlateau, Callback
 class LossHistory(Callback):
     """This class is used for saving the validation loss and the training loss per epoch."""
 
-    def on_train_begin(self, logs={}):
+    def on_train_begin(self, logs=None):
         """Initialise the lists where the loss of training and validation will be saved."""
+        if logs is None:
+            logs = {}
         self.losses = []
         self.val_losses = []
 
-    def on_epoch_end(self, epoch, logs={}):
+    def on_epoch_end(self, epoch, logs=None):
         """Save the loss of training and validation set at the end of each epoch."""
+        if logs is None:
+            logs = {}
         self.losses.append(logs.get("loss"))
         self.val_losses.append(logs.get("val_loss"))
 
@@ -62,8 +66,10 @@ class Metrics(Callback):
         # Options to save the weights of the model for future use
         self.save_path = save_path
 
-    def on_train_begin(self, logs={}):
+    def on_train_begin(self, logs=None):
         """Initialise the list for validation NDCG@k."""
+        if logs is None:
+            logs = {}
         self._data = []
 
     def recommend_k_items(self, x, k, remove_seen=True):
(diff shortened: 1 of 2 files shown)


recommenders/models/vae/multinomial_vae.py
```diff
@@ -16,13 +16,17 @@
 class LossHistory(Callback):
     """This class is used for saving the validation loss and the training loss per epoch."""
 
-    def on_train_begin(self, logs={}):
+    def on_train_begin(self, logs=None):
         """Initialise the lists where the loss of training and validation will be saved."""
+        if logs is None:
+            logs = {}
```
miguelgfierro:
```suggestion
        logs = logs or {}    
```

miguelgfierro: @Mr-Neutr0n this is good thanks for the notice. I added a suggestion to make the change one liner, can you please update?

recommenders/models/vae/multinomial_vae.py
```diff
@@ -16,13 +16,17 @@
 class LossHistory(Callback):
     """This class is used for saving the validation loss and the training loss per epoch."""
 
-    def on_train_begin(self, logs={}):
+    def on_train_begin(self, logs=None):
```
miguelgfierro, line 19: @Mr-Neutr0n notice that all the PRs should be to staging, not to main. See https://github.com/recommenders-team/recommenders/blob/main/CONTRIBUTING.md#steps-to-contributing

recommenders/models/vae/multinomial_vae.py
Mr-Neutr0n: good call, simplified to the one-liner. also added the DCO sign-off

miguelgfierro (approved):
Thanks!

PR #2288, opened 2026-03-06
feat: add lightgbm ranker

ds-wook:
#### Summary                                                                                    
          
  - Added a new quickstart notebook `examples/00_quick_start/lightgbm_movielens.ipynb`         
  demonstrating Learning-to-Rank with LightGBM (lgb.LGBMRanker) using the MovieLens 1M       
  dataset                                                                                    
  - Used `lgb.LGBMRanker` directly (no wrapper class) to show the library's native API as
  intended for an example notebook                                                           
  - Extended `recommenders/datasets/pandas_df_utils.py` to support negative sampling, which is
  required for meaningful NDCG evaluation                                                    
  - Wired smoke and functional tests into the existing test framework
 
#### Key Design Decisions       
                                                                                             
  - User-relative label binarization: Relevance labels are computed relative to each user's  
  own rating distribution, avoiding the bias of a fixed threshold
  - Negative sampling at evaluation: Unobserved items are included in the candidate pool so  
  NDCG is not trivially inflated                                                             
  - No data leakage: User/item statistics (`mean`, `count`, `std`) are computed from training data
  only, then joined to validation/test sets                                                  
  - Lambdarank objective with `mAP@10` and `NDCG@10` as evaluation metrics

#### Tests
 - Unit test: Runs the notebook end-to-end with mock100 synthetic data. load_item_df previously only supported real dataset sizes (100k, 1m, etc.), so mock100 support was added to allow unit tests to run without downloading data.
  - Smoke test: Runs the notebook end-to-end on a small data sample to verify it executes without errors
  - Functional test: Validates that map_at_10 and ndcg_at_10 meet expected values on the full
   MovieLens 1M dataset         
### References
- [Leveraging LightGBM Ranker for Efficient Large-Scale News Recommendation Systems - Recsys Challenge 2024](https://dl.acm.org/doi/pdf/10.1145/3687151.3687156)
- [Are Neural Rankers still Outperformed by Gradient Boosted Decision Trees? - ICLR 2021](https://openreview.net/pdf?id=Ut1vF_q_vC)

### Checklist:
* [x]  I have followed the [contribution guidelines](https://github.com/recommenders-team/recommenders/blob/main/CONTRIBUTING.md) and code style for this project.
* [x]  I have added tests covering my contributions.
* [ ]  I have updated the documentation accordingly.
* [x]  I have [signed the commits](https://github.com/recommenders-team/recommenders/wiki/How-to-sign-commits), e.g. `git commit -s -m "your commit message"`.
* [x]  This PR is being made to `staging branch` AND NOT TO `main branch`.


recommenders/models/lightgbm/lightgbm_utils.py
```diff
@@ -205,3 +207,161 @@ def transform(self, df):
...
+
+    Wraps LightGBM training, inference, save, and load into a single class
+    that follows a sklearn-like interface. Supports both binary classification
+    (CTR prediction) and learning-to-rank objectives (lambdarank, etc.).
+
+    Example::
+
+        ranker = LightGBMRanker(params={"objective": "binary", "metric": "auc"})
+        ranker.fit(train_x, train_y, valid_x, valid_y)
+        scores = ranker.predict(test_x)
```
miguelgfierro: This shouldn't go to the repo, we should use directly the `lgb.LGBMRanker` instead of this class. I see value though in the notebook.

examples/00_quick_start/lightgbm_ranker.ipynb
```diff
@@ -0,0 +1,731 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# LightGBMRanker: Learning-to-Rank with LambdaRank\n",
+    "\n",
+    "This notebook demonstrates how to build a learning-to-rank pipeline using `LightGBMRanker` with the `lambdarank` objective, which optimises NDCG directly.\n",
+    "\n",
+    "We use the **MovieLens 100k** dataset: each user is treated as a *query*, and their rated movies are the *candidate items*. The rating (1–5) serves as the graded relevance label.\n",
```
miguelgfierro: I see value in the example of learning to rank with lightgbm, but using the underlying code of lightgbm instead of the class proposed.

examples/00_quick_start/lightgbm_ranker.ipynb
```diff
@@ -0,0 +1,731 @@
...
+       "      <th>score</th>\n",
+       "      <th>label</th>\n",
+       "    </tr>\n",
+       "  </thead>\n",
+       "  <tbody>\n",
+       "    <tr>\n",
+       "      <th>0</th>\n",
+       "      <td>302</td>\n",
+       "      <td>0.275522</td>\n",
+       "      <td>5.0</td>\n",
```
miguelgfierro: The reported NDCG@5 of 0.91 and NDCG@10 of 0.93 look suspiciously high for MovieLens 100k with simple features. The
  issue is in how the evaluation is set up, not in the model itself.

  Two compounding problems:

  1. No negative items in the evaluation pool. The test set only contains movies the user actually rated. There are no
  unobserved items as candidates, so the model only needs to rank ~10-20 positives against each other — it never has to
  distinguish good items from bad ones among thousands of candidates.
  2. All labels are non-zero (ratings 1–5). Since every item in the pool has relevance > 0, even a random ordering
  produces a high NDCG. The ideal DCG and the actual DCG are close because there's no "irrelevant junk" dragging the
  score down.

  To fix this, one of the following approaches should be used:

  - Negative sampling: For each user, sample N items they did not interact with (label = 0) and add them to the
  evaluation pool. This forces the model to rank the few relevant items above many irrelevant ones, which is the actual
  recommendation task. The repo already has negative_feedback_sampler() in recommenders/datasets/pandas_df_utils.py and
  user_item_pairs() with a filter to exclude observed interactions — both can be used directly.
  - Leave-one-out with sampled negatives: Hold out one positive item per user, pair it with e.g. 100 random negatives,
  and measure whether the model ranks the positive item near the top. The NCF Dataset class in
  recommenders/models/ncf/dataset.py implements exactly this pattern (see n_neg_test parameter and _create_test_file()).
  - At minimum, binarize the labels: Treat ratings ≥ 4 as relevant (1) and the rest as irrelevant (0), so that low-rated
   items contribute zero gain to DCG.

  Without one of these corrections, the NDCG numbers don't reflect real ranking quality and aren't comparable to other
  models in the repo.

miguelgfierro (changes requested):
@ds-wook the initiative is good. I think it has value to create a ranker with movilens, however, we should use the original class instead of the class created, and you need to review the NDCG which is incorrect.
But overall, good contribution. 
Please notice that all the notebooks have the pattern name model_dataset. So this one should be lightgbm_ranker_movielens or lightgbm_movielens

ds-wook:
@miguelgfierro 
Thank you for the feedback and for reviewing my contribution.

I initially thought the main goal of this project was to implement LightGBM-related functionality within the repo, which is why I created a separate wrapper class. If that is not the intended direction, I will update the implementation accordingly and remove the wrapper.

For the example notebook, I understood it mainly as a way to verify the behavior of the wrapper class I implemented. Based on your comments, I will revise it as well.

The changes I plan to make are:

1. Update the code to use the native `lightgbm.LGBMRanker` from the LightGBM library directly.
2. Rework the evaluation by introducing negative sampling so that the NDCG calculation reflects a realistic recommendation setting.

I will proceed with these updates.
Thank you again for the helpful feedback.

ds-wook:
## Changed Files                                                                             
   
  | File | Change |                                                                            
  |---|---|       
  | `recommenders/models/lightgbm/lightgbm_utils.py` | Removed `LightGBMRanker` class and
  unused `import lightgbm` |
  | `recommenders/models/lightgbm/__init__.py` | Removed `LightGBMRanker` re-export |
  | `tests/unit/recommenders/models/test_lightgbm_utils.py` | Deleted (only tested the removed
  `LightGBMRanker` class) |
  | `examples/00_quick_start/lightgbm_movielens.ipynb` | New notebook (replaces
  `lightgbm_ranker.ipynb`) |

  ---

  ## lightgbm_movielens.ipynb Overview

  ### Purpose
  End-to-end learning-to-rank pipeline using `lgb.LGBMRanker` directly (lambdarank objective),
  replacing the custom `LightGBMRanker` wrapper that was previously in the repo.

  ### Dataset
  - MovieLens **1M**
  - Per-user chronological split: train 80% / valid 10% / test 10% via `python_chrono_split`

  ### Label Binarization (User-relative)
  rating > user_mean_rating  →  1  (above this user's personal average)
  rating ≤ user_mean_rating  →  0
  A user-relative threshold is used instead of a fixed cut-off (e.g. ≥ 4),
  so the label reflects whether a movie was above-average *for that specific user*,
  accounting for individual rating tendencies.

  ### Negative Sampling
  LambdaRank requires both positive and negative items within each query group
  to produce a meaningful ranking signal.

  | Stage | Strategy |
  |---|---|
  | Train | Add `N_NEG_TRAIN=100` unobserved items per user |
  | Validation | Add `N_NEG_VALID=100` unobserved items per user (stabilises early stopping) |
  | Evaluation | Add `N_NEG_EVAL=100` unobserved items per user to the evaluation pool |

  ### Feature Engineering (computed from training data only — no leakage)

  | Feature | Description |
  |---|---|
  | `user_mean_rating` | User's average rating |
  | `user_rating_count` | Number of ratings by this user |
  | `user_rating_std` | Std-dev of user ratings (rating consistency) |
  | `item_mean_rating` | Item's average rating |
  | `item_rating_count` | Number of ratings for this item (popularity) |
  | `item_rating_std` | Std-dev of item ratings (polarisation) |
  | `user_genre_mean_rating` | User's average rating for this item's primary genre |
  | `user_genre_count` | Number of items this user has rated in this genre |
  | `release_year` | Release year extracted from the movie title |
  | `main_genre` | Primary genre (LightGBM native categorical feature) |

  ### Model
  ```python
  LGBMRanker(
      objective="lambdarank",
      metric="ndcg",
      ndcg_eval_at=[5, 10],
      num_leaves=64,
      learning_rate=0.05,
      feature_fraction=0.8,
      n_estimators=10000,  # with early stopping
  )
```

**Comment by ds-wook** [comment]
### Changes                                                                                  
                                                                                               
  - `class NumEncoder(object)` → `class NumEncoder` (Python 3 style, no explicit `object` base)
  - Added `import pandas as pd` and `from collections import defaultdict` required for       
  annotations
  - Added type hints to all public functions and methods using Python 3.10+ built-in generics
  (no `typing` imports needed)

  ### Annotated signatures

  ```python
  def unpackbits(x: np.ndarray, num_bits: int) -> np.ndarray

  class NumEncoder:
      def __init__(
          self,
          cate_cols: list[str],
          nume_cols: list[str],
          label_col: str,
          threshold: int = 10,
          thresrate: float = 0.99,
      ) -> None

      def fit_transform(self, df: pd.DataFrame) -> tuple[np.ndarray, np.ndarray]
      def transform(self, df: pd.DataFrame) -> tuple[np.ndarray, np.ndarray]

  Instance variable annotations

  ┌───────────────────┬─────────────────────────────┐
  │     Variable      │            Type             │
  ├───────────────────┼─────────────────────────────┤
  │ dtype_dict        │ dict[str, str]              │
  ├───────────────────┼─────────────────────────────┤
  │ tgt_nume_cols     │ list[str]                   │
  ├───────────────────┼─────────────────────────────┤
  │ save_cate_avgs    │ dict[str, defaultdict]      │
  ├───────────────────┼─────────────────────────────┤
  │ save_value_filter │ dict[str, list]             │
  ├───────────────────┼─────────────────────────────┤
  │ save_num_embs     │ dict[str, dict[str, float]] │
  ├───────────────────┼─────────────────────────────┤
  │ Max_len           │ dict[str, int]              │
  ├───────────────────┼─────────────────────────────┤
  │ samples           │ int                         │
  └───────────────────┴─────────────────────────────┘

  Local variable annotations

  ┌───────────────┬────────────────────────┐
  │   Variable    │          Type          │
  ├───────────────┼────────────────────────┤
  │ rows          │ np.ndarray | None      │
  ├───────────────┼────────────────────────┤
  │ feat_encoding │ dict[str, list[float]] │
  └───────────────┴────────────────────────┘

  Python version requirements

  All annotations use syntax available natively in Python 3.10+:
  - Built-in generics (list[str], dict[str, str], tuple[...]) — no typing.List / typing.Dict
  - Union with | operator (np.ndarray | None) — no typing.Optional

File: examples/00_quick_start/lightgbm_movielens.ipynb [hunk 4]
```diff
@@ -0,0 +1,1981 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "cell-params",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Data settings\n",
+    "SIZE = \"1m\"\n",
```
miguelgfierro: in the rest of the notebooks we use:
```
# Select MovieLens data size: 100k, 1m, 10m, or 20m
MOVIELENS_DATA_SIZE = "100k"
```
instead of `SIZE`

tests/functional/examples/test_notebooks_python.py
```diff
@@ -264,6 +264,23 @@ def test_xlearn_fm_functional(notebooks, output_notebook, kernel_name):
...
+        notebook_path,
+        output_notebook,
+        kernel_name=kernel_name,
+        parameters=dict(SIZE=size),
+    )
+    results = read_notebook(output_notebook)
+
+    assert results["map"] >= 0.0
+    assert results["ndcg_at_5"] >= 0.0
+    assert results["ndcg_at_10"] >= 0.0
```
miguelgfierro: The way smoke vs functional works is, smoke is a quick test that goes in sequence with functional. So what you can do is to add here `MOVIELENS_DATA_SIZE = "1m"` and in the smoke: `MOVIELENS_DATA_SIZE = "100k"`

Also you need to add the approximate value in the asserts, notice how we do it in the other tests. The reason we are doing this is because we don't want to just make sure that the notebooks run, but that they provide values that are correct.

tests/smoke/examples/test_notebooks_python.py
```diff
@@ -117,6 +117,22 @@ def test_lightgbm_quickstart_smoke(notebooks, output_notebook, kernel_name):
...
+        output_notebook,
+        kernel_name=kernel_name,
+        parameters=dict(SIZE="sample"),
+    )
+    results = read_notebook(output_notebook)
+
+    assert results["map"] >= 0.0
+    assert results["ndcg_at_5"] >= 0.0
+    assert results["ndcg_at_10"] >= 0.0
+
```
miguelgfierro, line 133: The way smoke vs functional works is, smoke is a quick test that goes in sequence with functional. So what you can do is to add here `MOVIELENS_DATA_SIZE = "1m"` and in the smoke: `MOVIELENS_DATA_SIZE = "100k"`

Also you need to add the approximate value in the asserts, notice how we do it in the other tests. The reason we are doing this is because we don't want to just make sure that the notebooks run, but that they provide values that are correct.

tests/functional/examples/test_notebooks_python.py
```diff
@@ -264,6 +264,23 @@ def test_xlearn_fm_functional(notebooks, output_notebook, kernel_name):
     assert results["auc_score"] == pytest.approx(0.75, rel=TOL, abs=ABS_TOL)
 
 
+@pytest.mark.notebooks
+@pytest.mark.parametrize("size", ["sample"])
```
miguelgfierro: there's no need to use parametrize if you don't want to do multiple tests with different parameters

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "cell-sec1",
+   "metadata": {},
+   "source": [
+    "## 1. Data Preparation\n",
+    "\n",
+    "We use **MovieLens 1M**, which provides real user–item interactions with ratings 1–5.\n",
+    "\n",
+    "**LTR framing:**\n",
```
miguelgfierro: what is LTR?

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+    "        _all_items,\n",
+    "        user_col=USER_COL,\n",
+    "        item_col=ITEM_COL,\n",
+    "        user_item_filter_df=_observed_pairs,\n",
+    "        shuffle=True,\n",
+    "        seed=SEED,\n",
+    "    )\n",
+    "    neg_df = (\n",
+    "        candidates.groupby(USER_COL, group_keys=False)\n",
+    "        .apply(lambda g: g.head(n_neg))\n",
```
miguelgfierro: The notebook defines a custom `sample_negatives()` function, but the repo already provides [`negative_feedback_sampler()`](https://github.com/recommenders-team/recommenders/blob/staging/recommenders/datasets/pandas_df_utils.py#L267) in `recommenders/datasets/pandas_df_utils.py` which does the same thing. We should use that instead of adding custom code.

The only gap is that `negative_feedback_sampler` uses a ratio (`ratio_neg_per_user`) rather than a fixed count. To support your use case, add an `n_neg_per_user` parameter to the existing function. The change is minimal:

**In `recommenders/datasets/pandas_df_utils.py`:**

**1)** Add the parameter to the signature (after `ratio_neg_per_user`):

```python
def negative_feedback_sampler(
    df,
    col_user=DEFAULT_USER_COL,
    col_item=DEFAULT_ITEM_COL,
    col_label=DEFAULT_LABEL_COL,
    col_feedback="feedback",
    ratio_neg_per_user=1,
    n_neg_per_user=None,  # <-- new
    pos_value=1,
    neg_value=0,
    seed=42,
):
```

**2)** Replace the `neg_sample_size` line inside `sample_items`:

```python
# before
neg_sample_size = max(round(n_u * ratio_neg_per_user), 1)

# after
neg_sample_size = (
    n_neg_per_user
    if n_neg_per_user is not None
    else max(round(n_u * ratio_neg_per_user), 1)
)
```

No other changes needed — the existing `min()` and slice already cap at available items.

**In the notebook**, replace the custom `sample_negatives()` with:

```python
from recommenders.datasets.pandas_df_utils import negative_feedback_sampler

neg_train = negative_feedback_sampler(
    train_raw, col_user=USER_COL, col_item=ITEM_COL,
    col_feedback=LABEL_COL, n_neg_per_user=N_NEG_TRAIN, seed=SEED,
)
```

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+      "[LightGBM] [Warning] feature_fraction is set=0.8, colsample_bytree=1.0 will be ignored. Current value: feature_fraction=0.8\n",
+      "[LightGBM] [Warning] feature_fraction is set=0.8, colsample_bytree=1.0 will be ignored. Current value: feature_fraction=0.8\n",
+      "[LightGBM] [Info] Auto-choosing row-wise multi-threading, the overhead of testing was 0.018705 seconds.\n",
+      "You can set `force_row_wise=true` to remove the overhead.\n",
+      "And if memory is not enough, you can set `force_col_wise=true`.\n",
+      "[LightGBM] [Info] Total Bins 2653\n",
+      "[LightGBM] [Info] Number of data points in the train set: 1404193, number of used features: 17\n",
+      "[LightGBM] [Warning] feature_fraction is set=0.8, colsample_bytree=1.0 will be ignored. Current value: feature_fraction=0.8\n",
+      "Training until validation scores don't improve for 100 rounds\n",
+      "[100]\tvalid_0's ndcg@5: 0.3973\tvalid_0's ndcg@10: 0.433501\n",
```
miguelgfierro: There is no need to do NDCG@5, with NDCG@10 is enough

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+    "CAT_FEATURE_IDX = train_x.shape[1]  # index of the appended genre column\n",
+    "\n",
+    "\n",
+    "def append_genre(x: np.ndarray, df: pd.DataFrame) -> np.ndarray:\n",
+    "    codes = df[\"main_genre\"].map(genre_map).fillna(-1).astype(int).values.reshape(-1, 1)\n",
+    "    return np.hstack([x, codes])\n",
+    "\n",
+    "\n",
+    "train_x = append_genre(train_x, train_data)\n",
+    "valid_x = append_genre(valid_x, valid_data)\n",
```
miguelgfierro, line 452: The notebook cells execute in order: data prep → encoding → training → negative sampling for eval → evaluation → top-K → save/load. But the markdown headers are
  numbered 0-7 and section "4. Negative Sampling" comes after "3. Training". The training cell (section 3) already has negatives added during data prep (lines
  336-341 add N_NEG_TRAIN/N_NEG_VALID negatives). Section 4 only handles evaluation negatives. The numbering suggests negatives are added after training, which is
  confusing. Consider renaming section 4 to "4. Evaluation Negative Sampling" or reordering.

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+   "execution_count": 6,
+   "id": "cell-neg-sample",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/var/folders/hw/ycmgd2hd0d32d180_ldvpsgc0000gn/T/ipykernel_98834/278955635.py:89: FutureWarning: DataFrameGroupBy.apply operated on the grouping columns. This behavior is deprecated, and in a future version of pandas the grouping columns will be excluded from the operation. Either pass `include_groups=False` to exclude the groupings or explicitly select the grouping columns after groupby to silence this warning.\n",
+      "  .apply(lambda g: g.head(n_neg))\n",
```
miguelgfierro: .apply(lambda g: g.head(n_neg))
  Triggers a pandas deprecation warning about operating on grouping columns. Fix: add include_groups=False:
  .apply(lambda g: g.head(n_neg), include_groups=False)

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+    "# Add negatives to training and validation so lambdarank has a meaningful signal\n",
+    "train_raw = pd.concat(\n",
+    "    [train_raw, sample_negatives(train_raw, N_NEG_TRAIN, train_raw)], ignore_index=True\n",
+    ")\n",
+    "valid_raw = pd.concat(\n",
+    "    [valid_raw, sample_negatives(valid_raw, N_NEG_VALID, train_raw)], ignore_index=True\n",
+    ")\n",
+    "\n",
+    "\n",
+    "def make_groups(df: pd.DataFrame):\n",
```
miguelgfierro, line 339: make_groups uses sort=False inconsistently 
```
  def make_groups(df):
      df = df.sort_values(USER_COL).reset_index(drop=True)
      group_sizes = df.groupby(USER_COL, sort=False).size().values
```

  After sort_values(USER_COL), the data is already sorted, so sort=False in groupby happens to produce correct order. But it's fragile — if someone changes the
  sort, the groups would be wrong. Use sort=True (the default) to be safe.

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+     "execution_count": 5,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "ranker = LGBMRanker(n_estimators=NUM_BOOST_ROUND, **PARAMS)\n",
+    "\n",
+    "ranker.fit(\n",
+    "    train_x,\n",
```
miguelgfierro: In other notebooks we splitted this cell into two. In one cell we would have the model instantiation: `ranker = LGBMRanker(n_estimators=NUM_BOOST_ROUND, **PARAMS)` and in the cell below the training

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,1981 @@
...
+    "    col_rating=RATING_COL,\n",
+    "    col_prediction=\"prediction\",\n",
+    "    k=10,\n",
+    ")\n",
+    "eval_ndcg5 = ndcg_at_k(\n",
+    "    true_df,\n",
+    "    pred_df,\n",
+    "    col_user=USER_COL,\n",
+    "    col_item=ITEM_COL,\n",
+    "    col_rating=RATING_COL,\n",
```
miguelgfierro: Same idea here, I think we should split this cell into two. In the cell above you can do the prediction and data prep, and in the cell below you can compute the metrics

miguelgfierro (commented):
@ds-wook super good!
I added a few minor comments, but really strong contribution thanks

tests/smoke/examples/test_notebooks_python.py
```diff
@@ -117,6 +117,22 @@ def test_lightgbm_quickstart_smoke(notebooks, output_notebook, kernel_name):
...
 
 
+@pytest.mark.notebooks
+def test_lightgbm_movielens_smoke(notebooks, output_notebook, kernel_name):
+    notebook_path = notebooks["lightgbm_movielens"]
+    execute_notebook(
+        notebook_path,
+        output_notebook,
+        kernel_name=kernel_name,
+        parameters=dict(SIZE="sample"),
```
miguelgfierro: In other to inject parameters in the notebook, you need to add a label in it. See the instructions: https://github.com/recommenders-team/recommenders/tree/main/tests#how-to-create-tests-for-the-notebooks otherwise you are not adding the parameters

examples/00_quick_start/lightgbm_movielens.ipynb
ds-wook: LTR is Learning to Rank.

ds-wook (commented):
@miguelgfierro I reflected the feedback.

1. Notebook parameter injection (lightgbm_movielens.ipynb)                                   
   
  Added parameters tag to the parameters cell                                                  
  Added "tags": ["parameters"] to the cell metadata. Without this tag, papermill cannot locate
  the cell to inject values, so passing MOVIELENS_DATA_SIZE from the test had no effect.       
                             
  SIZE → MOVIELENS_DATA_SIZE                                                                   
  Aligned with the naming convention used across all other notebooks in the project.
                                                                                               
  ---                        
  2. LightGBM parameter cleanup (lightgbm_movielens.ipynb)                                     
                                                                                               
  feature_fraction → colsample_bytree
  feature_fraction is a LightGBM internal alias. When using the scikit-learn API (LGBMRanker), 
  the canonical parameter name is colsample_bytree. The old setting was triggering:            
  feature_fraction is set=0.8, colsample_bytree=1.0 will be ignored.                           
                                                                                               
  ndcg_eval_at moved out of PARAMS → passed as eval_at=[5, 10] in fit()                        
  Keeping ndcg_eval_at inside the params dict caused: Found 'ndcg_eval_at' in params. Will use 
  it instead of 'eval_at'. The correct approach is to pass eval_at directly as a fit()         
  argument.                  
                                                                                               
  Split LGBMRanker(...) initialization into a separate cell                                    
  Separates model definition from model training (.fit()) for readability.
                                                                                               
  ---                        
  3. Negative sampling refactor (lightgbm_movielens.ipynb + pandas_df_utils.py)                
                                                                                               
  Replaced custom sample_negatives with negative_feedback_sampler
  Removed the inline negative sampling function (based on user_item_pairs) and replaced it with
   the existing library utility negative_feedback_sampler. The old implementation was also     
  triggering a pandas FutureWarning on DataFrameGroupBy.apply.
                                                                                               
  Added n_neg_per_user parameter to negative_feedback_sampler (pandas_df_utils.py)             
  The existing function only supported ratio_neg_per_user (relative to positives). The notebook
   needs a fixed count per user (e.g. exactly 100 negatives). n_neg_per_user, when specified,  
  overrides the ratio.       
                                                                                               
  Fixed include_groups=False in pandas_df_utils.py
  groupby().apply() without include_groups=False triggers a pandas FutureWarning. After
  applying this fix, the user column disappears from the group frame, so it is restored via    
  user_df.assign(**{col_user: user_df.name}).
                                                                                               
  ---                        
  4. Evaluation metrics cleanup (lightgbm_movielens.ipynb)
                                                                                               
  Removed NDCG@5, renamed map → map_at_10
  - Before: stored map, ndcg_at_5, ndcg_at_10                                                  
  - After: stores map_at_10, ndcg_at_10      
                                                                                               
  map shadows Python's built-in and omits the cutoff. @10 makes the metric unambiguous. NDCG@5 
  was redundant given NDCG@10 is already tracked.                                              
                                                                                               
  ---                                                                                          
  5. Test updates (smoke / functional test_notebooks_python.py)
                                                                                               
  Fixed parameter name: SIZE="sample" → MOVIELENS_DATA_SIZE="100k" (smoke) / "1m" (functional)
                                                                                               
  Fixed result keys: results["map"], results["ndcg_at_5"], results["ndcg_at_10"] →             
  results["map_at_10"], results["ndcg_at_10"]                                                  
                                                                                               
  Replaced weak assertions with real expected values: >= 0.0 only verifies the notebook runs   
  without error — it says nothing about correctness. Replaced with pytest.approx() using values
   obtained from actually running the notebook:                                                
  - Smoke (100k): map_at_10 ≈ 0.1408, ndcg_at_10 ≈ 0.2442
  - Functional (1m): map_at_10 ≈ 0.1762, ndcg_at_10 ≈ 0.3184                                   
                                                            
  Removed unnecessary @pytest.mark.parametrize: Using parametrize with a single value adds no  value.

examples/00_quick_start/lightgbm_movielens.ipynb
ds-wook: I applied `include_groups=False` to the `.apply()` call in `negative_feedback_sampler` to suppress the pandas deprecation warning about operating on grouping columns.

tests/smoke/examples/test_notebooks_python.py
ds-wook: The notebook already has a parameters cell tag on the parameters cell, which allows papermill to inject values like MOVIELENS_DATA_SIZE at runtime. This has been verified by running the notebook tests.

ds-wook:
@miguelgfierro All requested changes have been implemented. Could you please take a look when you get a chance? Thank you!

recommenders/datasets/movielens.py
```diff
@@ -271,9 +275,25 @@ def load_item_df(
...
+        )
+        item_df = mock_df[[DEFAULT_ITEM_COL]].drop_duplicates().copy()
+        if title_col is not None:
+            item_df[title_col] = mock_df[DEFAULT_TITLE_COL].values[: len(item_df)]
+        if genres_col is not None:
+            item_df[genres_col] = mock_df[DEFAULT_GENRE_COL].values[: len(item_df)]
+        if year_col is not None:
+            item_df[year_col] = 2000
+        item_df = item_df.rename(columns={DEFAULT_ITEM_COL: movie_col})
+        return item_df
```
ds-wook, line 295: Runs the notebook end-to-end with `mock100` synthetic data. `load_item_df` previously only supported real dataset sizes (`100k`, `1m`, etc.), so `mock100` support was added to allow unit tests to run without downloading data.

tests/unit/examples/test_notebooks_python.py
```diff
@@ -76,6 +76,17 @@ def test_lightgbm(notebooks, output_notebook, kernel_name):
...
 
 
+@pytest.mark.notebooks
+def test_lightgbm_movielens(notebooks, output_notebook, kernel_name):
+    notebook_path = notebooks["lightgbm_movielens"]
+    execute_notebook(
+        notebook_path,
+        output_notebook,
+        kernel_name=kernel_name,
+        parameters=dict(MOVIELENS_DATA_SIZE="mock100"),
```
ds-wook, line 86: Set the parameter to `mock100` to prevent errors.

ds-wook (commented):
@miguelgfierro 
The notebook uses both `load_pandas_df` and `load_item_df` to load ratings and item metadata(genre, release year). However, `load_item_df` only supported real dataset sizes (`100k`, `1m`, etc.) and raised a ValueError for `mock100`. To allow unit tests to run without downloading data, `mock100` support was added to `load_item_df` in `recommenders/datasets/movielens.py`.

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,2014 @@
+{
```
ds-wook, line 1: <pre class="ql-syntax" spellcheck="false">Line #6.    ).reset_index(drop=True)
</pre><p>Update evaluation display to use the full eval pool (test interactions + eval negatives) to better reflect model performance across the entire candidate set, and filter displayed examples to users where relevant items appear in top-10.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/recommenders-team/recommenders/pull/2288/discussion/#comment-2985248349'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

ds-wook:
Verified that the unit test passes successfully with mock100.
```sh
$ pytest tests/unit/examples/test_notebooks_python.py::test_lightgbm_movielens -v
====================================================== test session starts =======================================================
platform darwin -- Python 3.11.0, pytest-9.0.2, pluggy-1.6.0 -- /Users/wooklee/.venvs/recommenders/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /Users/wooklee/workspace/recommenders
configfile: pyproject.toml
plugins: anyio-4.12.1, locust-2.43.3, typeguard-4.5.1, hypothesis-6.151.9
collected 1 item

tests/unit/examples/test_notebooks_python.py::test_lightgbm_movielens PASSED                                               [100%]
```

Verified that the smoke test passes successfully with 100k.
```sh
$ pytest tests/smoke/examples/test_notebooks_python.py::test_lightgbm_movielens_smoke -v
====================================================== test session starts =======================================================
platform darwin -- Python 3.11.0, pytest-9.0.2, pluggy-1.6.0 -- /Users/wooklee/.venvs/recommenders/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /Users/wooklee/workspace/recommenders
configfile: pyproject.toml
plugins: anyio-4.12.1, locust-2.43.3, typeguard-4.5.1, hypothesis-6.151.9
collected 1 item

tests/smoke/examples/test_notebooks_python.py::test_lightgbm_movielens_smoke PASSED                                        [100%]
```

Verified that the functional test passes successfully with 1m.
```sh
$ pytest tests/functional/examples/test_notebooks_python.py::test_lightgbm_movielens_functional -v
====================================================== test session starts =======================================================
platform darwin -- Python 3.11.0, pytest-9.0.2, pluggy-1.6.0 -- /Users/wooklee/.venvs/recommenders/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /Users/wooklee/workspace/recommenders
configfile: pyproject.toml
plugins: anyio-4.12.1, locust-2.43.3, typeguard-4.5.1, hypothesis-6.151.9
collected 1 item

tests/functional/examples/test_notebooks_python.py::test_lightgbm_movielens_functional PASSED                              [100%]
```

examples/00_quick_start/lightgbm_movielens.ipynb
ds-wook: I done

examples/00_quick_start/lightgbm_movielens.ipynb
ds-wook: I changed it to "4. Evaluation negative sampling".

examples/00_quick_start/lightgbm_movielens.ipynb
```diff
@@ -0,0 +1,2034 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cell-sec5",
+   "metadata": {},
+   "source": [
+    "## 5. Inference and Evaluation\n",
+    "\n",
+    "Score all test items and compute **NDCG@5** and **NDCG@10** over the combined pool of observed interactions (binary labels) and sampled negatives (label = 0)."
```
miguelgfierro:
```suggestion
    "Score all test items and compute **MAP@10** and **NDCG@10** over the combined pool of observed interactions (binary labels) and sampled negatives (label = 0)."
```

miguelgfierro (approved):
Minor typo

examples/00_quick_start/lightgbm_movielens.ipynb
ds-wook: @miguelgfierro I done thank you!

miguelgfierro:
hey @ds-wook this is great work. Would you like to add yourself to the AUTHORS.md? Feel free to create a new PR and follow the structure there

ds-wook:
@miguelgfierro Thanks! I’ll go ahead and add myself to AUTHORS.md and open a PR.

PR #2304, opened 2026-04-07
refactor: migrate vae pytorch

ds-wook:
### StandardVAE (`standard_vae.py`)
- Migrated `StandardVAE` implementation from TensorFlow/Keras to PyTorch, removing the `LossHistory` and `Metrics` Keras callback classes in favor of a clean, self-contained PyTorch training loop.
- Added GPU support: model and data are automatically moved to CUDA if available, with a fallback to CPU.
- Optimized training performance by preloading the full training and validation datasets to GPU once per `fit()` call, eliminating repeated CPU→GPU transfers per batch.
- Replaced NumPy batch shuffling with `torch.randperm` for GPU-side shuffling.
- Added `ndcg_every` attribute to control NDCG evaluation frequency per epoch, reducing pandas overhead during training.
- Inference (`recommend_k_items`) runs on CPU to free GPU memory during evaluation.
- Updated benchmark results in `standard_vae_deep_dive.ipynb` to reflect results from the PyTorch implementation.

### MultVAE (`multinomial_vae.py`)
- Migrated `MultVAE` implementation from TensorFlow/Keras to PyTorch, removing the `LossHistory`, `Metrics`, and `AnnealingCallback` Keras callback classes in favor of a unified PyTorch training loop.
- Introduced `_MultVAEModel(nn.Module)` as the internal neural network module with encoder, reparameterization, and decoder.
- Preserved the multinomial log-likelihood loss (`F.log_softmax`) and L2 normalization on encoder input, which are the key distinctions of Multi-VAE from Standard VAE.
- Implemented KL annealing directly inside the training loop, replacing the Keras `AnnealingCallback`.
- Added GPU support with automatic CUDA detection and CPU fallback; inference runs on CPU to free GPU memory.
- Exposed `optimal_beta` and `ndcg_per_epoch` as properties; kept `get_optimal_beta()` as a backward-compatible method.
- Added `Mult_VAE = MultVAE` alias for backward compatibility.
- Updated `multi_vae_deep_dive.ipynb` to use the PyTorch implementation, replacing TensorFlow/Keras imports and adjusting API calls accordingly.

### Checklist:


examples/02_model_collaborative_filtering/multi_vae_deep_dive.ipynb
```diff
@@ -1,1869 +1,1914 @@
...
...m/fj3ffvstTk5OpqQZoF69eixbtgyAEydOYGlpScOGDQHo168fYWFh9504K5VKKlWqVKZ9RszYjpudnhmvP3Nf5y4vBw8eJDg4uLzDEEII8R8zbdq0/2fvzsOjrO/9/z9ny8wkIWSBJBMWEYQQDGAFtT0VrEUFazAcPX6lyOki4lE8Sjku5NCfBKrVE6wctYKt1OXg14UiFWpEcaH+VPAgRFwgoghhTUggC2Sbfb5/BMamEJgJIXeGvB7X5dXM/bnnvl+TwjW878/GlClTyM/PZ9WqVcydO5elS5ced14gEKCwsJArrrii1fFnnnmGxMRE3nnnHXbt2sVNN93E22+/TUJCwknbYlmguZn9r60i4dxzAbD37m1wIhEROdMMGae8du1aMjIyOP/881sdDwaDvPzyy/z4xz8GoKKigqysrHB7amoqwWCQurq6Ts0LYDZD8Gya5CwiIt1edXU1paWl5OXlAZCXl0dpaSk1NTXHnfv000/zox/96Lje6DfffJMbb7wRgAEDBpCbm8sHH3xwyrZYdvCDD9m7bDlf/24hAI6MdIMTiYjImWZIj/OKFSu4/vrrjzv+wAMPEB8fz9SpUzv0flu2bDnta3g8HkIOGyUlJR2QyBjKbgxlN04s51d2Y8Ry9vaoqKggIyMDi8UCtCyemZ6eTkVFRXitEYBt27bx0UcfsXTpUhYvXtzqGuXl5fTp0yf82uVyceDAgVO2RaojvsOP6aj/f32bWq4T8vsx52SzZc8e2LOnQ67dllj+s6nsxonl/MpuDGVvW6cXzpWVlWzcuJEFCxa0Ol5UVMTu3bv5wx/+EF6wy+VyUV5eHj6npqYGs9lMcnJyVPfMzc3FbrefVu6EtWsJhnyMGjXqtK5jlJKSEmU3gLIbJ5bzK7sxOiK7x+Pp0EKvK/D5fNx///08/PDD4QK7M3XEdzh07J/NL5a9SqPdjrNvH4bOmokjI6NDrtuW7v73yiixnB1iO7+yG6O7Zz/Vd3inF86vvfYal112GSkpKeFjCxcuZMuWLTz99NOtFrDKzc3F7XazadMmRo8ezSuvvMKECRM6OzIAZrMJbX0sIiJnE5fLRWVlJYFAAIvFQiAQoKqqCpfLFT7n4MGD7Nmzh1tvvRWAI0eOEAqFaGho4IEHHiArK4v9+/eHe6grKirCa5ecrC1WBX0+GnaW4brmas795c+Nji
```
miguelgfierro, line 2: There is a difference between the tf results and pt ones:
TF:
```
Took 4.4747 seconds for prediction.
MAP@10:		0.048516
NDCG@10:	0.430152
Precision@10:	0.403833
Recall@10: 	0.075556
```

PT:
```
Took 0.3243 seconds for prediction.
MAP@10:		0.329040
NDCG@10:	0.459831
Precision@10:	0.434167
Recall@10: 	0.086624
```

Here some ideas: 

  1. Batch generator bug in the TF version (most likely primary cause)

  The old TF nn_batch_generator (standard_vae.py:340-367) has a double-indexing bug:

  shuffle_index = np.arange(np.shape(x_train)[0])
  np.random.shuffle(shuffle_index)
  x = x_train[shuffle_index, :]          # x is ALREADY shuffled

  # ...
  index_batch = shuffle_index[...]        # indices from the SHUFFLE array
  x_batch = x[index_batch, :]            # indexing shuffled data WITH shuffle indices

  It shuffles the data into x, then uses shuffle_index values (which are random row indices like [5, 2, 8,
  ...]) to index into the already-shuffled array. So x_batch = x[5] gets x_train[shuffle_index[5]] — a
  double-remapping. This means the batches are not contiguous windows of a shuffled dataset. Some training
  samples may be over-represented (appear in multiple batches) while others are skipped entirely within an
  epoch.

  On top of that, the generator never re-shuffles — the counter resets to 0 but the same shuffle_index and x
  are reused for every epoch. The model sees the exact same (broken) batch assignments for all 400 epochs.

  The PyTorch version uses torch.randperm per epoch — correct, fresh shuffling every time. This alone can
  account for a large performance difference.

  2. ndcg_every=5 changes model selection (StandardVAE only)

  The new StandardVAE only evaluates NDCG every 5 epochs:

  if (epoch + 1) % self.ndcg_every == 0:
      ndcg = ndcg_at_k(...)
  else:
      ndcg = self.val_ndcg[-1] if self.val_ndcg else 0.0

  The best-model checkpoint is saved based on NDCG improvement, but it only "sees" 20% of epochs. This can
  result in a different model being selected as "best" compared to the TF version which checks every epoch.
  Whether this helps or hurts depends on the loss surface, but it changes the outcome.

  3. Weight initialization differences (StandardVAE)

  _MultVAEModel correctly matches the TF initializers:
  nn.init.xavier_uniform_(self.encoder_h.weight)    # matches glorot_uniform
  nn.init.trunc_normal_(self.encoder_h.bias, std=0.001)  # matches truncated_normal

  But _VAEModel (StandardVAE) uses PyTorch defaults — kaiming_uniform for weights, uniform for biases. The
  old TF StandardVAE used Keras defaults (glorot_uniform for weights, zeros for biases). Different
  initialization → different loss landscape trajectory → different final model.

  4. Adam optimizer implementation differences

  TF legacy Adam and PyTorch Adam differ in epsilon defaults (TF: 1e-7, PT: 1e-8) and in the exact update
  formula (TF uses the biased second moment in some code paths). Over 400 epochs these differences compound.

  Why MAP is disproportionately affected

  MAP (Mean Average Precision) is extremely sensitive to the order of relevant items at the very top of the
  ranking. A model that ranks the first relevant item at position 1 vs position 3 can see massive MAP swings
  even if NDCG (which uses logarithmic discounting) barely changes. The TF model's broken batch generator
  likely hurt the model's ability to learn fine-grained ranking at the top of the list — exactly the regime
  MAP measures.

  Recommendation

  If the goal is a faithful port (same results), the contributor should fix the batch generator bug
  comparison, match initialization in _VAEModel, and set ndcg_every=1 by default. If the goal is simply a
  better PyTorch implementation, these improvements are fine to keep — but the initialization inconsistency
  between _VAEModel and _MultVAEModel should still be addressed.

recommenders/models/vae/multinomial_vae.py
```diff
@@ -1,196 +1,88 @@
...
-        """Constructor
-
-        Args:
-            beta (float): current value of beta.
-            anneal_cap (float): maximum value that beta can reach.
-            total_anneal_steps (int): total number of annealing steps.
-        """
-        # maximum value that beta can take
-        self.anneal_cap = anneal_cap
+from __future__ import annotations
```
miguelgfierro: I don't think we need this: The codebase targets Python 3.9+ (based on pyproject.toml and existing code), and the PR only uses  Optional, List, Tuple from typing — all of which work without from __future__ import annotations. The one  place it uses lowercase list[float] in standard_vae.py is inconsistent with the List[float] used elsewhere
  in the same file, but the fix is to pick one style, not to add the future import.

  The rest of the repo doesn't use from __future__ import annotations anywhere. It should be removed to stay  consistent.

recommenders/models/vae/standard_vae.py
```diff
@@ -1,223 +1,94 @@
...
-
-        K.set_value(self.beta, new_beta)
-
-    def on_epoch_end(self, epoch, logs=None):
-        """At the end of each epoch save the value of beta in _beta list."""
-        if logs is None:
-            logs = {}
-        tmp = K.eval(self.beta)
-        self._beta.append(tmp)
+    def __init__(
```
miguelgfierro, line 21: The old TF code used Keras defaults (glorot_uniform / Xavier uniform) for weight initialization. PyTorch defaults to  kaiming_uniform_, which is a
   different distribution. _MultVAEModel correctly adds explicit nn.init.xavier_uniform_ to match, but this class does not. Please add the same xavier_uniform_ + trunc_normal_ bias init here for consistency and to match the original TF behavior. Same for decoder_h.

recommenders/models/vae/standard_vae.py
```diff
@@ -1,223 +1,94 @@
 # Copyright (c) Recommenders contributors.
 # Licensed under the MIT License.
 
-import numpy as np
+from __future__ import annotations
```
miguelgfierro: StandardVAE._vae_loss uses F.binary_cross_entropy with mean reduction,   then multiplies by original_dim. The original TF code computes   binary_crossentropy(x, x_bar) per-sample (which averages over the
  original_dim features), then multiplies by original_dim — effectively summing
   over features. The PyTorch version uses reduction="mean" which averages over
   both batch AND features, then multiplies by original_dim. This means the
  reconstruction loss is averaged over the batch dimension in the PyTorch   version but summed in the TF version. The KL term uses torch.mean over batch.
   The scaling mismatch may cause different training dynamics. Should use   reduction="none", sum over features, then mean over batch — or verify   numerically that behavior matches.

recommenders/models/vae/standard_vae.py
```diff
@@ -1,223 +1,94 @@
...
-import tensorflow as tf
-from tensorflow.keras.layers import *
-from tensorflow.keras.models import Model
-from tensorflow.keras.losses import binary_crossentropy
-from tensorflow.keras import backend as K
-from tensorflow.keras.callbacks import ReduceLROnPlateau, Callback
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.optim import Adam
```
miguelgfierro, line 12: StandardVAE.ndcg_every = 5 changes evaluation semantics. The old code   computed NDCG every epoch. The new code defaults to every 5 epochs and fills
  gaps with the previous value (self.val_ndcg[-1] if self.val_ndcg else 0.0).
  This means: (a) the best model might be saved based on a stale NDCG from 4   epochs ago, (b) display_metrics() plots a staircase rather than a smooth
  curve, (c) MultVAE computes NDCG every epoch but ztandardVAE does not —
  inconsistent behavior between the two classes. The ndcg_every parameter is   not documented in __init__ and not exposed as a constructor arg.

recommenders/models/vae/standard_vae.py
```diff
@@ -376,77 +208,87 @@ def fit(self, x_train, x_valid, x_val_tr, x_val_te, mapper):
...
-                validation_data=(x_valid, x_valid),
-            )
-
-        # save lists
-        self.train_loss = history.losses
-        self.val_loss = history.val_losses
-        self.val_ndcg = metrics.get_data()
-
-    def get_optimal_beta(self):
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
miguelgfierro: self.device is set twice — once in __init__ and again at the top of fit().
   The second assignment is redundant (same value).

miguelgfierro (commented):
Added somesuggestions, the main feedback is to try to get to the same metric values, if we don't get to the same values in the metrics, it should be better with an explation,or ideally the sam

ds-wook:
@miguelgfierro 
Thanks for the detailed review! Here's an update on each point:

Initialization (_VAEModel vs _MultVAEModel) — Fixed. Both classes now consistently use xavier_uniform_ for weights and trunc_normal_(std=0.001) for biases, matching the TF paper initializers.

ndcg_every — Fixed. NDCG@k is now computed every epoch by default, which is equivalent to ndcg_every=1.

Batch generator — I believe the current PyTorch implementation is actually correct here. We use torch.randperm to shuffle indices and slice with [i * batch_size : (i + 1) * batch_size], which gives clean, non-overlapping batches without any off-by-one issue. The original TF version had a subtle comparison bug (< vs <=) that this approach avoids entirely by construction.

Scoring — I also think the current scoring logic is sound. In _recommend_k_items_internal, we zero out seen items from the score matrix and use argpartition to extract top-k, then reconstruct a sparse matrix by subtracting the zeroed copy from the original. This correctly preserves only the top-k scores while setting the rest to zero. Happy to discuss further if you see a case I might be missing!

ds-wook:
@miguelgfierro I found out what the problem is.

### Root cause identified: `map` vs `map_at_k` mismatch

The inflated MAP@10 value (0.3143 instead of the expected ~0.0518) was **not a model issue** — the PyTorch port is working correctly. The discrepancy comes from a change in the `recommenders` evaluation library.

NDCG/Precision/Recall all match the original TF results within ~1%:

| Metric       | TF original | Current PyTorch       |
| ------------ | ----------- | --------------------- |
| MAP@10       | 0.0518      | **0.3143** ← 6× inflated |
| NDCG@10      | 0.4462      | 0.4430                |
| Precision@10 | 0.4217      | 0.4187                |
| Recall@10    | 0.0813      | 0.0803                |

If the model were broken, all four metrics would be off — not just MAP.

#### The actual difference

The difference lives in `recommenders/evaluation/python_evaluation.py`:

```python
# map() — older definition (what the TF notebook actually used)
return (df_merge["rr"] / df_merge["actual"]).sum() / n_users
                        ^^^^^^^^^^^^^^^^^^  divides by total number of relevant items

# map_at_k() — Spark MLlib-style (what the current notebook was calling)
return (df_merge["rr"] / df_merge["actual"].apply(lambda x: min(x, k))).sum() / n_users
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  divides by min(relevant, k)
```
On ML-100k, the average user has far more than 10 relevant items, so min(actual, 10) = 10. The denominator becomes much smaller than actual (which can be in the tens or hundreds), inflating MAP roughly 6×.

### Fix
Switched both notebooks (`standard_vae_deep_dive.ipynb`, `multi_vae_deep_dive.ipynb`) `from map_at_k to map (imported as map_metric)` to reproduce the original TF benchmark values.

```python
from recommenders.evaluation.python_evaluation import map as map_metric
eval_map_1 = map_metric(test_df, top_k_df, col_prediction="prediction", k=10)
```
### Note
Neither definition is "wrong" — they are different conventions:

`map (legacy)`: reproduces the historical TF notebook numbers; recall-biased.
`map_at_k (current default)`: Spark MLlib / modern IR textbook definition.

For benchmark parity with the TF version, map is the correct choice.

recommenders/models/vae/standard_vae.py
```diff
@@ -1,223 +1,103 @@
...
+        # Match TF paper initializers: glorot_uniform weights, truncated_normal(std=0.001) biases
+        for layer in [
+            self.encoder_h,
+            self.z_mean_layer,
+            self.z_log_var_layer,
+            self.decoder_h,
+            self.decoder_out,
+        ]:
+            nn.init.xavier_uniform_(layer.weight)
+            nn.init.trunc_normal_(layer.bias, std=0.001)
```
ds-wook, line 52: > The old TF code used Keras defaults (glorot_uniform / Xavier uniform) for weight initialization. PyTorch defaults to kaiming_uniform_, which is a
different distribution. MultVAEModel correctly adds explicit nn.init.xavier_uniform to match, but this class does not. Please add the same xavier_uniform_ + trunc_normal_ bias init here for consistency and to match the original TF behavior. Same for decoder_h.

I've reflected it.

recommenders/models/vae/standard_vae.py
```diff
@@ -376,77 +217,83 @@ def fit(self, x_train, x_valid, x_val_tr, x_val_te, mapper):
...
+            self.train_loss.append(avg_train_loss)
+            self.val_loss.append(val_loss)
+            self.ls_beta.append(self.beta)
+
+            # NDCG@k on validation
+            top_k = self._recommend_k_items_internal(x_val_tr, self.k, remove_seen=True)
+            top_k_df = mapper.map_back_sparse(top_k, kind="prediction")
+            test_df = mapper.map_back_sparse(x_val_te, kind="ratings")
+            ndcg = ndcg_at_k(test_df, top_k_df, col_prediction="prediction", k=self.k)
+            self.val_ndcg.append(ndcg)
```
ds-wook, line 272: > This means: (a) the best model might be saved based on a stale NDCG from 4 epochs ago, (b) display_metrics() plots a staircase rather than a smooth
curve, (c) MultVAE computes NDCG every epoch but ztandardVAE does not —
inconsistent behavior between the two classes. The ndcg_every parameter is not documented in init and not exposed as a constructor arg.

I've reflacted it.

miguelgfierro:
@anargyri can you review the definition of MAP? I think you discussed something related to the metrics in the past?

ds-wook:
Hi @anargyri, just checking in on this — I’m planning next steps based on the MAP definition, so your input would be really helpful when you get a chance!

ds-wook:
@miguelgfierro Since I haven’t heard back yet, would it be okay if I move this to an issue and proceed with migrating to another model for now? Happy to revisit the MAP definition once we have alignment. Let me know what you think!

miguelgfierro:
@ds-wook yes. From this list: https://github.com/recommenders-team/recommenders/issues/2302 what's your preference?

ds-wook:
@miguelgfierro I want implement LightGCN

miguelgfierro:
> @miguelgfierro I want implement LightGCN

perfect

anargyri:
Hi folks, regarding the MAP definition, there has been a lot of discussion in the past and eventually we converged to Spark's implementation of MAP. See issue #1702 and PR #2004. That is, we use `min(k, actual)` in the denominator. This applies to `map_at_k()` function but `precision_at_k()` normalises in a similar way too. 

So I suggest you keep the `map_at_k()` call for reporting MAP@k metrics. The result will change but this is fine since the TF implementation was not consistent with the rest of the repo in this respect.

ds-wook:
@miguelgfierro @anargyri Sounds good! Let’s switch to `map_at_k()` and evaluate performance using that definition going forward.

anargyri:
Btw I noticed that in the notebooks we have not updated the MAP metric accordingly and we still report the value returned by `map()`. I opened a new issue on this #2309 and will make the change; we can discuss more on Monday @miguelgfierro @SimonYansenZhao .

miguelgfierro:
@ds-wook hey we met on Monday, and we are going to use map_at_k, which is the same implementation as Spark. 

So could you please replace the line:
```
from recommenders.evaluation.python_evaluation import map as map_metric
```
to be:
```
from recommenders.evaluation.python_evaluation import map_at_k
```
and rerun the notebook fully. I think that would be completed

ds-wook:
@miguelgfierro I done. Thank you!

miguelgfierro (approved):
this is great @ds-wook, congrats!

Issue #2140, opened 2024-08-10
[ASK] Confusion between top_k and by_threshold relevancy_method in python_evaluation.py


mnhqut:
### Description
What is the different between k and threshold value here if they all get assigned to top_k?
Isn't threshold supposed to be a rating value that the predictions should exceed instead of being the number of items in the top_k list ?
Thanks.
```
def merge_ranking_true_pred(
    rating_true,
    rating_pred,
    col_user,
    col_item,
    col_prediction,
    relevancy_method,
    k=DEFAULT_K,
    threshold=DEFAULT_THRESHOLD,
    **_,
):
    if relevancy_method == "top_k":
        top_k = k
    elif relevancy_method == "by_threshold":
        top_k = threshold
```

daviddavo:
I don't really understand it neither. It calls `get_top_k_items`, which does this:

https://github.com/recommenders-team/recommenders/blob/4f86e4785337d455aa3cb7e8920c3fab9a2a0140/recommenders/evaluation/python_evaluation.py#L859-L871

With `k` being `top_k` (`threshold`). It then calls `.head(k)`, but according to [pandas docs](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.head.html), `k` should be an integer.

The `DEFAULT_THRESHOLD` variable is just `10`, an integer. Have you tried using a float? What happens then? Does it work? Isn't this exactly the same as using `k`?

I think a user expects `by_threshold` to just return just items that have a rating above that threshold.

mnhqut:
Yes it is indeed very confusing how 'by_threshold' works. 

Doing a quick test, I passed "relevancy_method = "top_k", k =20 ", and got the exact same result as passing "relevancy_method = "by_threshold", threshold = 20, k = whatever ". So the threshold here does not seem to represent a (float) rating value.

As for the type of 'top_k', if I passed 'threshold' or 'k' as a float, it raised no error and worked as if the number got rounded up, for example 2.1 as 3 or 7.6 as 8.

PR #2310, opened 2026-04-25
Fix by_threshold relevancy method so ranking metrics stay in [0, 1]

### LINKED ISSUE #2154: [BUG] Ranking Evaluation Metrics Exceed 1 with "by_threshold" Relevancy Method
### Description
Hello!

I encountered an issue while evaluating the BPR (Bayesian Personalized Ranking) model with basically the same code provided in the example on a different dataset. Specifically, when using the "by_threshold" relevancy method with ranking metrics, the computed values for precision@k, ndcg@k, and map@k exceed 1, which seems incorrect. This issue does not occur when switching the relevancy method to "top_k."

I use the following parameter for BPR (all using the default seed):
```
bpr = cornac.models.BPR(
    k=200,
    max_iter=100,
    learning_rate=0.01,
    lambda_reg=0.001,
    verbose=True 
)
```
Using these evaluation
```
TOP_K = 10
threshold =50
eval_map = map_at_k(test, all_predictions, col_prediction="prediction",
                    relevancy_method='by_threshold', threshold=threshold, k=TOP_K)
eval_ndcg = ndcg_at_k(test, all_predictions, col_prediction="prediction",
                      relevancy_method='by_threshold', threshold=threshold, k=TOP_K)
eval_precision = precision_at_k(
    test, all_predictions, col_prediction="prediction",
    relevancy_method='by_threshold', threshold=threshold, k=TOP_K)
```
Here is the dataset I test on:  https://github.com/mnhqut/rec_sys-dataset/blob/main/data.csv

My result: 
MAP:		1.417529
NDCG:		1.359902
Precision@K:	2.256466
### Willingness to contribute

- [ ] Yes, I can contribute for this issue independently.
- [x ] Yes, I can contribute for this issue with guidance from Recommenders community.

Yinchaochen:
## Summary

`relevancy_method="by_threshold"` is documented as a way to determine relevance by a prediction-score threshold, but the implementation in `merge_ranking_true_pred` silently passed the `threshold` value to `get_top_k_items` as `k`. That made it a top-N count rather than a score cutoff, and let metrics like `precision_at_k`, `ndcg_at_k`, `map`, and `map_at_k` exceed `1`, which is impossible by definition.

This PR fixes the implementation to match the documented semantics and adds a regression test.

## What changed

- `recommenders/evaluation/python_evaluation.py`
  - `merge_ranking_true_pred`: when `relevancy_method == "by_threshold"`, filter `rating_pred_common` to rows where `col_prediction >= threshold` first, then apply the standard top-`k` cutoff. Hits are now bounded by `k`, so every metric stays in `[0, 1]`.
  - Updated the `threshold` docstring on every public metric (precision/recall/r-precision/ndcg/map/map_at_k/merge_ranking_true_pred) so users can tell it is a score cutoff rather than a count.

- `tests/unit/recommenders/evaluation/test_python_evaluation.py`
  - New `test_python_ranking_metrics_by_threshold` covers three cases:
    1. Threshold above the highest predicted score → every ranking metric is `0`.
    2. Threshold below the lowest predicted score → `by_threshold` collapses to `top_k` and the two methods agree.
    3. Mid threshold → metrics stay inside `[0, 1]` (the original bug let them exceed 1).

## Linked issues

- Fixes #2154 (precision/NDCG/MAP@k > 1 with `by_threshold`)
- Refs #2140 (`by_threshold` previously behaved indistinguishably from `top_k`)

## Behavior change

This is a behavior change, not just a docstring fix: `relevancy_method="by_threshold"` no longer reinterprets `threshold` as a count.

- If a caller was relying on the old behavior (using `threshold` as a count of top items), they should switch to `relevancy_method="top_k"` and pass that value as `k`.
- If a caller was passing a real score-based threshold (e.g. `threshold=0.5`), they now get the documented behavior — items below the score threshold are excluded.

## Verification

- `pytest tests/unit/recommenders/evaluation/test_python_evaluation.py -k "ranking or precision or recall or ndcg or map or merge_ranking or threshold or get_top_k"` → 10 passed
- `pytest tests/unit/recommenders/evaluation/test_python_evaluation.py::test_python_ranking_metrics_by_threshold` → 1 passed

I noticed `test_distributional_coverage` already fails on the unmodified `staging` branch — that is unrelated to this change.


tests/unit/recommenders/evaluation/test_python_evaluation.py
```diff
@@ -387,6 +387,56 @@ def test_python_r_precision(rating_true, rating_pred, rating_nohit):
     )
 
 
+def test_python_ranking_metrics_by_threshold(rating_true, rating_pred):
```
miguelgfierro, line 390: hey @Yinchaochen thanks for finding this bug, the previous implementation was weird. 

For these tests, instead of explaining what we had before and what we have now, can we just do a number of tests to make sure everything is correct?

Yinchaochen: Thanks for the feedback! I've rewritten the test to remove the regression narrative and replace the loose `[0, 1]` bounds check with concrete expected values. The new test has three cases:

 1. `threshold=100` (above all predicted scores) → all metrics are `0.0`
 2. `threshold=0` (below all predicted scores) → matches `top_k` results
 3. `threshold=13` → selects exactly the top-2 predictions per user, all true hits, so `precision == 1.0` and `recall == 19/45` (hand-verified from the fixture data)

miguelgfierro (approved):
Thanks for the contribution @Yinchaochen

PR #2313, opened 2026-04-28
Add support for allocating on-demand VMs for testing workflows

SimonYansenZhao:
### Description

This PR adds scripts that use the Compshare cloud service APIs to allocate VMs for the testing workflows.

Currently the testing workflows are using pre-allocated self-hosted VMs waiting for testing jobs.  And there are only 3 runners working in parallel, and they are changed monthly with 100USD.

This PR modifies the testing workflows to allocate the VMs on demand, so that more runners can be used for testing to speed up the testing.  It is also an experiment to see whether creating VMs on demand is more cost-effective and faster for testing than the current pre-allocated strategy.  The minimum hourly price to allocate a GPU VM is 0.38CNY (0.06USD).  More details about the price can be found at [spec.json](https://github.com/recommenders-team/recommenders/blob/simonz/ucloud-api/tools/ci/compshare/spec.json).

### Checklist:

- [X] I have followed the [contribution guidelines](https://github.com/recommenders-team/recommenders/blob/main/CONTRIBUTING.md) and code style for this project.
- [X] I have [signed the commits](https://github.com/recommenders-team/recommenders/wiki/How-to-sign-commits), e.g. `git commit -s -m "your commit message"`.
- [X] This PR is being made to `staging branch` AND NOT TO `main branch`.

### Todos
- [ ] Set up Docker and PyPI cache servers for speedup.


.github/workflows/compshare-vm.yml
```diff
@@ -0,0 +1,203 @@
+# ---------------------------------------------------------
```
miguelgfierro, line 1: @SimonYansenZhao not sure if you finally were considering Terraform. Here's something I found about them.

Terraform changed its license in 2023 from open-source (MPL 2.0) to Business Source License. Internal use is still free for now, but HashiCorp has indicated the OSS version is being wound down in favor of the paid Enterprise tier. Long-term, relying on Terraform means either accepting eventual licensing costs or living with roadmap uncertainty.

OpenTofu is the open-source fork — Linux Foundation-governed, MPL 2.0, drop-in compatible with Terraform's HCL syntax and providers. No licensing risk.

For what we need — spinning up a CPU/GPU VM in any cloud, running tests, collecting logs, and tearing it down — OpenTofu fits perfectly. The provider ecosystem covers AWS, GCP, and Azure equally well, so we can keep it cloud-agnostic from day one.

Just mentioning as a possible alternative to consider, not saying we should take this road, but just something to take into account.

miguelgfierro (approved):
looks great!

.github/workflows/compshare-vm.yml
SimonYansenZhao: Totally agree.  And OpenTofu is a fork of Terraform.  Once we have enough GPU quota on Azure or any other cloud services.  I can try OpenTofu.  But now the cloud service we are using does not support Terraform or OpenTofu.  So this PR is just a temporary solution for the testing workflow.

SimonYansenZhao: Actually the cloud service we are using does support Terraform, but only for their advanced services.  The cheaper one we are using is not.  At the beginning, I thought they supported, but I eventually realized that the cheaper one is different from their advanced.  That's why I have to use their APIs directly.

PR #2314, opened 2026-04-29
[codex] fix(gpu): use imported cuda namespace for gpu counting

### LINKED ISSUE #2198: [BUG] Direct use of `numba.cuda.gpus` lead to error that module 'numba' has no attribute 'cuda
### Description
After installing `recommenders[gpu]`, doing `len(numba.cuda.gpus)` will lead to the following error

```
>>> len(numba.cuda.gpus)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'numba' has no attribute 'cuda'
```

Changing to `from numba import cuda` and then `cuda.gpus` will resolve the issue.

### In which platform does it happen?
Linux Ubuntu 22.04

See above.

See above.

### Willingness to contribute

- [ ] Yes, I can contribute for this issue independently.
- [ ] Yes, I can contribute for this issue with guidance from Recommenders community.

Yinchaochen:
## Summary
- fix `get_number_gpus()` to use the already imported `cuda` namespace instead of `numba.cuda`
- add a regression test that simulates `torch` being unavailable and a `numba` module without a `cuda` attribute
- register the new regression test in `tests/test_groups.yml` so the PR gate runs it

## Root Cause
`recommenders.utils.gpu_utils` imports `cuda` with `from numba import cuda`, but `get_number_gpus()` fell back to `numba.cuda.gpus` when `torch` was unavailable. In environments where `numba.cuda` is not exposed through the top-level `numba` module, that path raises `AttributeError: module 'numba' has no attribute 'cuda'` and incorrectly reports zero GPUs.

## Impact
When PyTorch is unavailable, GPU counting now follows the same import path already used by `get_gpu_info()` and `clear_memory_all_gpus()`. That keeps `get_number_gpus()` aligned with the rest of the module and avoids the `numba.cuda` attribute error reported in #2198.

## Validation
- `pytest tests/unit/recommenders/utils/test_gpu_utils.py -k "falls_back_to_cuda_namespace_when_torch_is_missing"`

Fixes #2198

### DIFF
--- a/tests/unit/recommenders/utils/test_gpu_utils.py
+++ b/tests/unit/recommenders/utils/test_gpu_utils.py
@@ -21,9 +21,7 @@ except ImportError:
     pass  # skip this import if we are in cpu environment
 
 
-def test_get_number_gpus_falls_back_to_cuda_namespace_when_torch_is_missing(
-    monkeypatch,
-):
+def test_get_number_gpus_without_torch(monkeypatch):
     fake_cuda = SimpleNamespace(gpus=["gpu0", "gpu1"])
     real_import = builtins.__import__
 
--- a/tests/unit/recommenders/utils/test_gpu_utils.py
+++ b/tests/unit/recommenders/utils/test_gpu_utils.py
@@ -2,8 +2,11 @@
 # Licensed under the MIT License.
 
 
+import builtins
 import sys
+from types import SimpleNamespace
 import pytest
+import recommenders.utils.gpu_utils as gpu_utils
 
 try:
     import tensorflow as tf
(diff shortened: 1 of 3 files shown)


tests/unit/recommenders/utils/test_gpu_utils.py
```diff
@@ -18,6 +21,25 @@
     pass  # skip this import if we are in cpu environment
 
 
+def test_get_number_gpus_falls_back_to_cuda_namespace_when_torch_is_missing(
```
anargyri: Maybe this name is a bit too long and could be shortened.

anargyri (approved):

tests/unit/recommenders/utils/test_gpu_utils.py
Yinchaochen: ok, working on it

miguelgfierro:
@Yinchaochen is the pr finished or are you still working on it (draft)?

Yinchaochen:
> @Yinchaochen is the pr finished or are you still working on it (draft)?

It's now finished.

PR #2315, opened 2026-04-30
refactor: migrate lightgcn pytorch

ds-wook:
### Description

Translate the LightGCN model from TensorFlow v1 (sessions, placeholders, tf.compat.v1.disable_eager_execution) to PyTorch (nn.Module). The public API of the LightGCN class is preserved (fit, load, run_eval, score, recommend_k_items, infer_embedding), so downstream notebooks, benchmarks, and tests work as drop-in replacements after switching the imports. The ImplicitCF data module is unchanged because it has no TF dependency.

**File:** `recommenders/models/deeprec/models/graphrec/lightgcn.py`

- Use `nn.Embedding` for user/item parameters, initialized with `xavier_uniform_`  
  (equivalent to TF `VarianceScaling(scale=1.0, mode="fan_avg", distribution="uniform")`).

- Convert the normalized adjacency matrix to a `torch.sparse_coo_tensor` once and move it to the model device.

- `_propagate()` performs K-layer light graph convolution with `torch.sparse.mm`, then returns averaged user/item embeddings.

- Implement BPR loss as:  
  `softplus(-(pos − neg))` + L2 regularizer `0.5 · Σx²`  
  (algebraically equivalent to `tf.nn.l2_loss`).

- Checkpointing uses `torch.save` / `torch.load` of the `state_dict` to a `model.pt` file.  
  `load()` accepts either the file path or its parent directory.

- Add module-level type hints with `from __future__ import annotations` (Python 3.9+ compatible) and `TYPE_CHECKING` guards for `ImplicitCF` / `scipy.sparse`.

### Related Issues

https://github.com/recommenders-team/recommenders/issues/2302

### Checklist:


recommenders/models/deeprec/models/graphrec/lightgcn.py
```diff
@@ -15,10 +22,16 @@
 )
 from recommenders.utils.python_utils import get_top_k_scored_items
 
-tf.compat.v1.disable_eager_execution()  # need to disable eager in TF2.x
+if TYPE_CHECKING:
```
miguelgfierro: Is there value in `if TYPE_CHECKING`? or can we directly move 
```
import scipy.sparse as sp
from recommenders.models.deeprec.DataModel.ImplicitCF import ImplicitCF
```
to the top with the rest of the imports?

recommenders/models/deeprec/models/graphrec/lightgcn.py
```diff
@@ -69,177 +93,187 @@ def __init__(self, hparams, data, seed=None):
...
-            self.pos_i_g_embeddings,
-            transpose_a=False,
-            transpose_b=True,
-        )
+        # Trainable embeddings (matches TF VarianceScaling fan_avg uniform == xavier_uniform)
+        self.user_embedding = nn.Embedding(self.n_users, self.emb_dim)
+        self.item_embedding = nn.Embedding(self.n_items, self.emb_dim)
+        nn.init.xavier_uniform_(self.user_embedding.weight)
+        nn.init.xavier_uniform_(self.item_embedding.weight)
+        print("Using xavier initialization.")
```
miguelgfierro: here it would be better to use `logging` for consistency with the rest of the repo.

recommenders/models/deeprec/models/graphrec/lightgcn.py
```diff
@@ -28,8 +41,13 @@ class LightGCN(object):
         preprint arXiv:2002.02126, 2020.
     """
 
-    def __init__(self, hparams, data, seed=None):
-        """Initializing the model. Create parameters, placeholders, embeddings and loss function.
+    def __init__(
```
miguelgfierro, line 45: Great work on the TF→PyTorch migration — the translation is correct and the PyTorch idioms (sparse COO tensor, accumulating losses on device, xavier_uniform_) are all solid choices.

  One API design issue worth addressing before merging: the hparams: Any pattern is inherited from the  TF codebase, but now that we're on PyTorch we have a chance to make the interface explicit. Currently a user can't tell what parameters the model accepts without reading the constructor body or the
  HParams docs. We also create the optimizer implicitly inside fit(), which means training  hyperparameters are invisible at the call site.

  Proposed split:

  Constructor — only what's needed to build the model (architecture + graph structure):
```
  def __init__(
      self,
      n_users: int,
      n_items: int,
      norm_adj: "sp.spmatrix",
      embed_size: int = 64,
      n_layers: int = 3,
      seed: int | None = None,
  ) -> None:
```

  Note: data moves out of the constructor. Only the three values needed to build the embedding tables and graph buffer belong here.

  fit() — the control panel for all training behavior:

```
  def fit(
      self,
      data: "ImplicitCF",
      epochs: int = 50,
      learning_rate: float = 1e-3,
      batch_size: int = 1024,
      decay: float = 1e-5,
      eval_epoch: int = 5,
      top_k: int = 10,
      metrics: list[str] | None = None,
      save_model: bool = True,
      save_epoch: int = 5,
      model_dir: str = "./",
  ) -> None:
```

  fit() passes top_k and metrics down to run_eval() internally. This way the user can retrain with different hyperparameters without reinstantiating the model.

  run_eval() — takes its own explicit params:

```
  def run_eval(self, top_k: int = 10, metrics: list[str] | None = None) -> list[float]:
```

  Typical usage pattern:

```
  data = ImplicitCF(...)
  model = LightGCN(data.n_users, data.n_items, data.get_norm_adj_mat(), embed_size=64)
  model.fit(data, epochs=50, learning_rate=1e-3)
```

  # retrain with different lr — no reinstantiation needed
```
  model.fit(data, epochs=20, learning_rate=5e-4)
```

  This makes the contract of each method self-documenting and removes the dependency on an external config object entirely.

What do you think @ds-wook?

ds-wook: @miguelgfierro I agree with your suggestion and will make the changes.

recommenders/models/deeprec/models/graphrec/lightgcn.py
ds-wook: I used that pattern for a bit of minimal safety, but it shouldn’t cause any real issues to move them. I’ll update it accordingly.

recommenders/models/deeprec/models/graphrec/lightgcn.py
ds-wook: Done. Thank you!

ds-wook:
@miguelgfierro  Could you let me know when we might receive the review for this?
Also, I’d like to discuss what to proceed with for the next migration task.

recommenders/models/deeprec/models/graphrec/lightgcn.py
```diff
@@ -28,321 +43,404 @@ class LightGCN(object):
...
 
         Raises:
             IOError: if the restore operation failed.
-
         """
         try:
-            self.saver.restore(self.sess, model_path)
+            if model_path is not None and os.path.isdir(model_path):
+                model_path = os.path.join(model_path, MODEL_CHECKPOINT)
+            state_dict = torch.load(model_path, map_location=self.device)
```
miguelgfierro:
```suggestion
            state_dict = torch.load(model_path, map_location=self.device, weights_only=True)
```

torch.load() without weights_only=True — security risk and deprecation.

  state_dict = torch.load(model_path, map_location=self.device)

  PyTorch ≥2.0 emits a FutureWarning and ≥2.4 deprecates the default. Unpickling arbitrary checkpoints is a known
  code-execution vector.

recommenders/models/deeprec/models/graphrec/lightgcn.py
```diff
@@ -28,321 +43,404 @@ class LightGCN(object):
...
+
+            # Single CPU sync per epoch
+            loss = (loss_acc / n_batch).item()
+            mf_loss = (mf_acc / n_batch).item()
+            emb_loss = (emb_acc / n_batch).item()
 
             if np.isnan(loss):
-                print("ERROR: loss is nan.")
+                logger.error("loss is nan.")
                 sys.exit()
```
miguelgfierro:
```suggestion
                raise RuntimeError("Training diverged: loss is NaN.")
```

recommenders/models/deeprec/models/graphrec/lightgcn.py
```diff
@@ -28,321 +43,404 @@ class LightGCN(object):
...
+        # run_eval / infer_embedding) read these.
+        self.data: ImplicitCF | None = None
+        self.batch_size: int = 1024
+        self.decay: float = 0.0
+        self.optimizer: torch.optim.Optimizer | None = None
+
+    @property
+    def ua_embeddings(self) -> torch.Tensor:
+        """Aggregated (LGC-propagated) user embeddings."""
+        u_g, _ = self._propagate()
```
miguelgfierro:
```suggestion
    def ua_embeddings(self) -> torch.Tensor:
        """Aggregated (LGC-propagated) user embeddings."""
        with torch.no_grad():
            u_g, _ = self._propagate()
```

miguelgfierro: ua_embeddings / ia_embeddings properties run full graph convolution without torch.no_grad().

  @property
  def ua_embeddings(self) -> torch.Tensor:
      u_g, _ = self._propagate()
      return u_g

  Each call runs K sparse matmuls and, while in training mode, builds a computation graph that leaks until GC. Worse: if
   a caller accesses both properties, _propagate() runs twice.

  Fix: wrap in torch.no_grad():
  with torch.no_grad():
      u_g, _ = self._propagate()

recommenders/models/deeprec/models/graphrec/lightgcn.py
```diff
@@ -28,321 +43,404 @@ class LightGCN(object):
...
+        return u_g
+
+    @property
+    def ia_embeddings(self) -> torch.Tensor:
+        """Aggregated (LGC-propagated) item embeddings."""
+        _, i_g = self._propagate()
+        return i_g
```
miguelgfierro, line 125:
```suggestion
    def ia_embeddings(self) -> torch.Tensor:
        """Aggregated (LGC-propagated) item embeddings."""
        with torch.no_grad():
            _, i_g = self._propagate()
        return i_g
```

Same as before

miguelgfierro (approved):
This is great @ds-wook 


Some minor corrections, but this is good to go

miguelgfierro:
@ds-wook there is a conflict because we updated the old TF notebook with `map_at_k` that @anargyri was working on. To fix it, the only thing you need to do is to forget about the old code, and make sure you are using `map_at_k` (which you are already doing).

anargyri:
Also you need to rerun the notebook to get the new outputs (numbers and "MAP@K" text).

ds-wook:
@anargyri Notebook re-executed in place with fresh PyTorch outputs (50 epochs on MovieLens 100k). Renamed the eval label from "MAP:" to "MAP@K:" to match the map_at_k call. Final metrics now embedded in the notebook: MAP@K = 0.217, NDCG = 0.347, Precision@K = 0.300, Recall@K = 0.158.

ds-wook:
fork CI fails due to missing org secrets; will run on upstream review

ds-wook:
@miguelgfierro 
1. ua/ia_embeddings properties — Wrapped in torch.no_grad(). Docstring tells callers needing both to use _propagate() directly.
2. map_at_k merge conflict — Already on map_at_k; resolved by staging our version.
3. Notebook rerun — Renamed "MAP:" → "MAP@K:", re-executed lightgcn_deep_dive.ipynb end-to-end. Final metrics: MAP@K=0.217, NDCG=0.347, Precision@K=0.300, Recall@K=0.158.

SimonYansenZhao:
> fork CI fails due to missing org secrets; will run on upstream review

Could merge the staging branch again?  We included a fix for the secret error.

ds-wook:
@SimonYansenZhao
I’m encountering an issue while allocating VMs through the API.

For the first spec (`P40`), the request is parsed successfully but returns a requirements mismatch:

```text
["DEBUG:","Demand (P40) should not be any one of ([\"GPUType\"] - !P40)"]
```

However, for all subsequent GPU types (`2080`, `3080Ti`, `3090`, `4090`, `A800`, `H20`, `A100`, etc.), the API consistently returns:

```text
ERROR: {"RetCode":230,"Message":"Fail to parse parameters"}
```

Example request specs:

```json
{"GPUType":"2080","Memory":{"CPU":40,"GPU":8},"CPU":8,"Price":0.39}
{"GPUType":"3090","Memory":{"CPU":64,"GPU":24},"CPU":16,"Price":1.13}
{"GPUType":"A100","Memory":{"CPU":64,"GPU":80},"CPU":16,"Price":10.21}
```

This suggests the request format may not match the expected API schema, or some `GPUType` values are unsupported/invalid.

Could you help clarify:

1. The expected request schema for VM allocation
2. The valid `GPUType` enum values
3. Whether the nested `Memory` object format is supported

Thanks!

miguelgfierro:
I tried this locally and it works. Merging

PR #2316, opened 2026-05-02
Use map_at_k instead of map for ranking-metric reporting

### LINKED ISSUE #2309: [BUG] MAP metric reported in examples is not MAP@k
### Description

In `benchmark_utils.py` and in many of the Jupyter notebook examples, the MAP@k metric computed and reported comes from the `recommenders.evaluation.python_evaluation.map()` function rather than `map_at_k()`.

Replace instances of `map()` with `map_at_k()` wherever the MAP@k metric is reported.

### Willingness to contribute

- [x] Yes, I can contribute for this issue independently.
- [ ] Yes, I can contribute for this issue with guidance from Recommenders community.

Yinchaochen:
Closes #2309 (partial — see *Out of scope* below).

`benchmark_utils.ranking_metrics_python` and `ranking_metrics_pyspark` returned a dict whose first entry was labelled `"MAP"` but computed with the Spark-style `map()` function (normalised by `n_relevant`), while the rest of the entries used the `@k` variants and were labelled `"@k"`. Reading the row, a user reasonably expects all four numbers to be the `@k` family. This PR makes the label and the function agree, and adds a docstring cross-reference between `map` and `map_at_k` so readers landing on either function can pick the right one.

Pointed out by `@miguelgfierro` in #2309.

### Related Issues
- Closes (in part) #2309
- Refs #1702, #2004 (the earlier ground-work that defined the two functions)

### Checklist:

- [x] This PR is being made to `staging` branch **and not to `main` branch**.
- [x] If you are adding a new utility function, please ensure you have added the function to `recommenders/__init__.py` (already exported).

### What changed

- **`examples/06_benchmarks/benchmark_utils.py`** — `ranking_metrics_python` and `ranking_metrics_pyspark` now call `map_at_k` (resp. `SparkRankingEvaluation.map_at_k()`) and the dict key becomes `"MAP@k"` so it matches `"nDCG@k"` / `"Precision@k"` / `"Recall@k"`.
- **`examples/06_benchmarks/movielens.ipynb`** — the two source cells that consume that dict (the `np.nan`-placeholder for missing rows and the `cols = [...]` ordering) follow the rename. Cached cell outputs are left as-is — they will regenerate on the next notebook run.
- **`recommenders/evaluation/python_evaluation.py`** — `map()` and `map_at_k()` docstrings now cross-reference each other and call out the normalizer difference (`n_relevant` vs `min(k, n_relevant)`).
- **`recommenders/evaluation/spark_evaluation.py`** — same cross-reference on `SparkRankingEvaluation.map` / `.map_at_k`.
- **`tests/unit/recommenders/evaluation/test_python_evaluation.py`** — new `test_python_map_vs_map_at_k` that pins the invariant: the two functions agree at `k=10` on the existing fixture (every user has ≤ 10 relevant items), and `map_at_k` is strictly greater at `k=5` (user 3 has 10 relevant items, so `min(10, 5) < 10`).
- **`tests/test_groups.yml`** — registers the new test in the `pr_gate` group.

### Out of scope (deliberately deferred)

The notebooks under `examples/00_quick_start/` and `examples/02_model_collaborative_filtering/` still import the bare `map` symbol from `python_evaluation`. Switching them is one trivial three-line edit per notebook, **but** the value gets persisted via `store_metadata("map", eval_map)` and is asserted by `tests/functional/examples/test_notebooks_python.py` and `tests/smoke/examples/test_notebooks_*.py` to specific numbers (e.g. `pytest.approx(0.060579, rel=0.05, abs=0.05)`). Switching `map` → `map_at_k` shifts those numbers (the per-user denominator drops from `n_relevant` to `min(k, n_relevant)`), often by more than the tolerance, and re-deriving them requires running each notebook end-to-end against MovieLens (some need GPU). I left those for a follow-up so this PR doesn't break nightly. Happy to take that on too if you'd rather it land together — just point me at a runner.


examples/06_benchmarks/benchmark_utils.py
```diff
@@ -37,7 +37,7 @@
     compute_ranking_predictions,
 )
 from recommenders.evaluation.python_evaluation import (
-    map,
+    map_at_k,
```
miguelgfierro, line 40: this is super useful @Yinchaochen do you see any other notebook where we are using `map` but should be using `map_at_k`? 
FYI @anargyri

miguelgfierro:
@anargyri will take a look and then merge this

examples/06_benchmarks/movielens.ipynb
```diff
@@ -644,7 +644,7 @@
                 "%%time\n",
                 "\n",
                 "# For each data size and each algorithm, a recommender is evaluated. \n",
-                "cols = [\"Data\", \"Algo\", \"K\", \"Train time (s)\", \"Predicting time (s)\", \"RMSE\", \"MAE\", \"R2\", \"Explained Variance\", \"Recommending time (s)\", \"MAP\", \"nDCG@k\", \"Precision@k\", \"Recall@k\"]\n",
+                "cols = [\"Data\", \"Algo\", \"K\", \"Train time (s)\", \"Predicting time (s)\", \"RMSE\", \"MAE\", \"R2\", \"Explained Variance\", \"Recommending time (s)\", \"MAP@k\", \"nDCG@k\", \"Precision@k\", \"Recall@k\"]\n",
```
anargyri, line 647: This notebook should be rerun so that the reported metrics get updated. It does take some time to run though.

recommenders/evaluation/python_evaluation.py
anargyri, line 762: This "top k" should also be removed. But the signature should now change I think, we don't need `k` and `relevancy_method` (because top k is never relevant for this type of metric).

examples/06_benchmarks/benchmark_utils.py
anargyri: They can be left for later PRs.

anargyri: Taken care of in #2323.

examples/06_benchmarks/movielens.ipynb
anargyri: Let me merge this PR and I will run the benchmarks in a separate PR.

recommenders/evaluation/python_evaluation.py
anargyri: Let me merge this PR and I will address this in a separate PR.

anargyri (approved):

PR #2317, opened 2026-05-04
refactor: modernize lightgbm utils

ds-wook:
### Description

Changes to recommenders/models/lightgbm/lightgbm_utils.py:

Import ordering: Sorted imports in accordance with isort conventions (stdlib → third-party).
Added pandas import: Explicitly added import pandas as pd which was previously missing.
Type hints added throughout: All function signatures and instance variables now have full type annotations (e.g., list[str], dict[str, int], tuple[np.ndarray, np.ndarray]).
Class declaration modernized: class NumEncoder(object) → class NumEncoder (idiomatic Python 3 style).
Docstring cleanup: Removed redundant type info from docstring Args: sections since types are now expressed via annotations.
Removed inline comments: Removed low-value comments like # ordinal_encoding and # dynamic_targeting_encoding that just restated what the code does.
Minor cleanup: Removed an unnecessary intermediate samples local variable in fit_transform; used self.samples directly instead.

@miguelgfierro This PR covers only the LightGBM utility refactoring. The full PyTorch migration work is still in progress — once all model migrations are complete, everything will be submitted together in a single PR.

### Checklist:

- [ ] I have followed the [contribution guidelines](https://github.com/recommenders-team/recommenders/blob/main/CONTRIBUTING.md) and code style for this project.
- [ ] I have [signed the commits](https://github.com/recommenders-team/recommenders/wiki/How-to-sign-commits), e.g. `git commit -s -m "your commit message"`.
- [ ] This PR is being made to `staging branch` AND NOT TO `main branch`.

### DIFF
--- a/recommenders/models/lightgbm/lightgbm_utils.py
+++ b/recommenders/models/lightgbm/lightgbm_utils.py
@@ -1,20 +1,22 @@
 # Copyright (c) Recommenders contributors.
 # Licensed under the MIT License.
 
-import logging
-import numpy as np
-import category_encoders as ce
-from tqdm import tqdm
 import collections
 import gc
+import logging
+
+import category_encoders as ce
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
 
 
-def unpackbits(x, num_bits):
+def unpackbits(x: np.ndarray, num_bits: int) -> np.ndarray:
     """Convert a decimal value numpy.ndarray into multi-binary value numpy.ndarray ([1,2]->[[0,1],[1,0]])
 
     Args:
-        x (numpy.ndarray): Decimal array.
-        num_bits (int): The max length of the converted binary value.
+        x: Decimal array.
+        num_bits: The max length of the converted binary value.
     """
     xshape = list(x.shape)
     x = x.reshape([-1, 1])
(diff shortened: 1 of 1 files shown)


recommenders/models/lightgbm/lightgbm_utils.py
```diff
@@ -1,71 +1,80 @@
...
 
-class NumEncoder(object):
+class NumEncoder:
     """Encode all the categorical features into numerical ones by sequential label encoding, sequential count encoding,
     and binary encoding. Additionally, it also filters the low-frequency categories and fills the missing values.
     """
 
-    def __init__(self, cate_cols, nume_cols, label_col, threshold=10, thresrate=0.99):
+    def __init__(
+        self,
```
miguelgfierro, line 33: @ds-wook we were reviewing this pr and one thing we were wondering is whether this approach for the docstring is the recommended approach. 

Is there are reference where we can look at? How will the new documentation look like?

FYI @anargyri

ds-wook: @miguelgfierro Thanks for the question! This approach follows the Google style docstring convention when type annotations are present in the function signature.
Per the Google Python Style Guide:
“If a function has type annotations, omit the type from the docstring.”
So since this PR adds type hints directly to the signatures (e.g. def fit_transform(self, df: pd.DataFrame) -> tuple[np.ndarray, np.ndarray]), removing the redundant (type) from the Args: section is correct and preferred.
The generated documentation (e.g. via Sphinx + Napoleon) will still render the types — it picks them up from the annotations. You can see an example of how this renders at the Sphinx Napoleon docs.
Our Coding Guidelines wiki confirms we use Google style: https://github.com/recommenders-team/recommenders/wiki/Coding-Guidelines

miguelgfierro (approved):

miguelgfierro:
I think it would be good @ds-wook to think about modernizing the rest of the repo following the same guidelines you proposed

PR #2322, opened 2026-05-10
fix: honor benchmark recommendation args

### LINKED ISSUE #1941: [BUG] Benchmark example functions' arguments are not used.
### Description
Some arguments of benchmark utility functions are not used. 
E.g. `remove_seen` and `top_k` [examples/06_benchmarks/benchmark_utils.py](https://github.com/microsoft/recommenders/blob/main/examples/06_benchmarks/benchmark_utils.py)

For example, use the `remove_seen` argument in the codes, e.g. in `recommend_k_als`,

```
...
# Remove seen items
if (remove_seen):
    ...
```

or if the arguments are just placeholders for the compatibility reason, we may use `**_`

he-yufeng:
## Summary

Fixes #1941.

`recommend_k_*` helpers in the benchmark notebook utility accept `top_k` and `remove_seen`, but a few code paths were ignoring one or both arguments. This made benchmark runs inconsistent across algorithms: some returned all candidate items, and ALS/NCF always removed seen items even when callers asked not to.

This changes the helper layer to apply the same ordering everywhere:

1. build candidate predictions
2. apply `remove_seen` when requested
3. keep the top `k` predictions per user

I kept the change inside `examples/06_benchmarks/benchmark_utils.py` so it does not affect the model implementations themselves.

## Validation

Ran:

```bash
python -m py_compile examples/06_benchmarks/benchmark_utils.py
python -m flake8 examples/06_benchmarks/benchmark_utils.py
git diff --check
```

Also ran a small smoke test with a fake NCF model to verify that `top_k=1` limits each user to one row and `remove_seen=True` removes training pairs before ranking.

### DIFF
--- a/examples/06_benchmarks/benchmark_utils.py
+++ b/examples/06_benchmarks/benchmark_utils.py
@@ -9,13 +9,15 @@ import cornac
 
 try:
     from pyspark.ml.recommendation import ALS
+    from pyspark.sql import Window
+    from pyspark.sql.functions import row_number
     from pyspark.sql.types import StructType, StructField
     from pyspark.sql.types import FloatType, IntegerType, LongType
 except ImportError:
     pass  # skip this import if we are not in a Spark environment
 try:
     import surprise  # Put SVD surprise back in core deps when #2224 is fixed
-except:
+except ImportError:
     pass
 
 from recommenders.utils.timer import Timer
@@ -37,12 +39,16 @@ from recommenders.models.surprise.surprise_utils import (
     compute_ranking_predictions,
 )
 from recommenders.evaluation.python_evaluation import (
+    exp_var,
+    get_top_k_items,
     map_at_k,
+    mae,
     ndcg_at_k,
     precision_at_k,
     recall_at_k,
+    rmse,
+    rsquared,
 )
-from recommenders.evaluation.python_evaluation import rmse, mae, rsquared, exp_var
 
 try:
     from recommenders.utils.spark_utils import start_or_get_spark
@@ -71,6 +77,30 @@ except ImportError:
 tmp_dir = TemporaryDirectory()
 TRAIN_FILE = os.path.join(tmp_dir.name, "df_train.csv")
 TEST_FILE = os.path.join(tmp_dir.name, "df_test.csv")
+
+
+def _get_top_k_pandas(predictions, top_k):
+    if top_k is None:
+        return predictions
+    return get_top_k_items(
+        predictions,
+        col_user=DEFAULT_USER_COL,
+        col_rating=DEFAULT_PREDICTION_COL,
+        k=top_k,
+    ).drop("rank", axis=1)
+
+
+def _get_top_k_spark(predictions, top_k):
+    if top_k is None:
+        return predictions
+    window = Window.partitionBy(DEFAULT_USER_COL).orderBy(
+        predictions[DEFAULT_PREDICTION_COL].desc()
+    )
+    return (
+        predictions.withColumn("rank", row_number().over(window))
+        .filter("rank <= {}".format(top_k))
+        .drop("rank")
+    )
 
 
 def prepare_training_als(train, test):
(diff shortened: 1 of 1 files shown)


ds-wook:
Hi! Just a quick note — PRs shouldn’t be opened against the main branch. Could you please submit it to the staging branch instead?

For reference, please see the contributing guide:
https://github.com/recommenders-team/recommenders/blob/main/CONTRIBUTING.md

he-yufeng:
Thanks for the pointer. I retargeted this PR to `staging` and rebased the branch onto current `upstream/staging`.

Validation after the rebase:

- `python -m py_compile examples/06_benchmarks/benchmark_utils.py`
- `git diff --check upstream/staging..HEAD`

The earlier CI failure was from the old `main`-based PR run; this should rerun against `staging` now.

he-yufeng:
Rebased onto current upstream/staging and resolved the benchmark_utils.py conflict. Validation after rebase: python -m py_compile examples/06_benchmarks/benchmark_utils.py; git diff --check upstream/staging..HEAD; changed-file secret scan passed.

examples/06_benchmarks/benchmark_utils.py
```diff
@@ -305,30 +349,34 @@ def recommend_k_ncf(model, test, train, top_k=DEFAULT_K, remove_seen=True):
...
+            train,
+            topk_scores,
+            on=[DEFAULT_USER_COL, DEFAULT_ITEM_COL],
+            how="outer",
         )
+        if remove_seen:
+            topk_scores = merged[merged[DEFAULT_RATING_COL].isnull()].drop(
+                DEFAULT_RATING_COL, axis=1
+            )
+        topk_scores = _get_top_k_pandas(topk_scores, top_k)
```
miguelgfierro: merged is computed unconditionally but is only used when remove_seen=True (line 357). When remove_seen=False, this join is wasted. I suggest to move it inside the if block:
  ▎ if remove_seen:
  ▎     merged = pd.merge(
  ▎         train,
  ▎         topk_scores,
  ▎         on=[DEFAULT_USER_COL, DEFAULT_ITEM_COL],
  ▎         how="outer",
  ▎     )
  ▎     topk_scores = merged[merged[DEFAULT_RATING_COL].isnull()].drop(
  ▎         DEFAULT_RATING_COL, axis=1
  ▎     )
  ▎ topk_scores = _get_top_k_pandas(topk_scores, top_k)

examples/06_benchmarks/benchmark_utils.py
```diff
@@ -338,8 +386,10 @@ def recommend_k_bpr(model, test, train, top_k=DEFAULT_K, remove_seen=True):
             col_prediction=DEFAULT_PREDICTION_COL,
             remove_seen=remove_seen,
         )
+        topk_scores = _get_top_k_pandas(topk_scores, top_k)
```
miguelgfierro: model.recommend_k_items on line 384 already limits results to top_k items per user, so calling _get_top_k_pandas here does nothing and can be removed. See
  ▎ recommend_k_lightgcn (line 451) for the correct pattern: pass top_k to the model and skip the extra filtering step

examples/06_benchmarks/benchmark_utils.py
```diff
@@ -357,6 +407,7 @@ def recommend_k_bivae(model, test, train, top_k=DEFAULT_K, remove_seen=True):
             predcol=DEFAULT_PREDICTION_COL,
             remove_seen=remove_seen,
         )
+        topk_scores = _get_top_k_pandas(topk_scores, top_k)
```
miguelgfierro, line 410: Same as recommend_k_bpr: model.recommend_k_items on line 405 already limits results to top_k items per user, so this call is redundant and should be removed

miguelgfierro (commented):
This is ready good @he-yufeng thanks for noticing.

I added some minor suggestions. 

After these suggestions are added, could you please rerun the benchmark notebook in case the results change?

he-yufeng:
Thanks for the review. I pushed a follow-up in `98ec5157` that moves the NCF train merge under `remove_seen` and passes `top_k` directly into BPR's `recommend_k_items`, removing the redundant post-filter there.

Validation run locally:
- `python -m py_compile examples/06_benchmarks/benchmark_utils.py`
- `git diff --check upstream/staging..HEAD`

I did not rerun the full `movielens.ipynb` locally because this Windows checkout does not have the notebook stack / Spark / Cornac / GPU dependencies installed, and the full benchmark includes PySpark and GPU algorithms. I also left the BiVAE `_get_top_k_pandas` call in place for now because this path currently goes through `predict_ranking(...)`, not a model `recommend_k_items(..., top_k=...)` call, so removing it would stop enforcing `top_k` for that benchmark path. Happy to adjust if you prefer switching BiVAE to a model-native top-k API instead.

anargyri:
Could you merge staging, there is a fix for the testing workflow now.

he-yufeng:
Updated this branch onto current `upstream/staging` so it picks up the testing workflow fix.

Validation after the rebase:

```text
python -m py_compile examples\06_benchmarks\benchmark_utils.py
git diff --check upstream/staging..HEAD
```

The branch still contains the two DCO-signed commits from this PR.

he-yufeng:
I checked the latest failing run after the staging rebase. The failure appears to happen before the benchmark tests start, during the Compshare GPU VM provisioning step:

```text
COMPSHARE_PRIVATE_KEY:
COMPSHARE_PUBLIC_KEY:
ERROR: {"RetCode":230,"Message":"Fail to parse parameters"}
```

The GPU job repeatedly fails while trying to allocate/get VM info, and the CPU Spark jobs are then cancelled. This does not look related to the `benchmark_utils.py` changes in this PR.

Could a maintainer rerun the workflow or check the Compshare CI credentials/provisioning configuration?

miguelgfierro (approved):
LGTM thanks @he-yufeng 
I will rerun the notebook

Issue #2344, opened 2026-06-01
[BUG] Replace numba with PyTorch for getting GPU info


anargyri:
### Description

______________________________ test_get_gpu_info _______________________________

    def get_gpu_info():
        """Get information of GPUs.
    
        Returns:
            list: List of gpu information dictionary as with `device_name`, `total_memory` (in Mb) and `free_memory` (in Mb).
            Returns an empty list if there is no cuda device available.
        """
        gpus = []
        try:
            for gpu in cuda.gpus:
                with gpu:
>                   meminfo = cuda.current_context().get_memory_info()

recommenders/utils/gpu_utils.py:46: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../.venvs/Recommenders/lib/python3.9/site-packages/numba/cuda/cudadrv/devices.py:220: in get_context
    return _runtime.get_or_create_context(devnum)
../.venvs/Recommenders/lib/python3.9/site-packages/numba/cuda/cudadrv/devices.py:138: in get_or_create_context
    return self._get_or_create_context_uncached(devnum)
../.venvs/Recommenders/lib/python3.9/site-packages/numba/cuda/cudadrv/devices.py:153: in _get_or_create_context_uncached
    with driver.get_active_context() as ac:
../.venvs/Recommenders/lib/python3.9/site-packages/numba/cuda/cudadrv/driver.py:505: in __enter__
    driver.cuCtxGetDevice(byref(hdevice))
../.venvs/Recommenders/lib/python3.9/site-packages/numba/cuda/cudadrv/driver.py:327: in safe_cuda_api_call
    self._check_ctypes_error(fname, retcode)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <numba.cuda.cudadrv.driver.Driver object at 0x7dfe2c8f8730>
fname = 'cuCtxGetDevice', retcode = 201

    def _check_ctypes_error(self, fname, retcode):
        if retcode != enums.CUDA_SUCCESS:
            errname = ERROR_MAP.get(retcode, "UNKNOWN_CUDA_ERROR")
            msg = "Call to %s results in %s" % (fname, errname)
            _logger.error(msg)
[...]

PR #2349, opened 2026-06-12
fix: query GPU memory with torch first

he-yufeng:
## Summary
- prefer `torch.cuda.mem_get_info()` for `get_gpu_info()` when PyTorch can see CUDA devices
- keep the existing numba path as a fallback for environments without torch
- add a regression test that fails if the torch-visible CUDA path still touches numba's CUDA context

## Why
Issue #2344 shows `get_gpu_info()` can trip a numba CUDA context error while collecting GPU memory. `get_number_gpus()` already tries torch first, so this makes `get_gpu_info()` follow the same safer path and avoids creating a numba context when torch already has the data.

## To verify
- `python -m py_compile recommenders\utils\gpu_utils.py tests\unit\recommenders\utils\test_gpu_utils.py`
- `.\.venv312\Scripts\python.exe -m pytest tests\unit\recommenders\utils\test_gpu_utils.py::test_get_gpu_info_uses_torch_cuda tests\unit\recommenders\utils\test_gpu_utils.py::test_get_number_gpus_without_torch -q`
- `git diff --cached --check` before commit


he-yufeng:
I checked the red GPU shards. The CPU/Spark groups passed, and the GPU jobs fail before tests run because the external GPU worker setup cannot provide a usable device: repeated cloud allocation errors are followed by \No devices were found\ and exit code 6. This does not appear to be caused by the code change. Could the failed GPU jobs be rerun when capacity is available?

anargyri:
All PRs have to be merged into __staging__ branch. Could you update please.

miguelgfierro (approved):
@he-yufeng thanks for the contribution.

I removed the numba part because we are going full on pytorch

PR #2366, opened 2026-08-24
Migrate xDeepFM from TensorFlow to PyTorch

miguelgfierro:
Migrates xDeepFM from TensorFlow 1.x compat mode to PyTorch, the first of the three models in block 3 of https://github.com/recommenders-team/recommenders/issues/2302.

## What it delivers

- `recommenders/models/deeprec/models/pytorch/xdeepfm.py` — `XDeepFMModel` as a standalone `nn.Module` with all four components (linear, FM, CIN, DNN), plus the `fit` / `run_eval` / `predict` / `load_model` lifecycle. Architecture goes on the constructor and training knobs on `fit`; there is no `hparams` object.
- `recommenders/models/deeprec/models/pytorch/fcn_net.py` — `FcnNet`, the fully-connected scoring head. It takes its activation as an `nn.Module` and its initializer as a callable, because DKN needs sigmoid on its scorer, ReLU on its attention head and a uniform init where xDeepFM uses ReLU and a truncated normal. DKN currently hand-rolls that same network twice, so the port has one class to call instead of two copies to write.
- `recommenders/models/deeprec/io/ffm_dataset.py` — `FFMDataset`, the replacement for `FFMTextIterator`. Batches come out as numpy arrays laid out for `embedding_bag`, and it still reads one mini-batch at a time, so files larger than memory keep working.
- 35 unit tests on synthetic data, one file per module: the loader's parsing, batching, field bagging and emitted dtypes; the head's activation, initializer, dropout in both modes and batch-norm switch; and for the model the CIN's masked first layer, closed-form checks of the linear and FM logits, the additivity of the four components and the training lifecycle. The shared synthetic FFM files live in `tests/unit/recommenders/models/conftest.py`.
- The TensorFlow `xDeepFM.py` and `FFMTextIterator` are removed. `BaseIterator` stays for DKN, the sequential models and newsrec, and no longer imports TensorFlow.

## What the benefit is

xDeepFM was the largest remaining TF-1.x-graph model outside the sequential family, and the one blocking a PyTorch rewrite of `base_model.py`. Removing it deletes the last `tf.compat.v1` session/placeholder pair outside `dkn`/`sequential`, and it gives DKN a working precedent plus the scoring head to build on.

Users get a model they can subclass, checkpoint with `state_dict`, and step through in eager mode, with an explicit signature instead of a yaml-driven `HParams` blob.

## Numerical equivalence

The port was verified against the TensorFlow model directly: build both, copy the TF variables into the torch module, and run the same batch through each. Logits agree to float32 precision (relative error ~1e-7) on each component in isolation and on all four combined, with and without batch normalization, and across `log_loss`, `cross_entropy_loss` and `square_loss`.

End to end on Criteo, at the quick-start notebook's own default settings (10 epochs, batch 4096, seed 42). The TensorFlow figures are the output stored in the notebook committed on `staging`, so this compares against the reference result rather than against a re-run:

| | AUC | logloss |
|---|---|---|
| TensorFlow (committed notebook output) | 0.7356 | 0.5017 |
| PyTorch (this branch) | 0.7355 | 0.5014 |

At the functional test's settings (5 epochs, batch 1024) both were re-run on the same machine: TensorFlow 0.7314 / 0.5057 across seeds 0.7307–0.7314, PyTorch 0.7331 / 0.5034 across seeds 0.7318–0.7333.

The functional test's previous expectation of `auc=0.742` was stale — today's TensorFlow model does not reproduce it either — so the expected values are refreshed from actual runs.

## Deliberate behaviour changes

- **`_build_fast_CIN` is not ported.** It was 160 of the TF model's 534 lines, a TF-op-level parameter-decomposition speed hack gated on `fast_CIN_d`, which defaults to `0` (off) in `deeprec_utils.py`, in the shipped yaml and in the notebook. One einsum-based CIN covers the standard path.
- **The `res` / `direct` / `bias` / `is_masked` switches of `_build_CIN` are not ported.** They were private-method arguments that `_build_graph` hardcoded to a single combination with no config path, so no user could reach any other branch. The CIN implements that one combination.
- **The yaml string switches are gone from the signature.** Every shipped config sets `activation: [relu, relu]`, `cross_activation: identity` and `init_method: tnormal`, so the DNN uses ReLU, the CIN applies no activation, and weights use a truncated normal. `activation`, `cross_activation` and `init_method` are no longer arguments.
- **Redundant flags are gone.** `user_dropout` was a second switch for a state `dropout=[0.0, ...]` already expressed; `is_clip_norm` and `save_model` likewise, so `max_grad_norm=None` and `model_dir=None` now mean "off". `run_eval` and `predict` take ordinary defaults rather than inheriting the values a previous `fit` happened to use.
- **The optimizer is Adam**, as in the SLi-Rec port, rather than TF's nine-way `optimizer` string. Every shipped config uses `adam`.
- Argument names are snake_case (`use_linear_part` rather than `use_Linear_part`).

## Risks CI cannot catch

The embedding table gets dense gradients, where TF's `embedding_lookup_sparse` produced `IndexedSlices`. In practice TF was already dense here because `embed_l2` regularizes the whole table, but a config with `embed_l2 = embed_l1 = 0` and a very large `feature_count` will use more memory per step than the TF version did.

`tests/unit/recommenders/models/` holds nine files that are not Black-clean on `staging`. They are untouched here, so a directory-wide `black` run will show unrelated diffs; that cleanup is out of scope.

## Follow-ups

DKN and DKN_item2item are next; they reuse `models/pytorch/fcn_net.py` and the same loader pattern. `base_model.py` stays until then, since DKN still depends on it.


recommenders/models/deeprec/models/pytorch/layers.py
```diff
@@ -0,0 +1,125 @@
...
+  ``user_dropout`` is set.
+"""
+
+from __future__ import annotations
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+ACTIVATIONS = {
```
ds-wook: Why did you set `ACTIVATIONS` as a constant?

ds-wook (approved):

recommenders/models/deeprec/models/pytorch/layers.py
miguelgfierro: I remade the code

miguelgfierro:
new iteration, please take a look @anargyri @ds-wook

ds-wook (approved):

PR #2369, opened 2026-08-31
Add Hugging Face backup URL for MovieLens downloads

miguelgfierro:
### Description

MovieLens downloads now fall back to a backup mirror when the original GroupLens server is unreachable. The original URL remains the primary source; if the download fails, it is retried against the [Recommenders MovieLens backup](https://huggingface.co/datasets/Recommenders/MovieLens) on Hugging Face, which hosts byte-identical copies of the four dataset zips.

This removes GroupLens as a single point of failure for every MovieLens-based test, notebook and example. The mirror is verified identical to the originals (matching SHA-256 for `ml-100k`, `ml-1m`, `ml-10m` and `ml-20m`), so the fallback returns exactly the same data.

The number of download attempts is now a `maybe_download` argument. The original MovieLens URL uses 3 attempts so an outage fails over to the backup quickly, while the backup keeps the default of 5. All other datasets are unaffected and keep the previous default.

This aligns MovieLens with the approach already used for MIND and Criteo, which are served from the same Hugging Face organisation.

### Related Issues

The GroupLens TLS certificate expired on 28 Aug 2026, so every MovieLens download currently fails certificate verification (the `http://` URL redirects to `https://`). This is a live example of the outage this change protects against.

### References

- Backup dataset: https://huggingface.co/datasets/Recommenders/MovieLens

### Checklist:


miguelgfierro:
there is an issue with the VMs https://github.com/recommenders-team/recommenders/actions/runs/33408312919/job/99802620006 

FYI @SimonYansenZhao

anargyri (approved):

